PR #54, opened 2026-05-28
feat(motion): centralize easing curves + 4 polished animations

IvanWng97:
## Summary

Adds a tiny `tui/anim.rs` module (one enum, one free function) and re-eases three high-visibility animations across the pixel-art office. No new content; same scene, smoother feel.

## What's eased

| Animation | Before | After |
|---|---|---|
| Floor slide transition | linear ease-out-quadratic | `EaseInOutCubic` (smooth S-curve) |
| Version popup entrance | instant | `EaseOutCubic` 200ms scale + fade-in |
| Version popup dismissal | instant | `EaseInQuad` 120ms scale + fade-out |

## What was reverted before merge

- **A\* walk path-tail ease** — added in `6b85f75`, reverted in `47ae204`. The `EaseOutCubic` on global walk progress decelerated position near the path tail, but the leg-swap cadence stayed wall-clock-driven (220 ms per swap, 2-frame walk cycle), so legs kept marching while the body barely moved — a visible "sliding" / moonwalk artifact. A distance-locked frame override was prototyped to keep the smooth stop, but the underlying trade-off (snap-to-stop vs. moonwalk vs. asset work for more walk frames) wasn't worth carrying. Walks are linear again, same as before this PR.

## What was deliberately cut after architecture review

- **Tooltip fade** — ratatui has no native alpha; color-lerp workarounds read as hacky across terminals. Instant tooltips are fine.
- **Theme cross-fade** — `set_theme` flushes the per-floor `FrameCache`; cross-fading would strobe characters for ~15 frames during the blend. Hard cut stays.
- **Camera breathe** — global pixel offset would shift static geometry (wall clock, neon panel) and invalidate the cached layout's hit-test every frame. Character-level breath already exists via `with_breath` in `anchors.rs`.

## Architecture

`tui/anim.rs` exposes:

```rust
pub enum Easing { Linear, EaseOutCubic, EaseInOutCubic, EaseInQuad }

impl Easing { pub fn apply(self, t: f32) -> f32 }

pub fn eased_progress(
    started_at: SystemTime,
    duration_ms: u32,
    easing: Easing,
    now: SystemTime,
) -> f32
```

Free function over a stateful struct because `FloorTransition` and `TuiRenderer` already carry their own `started_at` + duration fields; matches the stateless helper pattern in `pixel_painter/anchors.rs`.

`SystemTime` (not `Instant`) matches existing animation state for v2 daemon-split compatibility. See CLAUDE.md "Known sharp edges".

Popup tween state lives on `TuiRenderer` as `version_popup_started_at: Option<SystemTime>`, edge-tracked by `set_version_popup`. `paint_version_popup` and `version_popup_url_rect` take a `scale: f32` parameter; the URL hit-test gate (`scale >= 0.7`) prevents phantom clicks during the entrance animation.

## Test plan

- [x] `scripts/preflight.sh` — fmt + machete + deny + clippy (-D warnings) + workspace tests (all pass)
- [x] Unit tests for `Easing` curves and `eased_progress` boundaries
- [x] Edge-detection test for `set_version_popup` timestamp (None → false-noop → true-edge → true-noop → false-edge)
- [x] Animation envelope test: entrance start <0.1, mid >0.8, end ≈1.0; dismissal mid in (0,1), end <0.01
- [ ] Manual: PageUp/PageDown on a multi-floor scene — slide should feel smoother
- [ ] Manual: edit `~/.config/pixtuoid/config.toml`'s `last-seen-version` to an older version, relaunch — popup should scale in over 200ms; press Enter and it should scale out over 120ms

## Invariants kept

- `pixtuoid-core` untouched (easing is display-layer only)
- No new dependencies in `Cargo.toml`
- Theme system unchanged
- Hook protocol unchanged
- Public API surface unchanged for downstream users

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -33,6 +33,7 @@ crates/
 │   ├── runtime.rs          tokio task wiring (source ── (Transport, AgentEvent) ──► reducer ──► renderer)
 │   ├── install/            settings.json merge, atomic write, advisory lock, stow-symlink safe
 │   └── tui/                ratatui App + TuiRenderer (Renderer trait impl)
+│       ├── anim.rs         centralized easing curves + eased_progress(start, duration_ms, easing, now) free function — used by floor slide, A* walk path ease, and version popup entrance/dismissal animations
 │       ├── renderer.rs     draw_scene orchestrator (DrawCtx struct), half-block flush, terminal lifecycle
 │       ├── widgets/        ratatui widget paint fns, split into sub-modules:
 │       │                   mod.rs (TickerQueue, shared helpers), hud.rs (footer, wall display,
--- a/crates/pixtuoid/examples/snapshot.rs
+++ b/crates/pixtuoid/examples/snapshot.rs
@@ -196,7 +196,7 @@ fn main() -> Result<()> {
         coffee_holders: &std::collections::HashSet::new(),
         coffee_fetched_at: &std::collections::HashMap::new(),
         new_coffee_carriers: Vec::new(),
-        popup_scale: 1.0,
+        popup_scale: 0.0,
     };
     draw_scene(&mut term, &scene, &pack, now, &mut draw_ctx)?;
 
@@ -654,7 +654,7 @@ fn save_as_gif(
             coffee_holders: &std::collections::HashSet::new(),
             coffee_fetched_at: &std::collections::HashMap::new(),
             new_coffee_carriers: Vec::new(),
-            popup_scale: 1.0,
+            popup_scale: 0.0,
         };
         draw_scene(term, scene, pack, now, &mut draw_ctx)?;
 
--- a/crates/pixtuoid/examples/snapshot.rs
+++ b/crates/pixtuoid/examples/snapshot.rs
@@ -196,7 +196,7 @@ fn main() -> Result<()> {
         coffee_holders: &std::collections::HashSet::new(),
         coffee_fetched_at: &std::collections::HashMap::new(),
         new_coffee_carriers: Vec::new(),
-        version_popup: false,
+        popup_scale: 1.0,
     };
     draw_scene(&mut term, &scene, &pack, now, &mut draw_ctx)?;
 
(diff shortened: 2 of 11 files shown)


IvanWng97:
## Code review fixes applied

All 10 findings from the high-effort code review addressed in 3 follow-up commits on this branch.

### Correctness fixes (commit `6b85f75`)
- **#1 + #2** — Path ease now applied to GLOBAL walk progress (not per-leg). Eliminates the 3× velocity discontinuity at leg boundaries and the path≤2 vs path≥3 inconsistency. All walks now decelerate smoothly regardless of route shape.

### Popup interaction overhaul (commit `0272ad1`)
- **#4** — Mouse handler gates on `popup_scale > 0.0` (via cached `last_popup_scale`) instead of the local `version_popup` bool. Clicks during the 120ms dismissal animation no longer leak through to scene actions.
- **#5** — `set_version_popup` captures the current scale at edge time. `version_popup_scale` lerps from `scale_at_edge` to the target. Dismissing mid-entrance now glides smoothly from the current scale to 0 (no snap-to-1.0 flash).
- **#6** — URL click gate threshold relaxed from `0.99` → `0.7`. URL becomes clickable at ~67ms into the 200ms entrance (instead of ~157ms).
- **#7** — Width/height floors lowered from `.max(4)/.max(3)` to `.max(2)/.max(2)`. Popup now smoothly shrinks below 4×3 instead of snapping to a 4×3 stub.
- **#8** — Mouse handler reads `renderer.last_popup_scale()` (frame-cached) instead of computing scale from a fresh `SystemTime::now()`. Paint-time and click-time scales agree; dual-clock phantom-click window closed.

### Cleanup (commit `f6e7d5b`)
- **#3** — `popup_scale: 0.0` in both `examples/snapshot.rs` construction sites. Visual-verification snapshots no longer have the version popup overlaid.
- **#9** — Removed unused `Easing::EaseOutExpo` variant + its tests (YAGNI).
- **#10** — Added `tui/anim.rs` to `CLAUDE.md`'s crate layout section.

### Test coverage
- New `dismiss_mid_entrance_does_not_snap_to_full` test verifies the lerp-from-edge fix for #5.
- Existing `walk_progress_is_eased_globally` test (renamed + rewritten) verifies global ease for #1+#2.
- All 425 workspace tests pass; preflight green.

Ready for re-review.

PR #62, opened 2026-05-30
fix(motion): eliminate walk teleport/replay regressions + headless render harness

IvanWng97:
## Summary

Fixes five walk-animation glitches introduced by the real-physics pacing in #61, and adds a headless integration-test harness for the TUI render pipeline.

All five bugs were surfaced and pinned by new **frame-by-frame coordinate-continuity tests** — sample `character_anchor` (the real on-screen sprite pixel) at the 33 ms frame cadence under occupancy-overlay churn; a teleport is a >20 px single-frame jump vs ~15 px cruise.

### Motion fixes (`c614e05`)
1. **Mid-walk flash/teleport + periodic stutter** — `route_walking_pose` re-routed A* every frame, so per-frame overlay churn remapped the frozen progress `t` onto a new polyline. Now each leg's polyline is frozen once into `MotionState.walk_path` and reused (this also removes the per-frame A* cost spike behind the "卡一下然后就好" stutter).
2. **Off-screen-floor fast-forward replay** — returning to a hidden floor replayed the whole backlog one phase/frame. Detect a stale resume (gap > one wander cycle) and reuse the analytic bootstrap fast-forward (O(1), no replay).
3. **Wander desk↔waypoint teleport** — `WanderPhase::Seated` fell through to core's *stateless* `idle_pose`, whose independent timeline drifts from the tui motion machine. Render `SeatedIdle` directly (tui motion is the wander authority).
4. **Bootstrap/resume rush** — anchoring the phase clock mid-cycle rushed expired legs one/frame. Always anchor at `now`; keep the `cycle_n` fast-forward.
5. **Exit-while-wandering teleport** — the exit walk hardcoded `from = desk`. Snapshot the agent's current position into `MotionState.exit` so a session ending mid-coffee-run walks from where the sprite actually is.

### Headless render harness (`d0ef45b`)
`TuiRenderer<B: Backend>` is backend-generic, so tests drive the real `render()`/`navigate_floor()` path via ratatui `TestBackend` with no terminal. Output-first assertions (`buf()` pixels + a `frame_buffer()` ratatui-cell inspector); white-box `#[cfg(test)]` accessors only where an invariant isn't observable from output. 31 scenarios across floor navigation, themes, lighting, coffee, pets, version popup, help overlay, footer/labels/tooltips, hit-testing, and graceful degradation.

## Testing
- **552 tests / 25 suites** pass (`cargo test --workspace --features pixtuoid-core/test-renderer`)
- `cargo clippy --workspace --all-targets -- -D warnings` clean
- `cargo fmt --all --check` clean
- Each motion fix's continuity test was watched fail (RED) before the fix.

## Scope notes
- Branch was cut fresh from `origin/main`.
- Skipped as not feasible headlessly: event-loop keyboard state, `open::that` clicks, runtime resize (`TestBackend` size is fixed), and time-of-day pixel effects (timezone-dependent via `chrono::Local`).


IvanWng97:
Addressed review feedback in `eebb13f` (both the local 4-agent review and the online ultra review):

**CLAUDE.md staleness** (online review + local architect) — fixed: `exit` documented as a 3-tuple, added `walk_path`/`WalkPathSnapshot`, the Seated→SeatedIdle render authority, anchor-at-now bootstrap + stale-resume, and a new "Known sharp edges" entry for the path-freeze (so it's not stripped as redundant caching).

**Freeze failure-semantics** (silent-failure-hunter) — the "empty frozen path" is unreachable (`AStarRouter` falls back to `[from,to]`, never empty), but the underlying point holds for that fallback: now only genuinely cornered routes (>2 points) are frozen, so a transient A* fallback re-routes next frame instead of sticking "walk through walls" for the whole leg.

**Test guards** (pr-test-analyzer) — the marquee continuity tests passed even with the freeze reverted (only the white-box call-count test caught it). Added `frozen_leg_anchor_continuous_across_router_shape_change`, a deterministic output-level guard that flips the router shape mid-walk and **fails when the freeze is reverted** (mutation-verified). Added `multiple_agents_share_overlay_without_teleport` for the realistic multi-agent trigger. Tightened `tool_glow` (0→200) and removed the silent no-op in the coffee-machine hit-test.

**Deferred (verified, not blockers):** the snap-back `_snap_prev` "overshoot" finding is contradicted by `wander_interrupted_by_active_does_not_teleport` (samples the snap-back anchor per frame, passes ≤20px) and the per-frame re-route there is pre-existing — flagged as a follow-up, not a regression in this PR.

554 tests / 25 suites, clippy `-D warnings` + fmt clean.

PR #59, opened 2026-05-29
Add Codex hook installation support

trjh:
## Summary
- add `install-hooks --target codex` / `uninstall-hooks --target codex` support for Codex `config.toml`
- decode Codex hook events including `UserPromptSubmit`, `Stop`, and subagent start/stop events
- preserve sibling commands in mixed Codex hook groups while replacing/removing only pixtuoid hook commands
- update docs and regression coverage for Codex install/uninstall

Written with Codex/Reviewed with Claude -- but not really reviewed by a human sorry.  Seems to work!

## Testing
- `docker run --rm -u 501:20 -v /Users/timh/src-personal/pixtuoid:/workspace -w /workspace rust:latest cargo fmt --all --check`
- `docker run --rm -u 501:20 -v /Users/timh/src-personal/pixtuoid:/workspace -w /workspace rust:latest cargo test --workspace`

PR #63, opened 2026-05-30
feat(install): multi-CLI hook install (Target abstraction + Codex)

IvanWng97:
## Summary

Refactors the Claude-only `install-hooks`/`uninstall-hooks` into an **extensible multi-CLI install layer**, and adds **Codex** as the second target. Adding the next hook-supported CLI is now ~one module + one registry line.

**Phases 1–2 only** (this PR). Phase 3 (Codex decoder → sprites) is a separate PR gated on a manual live-Codex checkpoint.

### Architecture
- `install::Target` — `const`-data struct (fn-pointers for per-CLI path / hook-command / format-specific merge) + `TARGETS` registry. No trait/`dyn` (fixed synchronous set).
- One shared `run_install`/`run_uninstall` orchestrator that **owns the backup call** → no target can skip it.
- `io.rs` is now **format-neutral** (`String` in/out; JSON vs TOML lives in each target module). Sibling paths (backup/lock/tmp) use string-append, not `with_extension`.
- `plan_targets` — pure, unit-tested policy: auto-detect + confirm (TTY); non-TTY with multiple CLIs → error (never silently writes Codex's high-value `config.toml`); `--config`/`--settings` without `--target` → Claude (legacy contract).
- CLI: `--target claude|codex|all`, `--config` (alias `--settings`, back-compat), `--yes`. TTY via `std::io::IsTerminal` (no `unsafe`/`libc`).

### Codex specifics (verified against OpenAI docs + `openai/codex` source)
- Hooks are **on by default** → we do **not** write `[features] hooks = true`.
- `_pixtuoid = true` handler sentinel → surgical, path-independent uninstall (Codex's schema has no `deny_unknown_fields`). Two-level group-emptying semantics; mixed groups keep user handlers.
- `timeout = 5`, honest `statusMessage`, `PermissionRequest` event included, `command` carries `PIXTUOID_SOURCE=codex` + absolute path (errors on non-UTF-8).
- Courtesy note on install: `config.toml` comments aren't preserved (backup is the recovery path).

### Review findings from the original draft (#59), structurally prevented
backup always taken (a) · string-append backup names (b) · `Err` on non-UTF-8 hook path (c) · no `transcript_path`-sniffing ported (d).

## Testing
- `cargo test --workspace --features pixtuoid-core/test-renderer` → 532 passed
- `cargo test -p pixtuoid --test install` → 4 passed (incl. the `--settings` back-compat oracle, unchanged)
- `./scripts/preflight.sh` → all checks passed (rustfmt + machete + deny + clippy `-D warnings`)

## Provenance
Absorbs the Codex TOML merge logic from #59 (@trjh), with fixes — credited via `Co-Authored-By`. #59 to be closed with thanks when this lands.

Spec: `docs/superpowers/specs/2026-05-29-multi-cli-install-hooks-design.md` · Plan: `docs/superpowers/plans/2026-05-29-multi-cli-install-hooks.md`


IvanWng97:
## 🔍 Code review (high-effort, 7 angles → verify) — 10 confirmed findings

**Root cause for #1–#3, #5:** the refactor swapped the pre-refactor *semantic* `Value` comparison (`merged != doc`) for a **byte-string** comparison (`merged == content`), but `to_string_pretty` always reformats (key reorder, indent, comment strip). So a semantic no-op reads as "changed".

### Correctness — must fix
1. **`mod.rs:214` (uninstall) — data-loss invariant violation.** Running `uninstall-hooks` on a config with user content but **no pixtuoid hooks** byte-differs after reserialization → takes the "changed" branch → **rewrites the file AND deletes `*.pixtuoid.bak`**. Violates CLAUDE.md "backup is the only recovery path — never destroyed on a no-op."
2. **`mod.rs:216` (uninstall, missing Claude file).** `merge_uninstall("")` → `"{}"` ≠ `""` → **creates** a `{}` `settings.json` where none existed and prints "ok: removed". Old code had an `if !exists()` early return (deleted).
3. **`mod.rs:179` (install no-op).** Re-install on an already-current hand-formatted config reformats the whole file every run (lower severity — backup is idempotent).
4. **`mod.rs:196` (PATH warning).** The "hook not on PATH" warning now lives only in the mutating branch → a no-op re-install gives no warning even when hooks can't fire.
5. **`codex.rs:146`.** A malformed (non-table) `hooks` value → unchanged doc still reserializes differently → backup-destroying rewrite (compounds #1).
6. **`claude.rs:36`.** Invalid-JSON error hardcodes "settings.json", ignoring `--config <path>` → points the user at the wrong file.
7. **`mod.rs:145` (uninstall) — no confirm gate.** Asymmetric with `install()`: bare `uninstall-hooks` with both CLIs present (TTY) silently rewrites BOTH configs and deletes BOTH backups, unprompted. The destructive op is the unguarded one.
8. **`mod.rs:139` (`--target all`) — no rollback.** Claude installs, then a malformed `~/.codex/config.toml` makes `?` abort — Claude already mutated, user told "exit 1" and assumes nothing happened.
9. **`codex.rs:31`.** `PIXTUOID_SOURCE=codex <abs-path>` is unquoted → a hook path containing a space splits under the shell → hooks silently never fire (no warning, `needs_path_warning=false`).

### Cleanup
10. **`mod.rs:39`.** `plan_targets` takes `_yes` it never reads (the confirm gate lives in `install()`). Drop the param.

Fixing all 10 now; will push and reply.

🤖 high-effort review via Claude Code

IvanWng97:
## ✅ All 10 findings addressed in `88d791e`

**Root cause (F1/F2/F3/F5):** `merge_install`/`merge_uninstall` now return `MergeOutcome { content, changed }`, where `changed` is a **semantic (parsed-doc) diff** instead of a byte-string diff. This restores the pre-refactor `merged != doc` behavior the string API had lost.
- **#1** no-op uninstall no longer rewrites the file or deletes `*.pixtuoid.bak` (invariant restored).
- **#2** uninstall on a missing file is now a true no-op (empty doc == empty doc → no file created).
- **#3** re-install on a current config reports "already up to date" (no reformat churn).
- **#5** the malformed-`hooks` path is covered by the same semantic gate.

**Others:**
- **#4** PATH warning moved before the no-op return → emits on every install.
- **#6** invalid-JSON/TOML errors now carry the real `--config` path (`with_context`); per-target messages are generic.
- **#7** `uninstall()` now has the same confirm gate as `install()` (factored into `needs_confirm`/`confirm_targets`) — the destructive op no longer runs unprompted on >1 auto-detected target.
- **#8** `run_each` runs each target independently and reports per-target failure instead of aborting mid-loop and hiding an already-applied target.
- **#9** Codex hook command single-quotes the path → spaces no longer split under the shell.
- **#10** dropped the unused `_yes` param from `plan_targets`.

**New regression tests:** semantic-no-op (claude + codex unit), and CLI-level guards for backup-preservation on no-op uninstall, missing-file-no-create, and spaced-path quoting.

`cargo test --workspace` (358 pkg + 4 integration) + `./scripts/preflight.sh` → green.

PR #66, opened 2026-05-30
fix(motion): freeze snap-back origin instead of re-reading it per frame

IvanWng97:
## What

Snap-back walks now originate from a **frozen** start point captured when the leg arms, instead of re-reading `history.recent` for the Walking `from` on every frame.

## Why

The snap-back arm captured its origin into the `snap_back` tuple but bound it `_snap_prev` (unused) and used the per-frame `prev` for `Pose::Walking { from }`. Because `route_walking_pose` records the advancing walker position into the single-slot `PoseHistory` each frame, the next frame read that advanced point back as the origin — so `from` crept toward the desk frame-by-frame (a contraction, not a walk from a fixed start).

Consequences (both latent — **no observable teleport today**, the contraction is smooth and stays under the 20px continuity cap):
- the leg finished **~2–4× ahead** of its frozen physics profile (ignored its own 900ms budget);
- the per-frame `from` drift stopped the `walk_path` freeze's `wp.from == from` reuse guard from matching after frame 0, forcing a **fresh A\* every frame** on cornered snaps (extra cost + a latent flash window under occupancy churn).

## Fix

Thread the frozen `snap_prev` into `Pose::Walking { from }` (drop the `_`). This **mirrors the exit branch**, which already freezes its origin `Point` — the snap-back field was added in #62 for exactly this purpose but left unwired.

This deliberately does **not** address the separate intra-frame K-call non-idempotency (the `walk_arrived`-clears-`snap_back` mid-fan-out that causes the label/hitbox to lead the sprite by ~11px); that's a distinct follow-up.

## Verification

- New red→green regression test `snap_back_origin_is_frozen_across_frames` — asserts the Walking origin is identical on every frame of the leg (fails on `main`: origin drifts 98→97px by frame 4).
- Full preflight green: rustfmt + cargo-machete + cargo-deny + clippy `-D warnings` + entire workspace suite. All 9 `snap_back_*` + both continuity (`*_does_not_teleport`) tests pass.

### DIFF
--- a/crates/pixtuoid/src/tui/motion/mod.rs
+++ b/crates/pixtuoid/src/tui/motion/mod.rs
@@ -71,8 +71,11 @@ pub struct MotionState {
     /// wander position if it was out, else the desk anchor — so the exit walk
     /// starts where the sprite actually is instead of teleporting to the desk.
     pub exit: Option<(SystemTime, WalkProfile, Point)>,
-    /// `(walk_started_at, profile, snap_target)` for the state-transition
-    /// snap-back walk (replaces the old `since_state < SNAP_BACK_MS` guard).
+    /// `(walk_started_at, profile, from)` for the state-transition snap-back
+    /// walk (replaces the old `since_state < SNAP_BACK_MS` guard). `from` is
+    /// the FROZEN walk origin — the position recorded when the leg armed —
+    /// reused every frame so the walk doesn't drift toward the desk (mirrors
+    /// `exit`).
     pub snap_back: Option<(SystemTime, WalkProfile, Point)>,
 
     // --- cyclic wander state ---
--- a/crates/pixtuoid/src/tui/pose/mod.rs
+++ b/crates/pixtuoid/src/tui/pose/mod.rs
@@ -462,7 +462,7 @@ pub fn derive_with_routing(
                 // unwrap in non-test code).
                 match ms_entry.snap_back.clone() {
                     None => raw,
-                    Some((started_at, profile, _snap_prev)) => {
+                    Some((started_at, profile, snap_prev)) => {
                         let elapsed_ms = now
                             .duration_since(started_at)
                             .unwrap_or(Duration::ZERO)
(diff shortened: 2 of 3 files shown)


IvanWng97:
## Multi-agent code review — verdict: ✅ mergeable, no blockers

Ran a 4-lens review (correctness · architecture · test-quality · silent-failure) with adversarial verification of every finding (default-refuted unless reproducible from the code).

| Lens | Verdict |
|---|---|
| Correctness / logic / edge cases | ✅ clean |
| Silent-failure / fallback | ✅ clean (`None => raw` provably unreachable; no stuck-Walking introduced) |
| Architecture / invariants | concerns → all **nit/low** |
| Test quality | concerns → all **nit/low** |

**Confirmed sound:** the fix is structurally identical to the already-frozen EXIT branch (arms origin → reads it every frame → builds `Pose::Walking { from }`); it restores the `walk_path` freeze reuse guard (`wp.from == from` could never match after frame 0 under the drifting `prev`); time-compression/GC-window interplay is unchanged (profile is built from the same arm-time `prev`, so `walk_arrived` fires at the same instant). 3 findings were **refuted** outright in verification.

**Addressed in `659ab51`** (the two high-value low findings):
- Added `snap_back_cornered_leg_freezes_path_no_reroute` — the origin-freeze test only covers straight 2-point paths, so the cornered `walk_path`-freeze benefit was unasserted. New test mirrors the entry-branch freeze test (counts `router.calls`); **verified red without the fix** (`router.calls` 2≠1).
- Fixed the `MotionState.snap_back` doc comment (`snap_target` → frozen `from`) that this PR makes load-bearing.
- Documented the origin test's ~80px snap-distance rationale.

**Noted, intentionally not in this PR:**
- Gate-divergence (the `dist >= SNAP_BACK_MIN_DIST` gate reads the fresh per-frame `prev` while the walk uses frozen `snap_prev`): real but **pre-existing**, non-regression, and cosmetically benign (fires only within ~7px of the desk, well inside the 20px continuity tolerance). Optional targeted test → follow-up.
- The K-call intra-frame non-idempotency (~11px label/hitbox-leads-sprite desync from `walk_arrived` clearing `snap_back` mid-fan-out) — separate deferred item.

Full preflight green (rustfmt + machete + deny + clippy `-D warnings` + workspace); snap_back suite 10/10.

PR #68, opened 2026-05-31
test(motion): lock in per-frame derive_with_routing idempotency

IvanWng97:
## What

Two regression tests pinning that `derive_with_routing` is **idempotent within a single frame**.

## Why

The render loop calls `derive_with_routing` up to **4× per agent per frame** at the same `now`, sharing one `history` + `motion`:
- `pixel_painter:561` (seated/z-order map), `pixel_painter:876` (paints the sprite), `hit_test:258` (hover box), `paint_label_widgets:287` (label).

If any call's side-effects (`history.record` advancing the walker position, or `walk_arrived` clearing `snap_back`) leaked into a later call, the four consumers would disagree on whether the agent is still walking — the **label and hit-box would lead the painted sprite** (the "K-call desync", originally raised as a deferred follow-up).

## Finding: it no longer reproduces on `main`

I verified the premise before writing any fix (per repo convention) — and the desync **does not reproduce on `cc48b1e`**. #66 froze the snap-back origin, which was the last history-consuming walk that still re-read the *advancing* position each frame; now `route_walking_pose` records the **same point on every call within a frame**. Combined with the exit branch (already freezes its origin) and wander (guarded by `advance_wander`'s `last_advanced_at`), the whole path is idempotent. So there is **no fix to make** — only an invariant to lock in.

## Tests

- `snap_back_derive_is_idempotent_within_a_frame` — full snap-back lifecycle (arm → walk → `walk_arrived` clear), derive 4×/frame, assert pose + recorded history are stable.
- `wander_derive_is_idempotent_within_a_frame` — a coffee-run wander through the **real A\* router**, derive `character_anchor` 4×/frame, assert the anchor is stable.

A future walk that re-reads the advancing history position per frame would re-break this and fail these tests.

No production change. Full preflight green (rustfmt + machete + deny + clippy `-D warnings` + workspace).

### DIFF
--- a/crates/pixtuoid/src/tui/pose/tests.rs
+++ b/crates/pixtuoid/src/tui/pose/tests.rs
@@ -391,6 +391,127 @@ fn snap_back_cornered_leg_freezes_path_no_reroute() {
 }
 
 #[test]
+fn snap_back_derive_is_idempotent_within_a_frame() {
+    // Regression: the render loop calls derive_with_routing up to 4x per agent
+    // per frame (seated map, sprite paint, hit-test, label) at the SAME `now`,
+    // sharing one history + motion. It must be idempotent within a frame —
+    // every call returns the same pose and its side-effects (history.record,
+    // snap_back arm/clear) settle to the same state — else the label and
+    // hit-box lead the painted sprite (the "K-call desync"). This holds only
+    // because every history-consuming walk freezes its origin: snap-back since
+    // #66, exit always, wander via advance_wander's last_advanced_at. A future
+    // walk that re-reads the advancing history position per frame would
+    // re-break it. Step a full snap-back lifecycle (arm -> walk -> walk_arrived
+    // clear), deriving 4x per frame, asserting pose + history are stable.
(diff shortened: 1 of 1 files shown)


IvanWng97:
Good catch — verified and fixed in `10391cb`.

Confirmed the math: `octile([desk+(14,10), desk+(6,4)]) = 104` → triangular `T≈800ms`, and with pause that exceeds `SNAP_BACK_MS=900` → time-compression pushed `walk_arrived` to ~1125ms+, past the 30-frame/990ms window. The leg was actually clearing via the **window-expiry `else` branch** — which can't re-arm — so the clear→**re-arm** path (the one most at risk of a within-frame split) was untested and the comment was wrong.

Fix: shortened the snap to `prev0 = desk+(8,5)` (octile ~24, profile < `SNAP_BACK_MS` so no compression) — `walk_arrived` now fires well inside the window. Also added an explicit assertion that the leg reaches Seated **before** 900ms, so the test provably exercises the `walk_arrived` clear→re-arm path and the comment can't silently drift again. Both tests still green; full preflight passes.

PR #70, opened 2026-05-31
feat(office): 3-seat sofa (meeting + lounge) + fix seated-overlap z-sort

IvanWng97:
## What

- The shared sofa now seats **3** (was 2): meeting-room sofas (2 sofas × 3 = 6/room) **and** the lounge couch (1 → 3).
- Sprite widened `16×7 → 20×7` (3 cushions) and **recoloured** slate-teal for contrast against the warm floor.
- Fixed the bug where a seated occupant's breathing let the **sofa paint over the character** on some frames.
- The lounge couch is now a **group-chat venue** (its 3 sitters bubble together, like the meeting room).

## How

**Layout (`pixtuoid-core`)** — meeting seats `dx {-4,+4}→{-6,0,+6}`; lounge → 3 `Couch` waypoints (`room_id` stays `None`); `SceneLayout.couch_sprite_center` so the couch sprite/rug/table paint once; walkable mask `(8,7)` per seat → unions to the exact 20px footprint (not over-blocked).

**Render (`tui`)**
- *Overlap fix:* the seated occupant's y-sort key uses `anchor_no_breath.y` (not the breathing anchor); the ±1px breath can no longer flip the sofa above its sitter. The sprite still visibly breathes — only z-order is pinned. The existing insertion-order tie-break (decor before characters) keeps the sitter on top.
- *Lounge group-chat:* the 3 couch seats collapse to one venue via `chitchat::venue_wp_idx` (shared first-couch index) — **not** by overloading the meeting-only `room_id` (which indexes `meeting_tables`; reusing it would panic `meeting_slots_face_the_table`).

**Sprite** — `meeting_sofa.sprite` 3 cushions at cols 4/10/16 (one per seat waypoint); 2-tone: fabric `C #7e6a55 → #4f6d77` + new highlight `G #6a8e98`. Cool slate pops vs the warm brown floor (the old taupe blended) and fits the blue/cyan accent family. Visually verified via snapshot → crop → read.

## Design note
Aligned with `feature-dev:code-architect`; I corrected two of its recommendations: (1) lounge grouping at the chitchat layer instead of `room_id = usize::MAX` (which would panic `meeting_tables[room_id]`), and (2) `(8,7)`-per-seat mask instead of `(20,7)`-triple-stamp (which over-blocked ~32px, violating the top-down footprint rule).

## Tests
- `sofas_seat_three_people` (red→green: 3 couch + 6 meeting seats, `couch_sprite_center`, `room_id=None`).
- `lounge_couch_seats_share_one_venue` (deterministic venue grouping).
- Regenerated 4 waypoint goldens (couch 1→3, meeting 4→6) + 2 walkable goldens (couch obstacle 18→22 cols; connectivity intact) — diff verified intentional.
- Relaxed the meeting-room fill canary `<1/2 → <3/4` budget (more slots grew the waypoint pool → deterministic timing shift to 702/1200).
- Full preflight green (fmt + machete + deny + clippy `-D warnings` + workspace). `tui/CLAUDE.md` updated.


IvanWng97:
Both findings verified before acting.

**Finding 1 (`paint_lightning_flash`) — false positive.** I did not add it; `git log -S paint_lightning_flash -- pixel_painter/mod.rs` shows it was introduced solely by `f8a738b` (#69), and it appears in neither of my commits (`04b2b1d`, `2f726be` → 0 occurrences each). It's in this branch only because I merged `origin/main` (#69) in to resolve the conflict. The review diffed against pre-#69 `main` and mis-attributed the merged-in weather code. No change needed (and adding a "wires paint_lightning_flash" note would be misleading — that's #69's, not this PR's).

**Finding 2 (meeting-sofa mask) — premise was inverted; documented instead.** The suggested fix `(20, OBSTACLE_PAD)` is a **24px** block that spans the ~26px-wide meeting room and **disconnects it** — `walkable_mask_is_fully_connected` fails (I tried it). The `16` isn't a stale sprite size: `16 + 2·OBSTACLE_PAD = 20` reproduces the exact 20px X footprint, while the pad supplies the 7→11px **vertical** approach clearance (sit access between sofa and table). The room is too narrow for horizontal clearance, and per the top-down walkable rule characters walk *right up to* furniture sides anyway — so 0 horizontal clearance is correct, not a regression. Kept the behavior and added a comment in `9bfe466` explaining the intent (de-fragilized: it now reads as deliberate, not coincidental).

Full preflight green; no golden/connectivity change.

PR #73, opened 2026-05-31
feat(wayfinder): per-furniture approach sides — stand on the side nearest the desk

IvanWng97:
## Why

Wandering agents targeted a waypoint's `pos`, which for obstacle furniture (pantry counter, vending, printer, phone booth, standing desk) is the **blocked geometric center**. `find_path` then `snap_to_walkable`'d the goal to whatever cell fell out — often the far/south side — so an agent coming from a desk *above* the pantry walked **below** it, and the sprite then **popped up** to a fixed north anchor on arrival. Reported by the owner: "character should go top-down to the pantry, not below it then up."

`facing` already existed on `Waypoint` but was cosmetic (meeting-sofa sprite flip only). Single-point furniture had no notion of an approach side.

## What

New pure-geometry fn `core::layout::stand_point(kind, pos, counter_size, mask, origin)`:

- Probes all **4 sides** of the furniture for the first walkable cell just off the footprint; returns the one **nearest the agent's home desk**.
- Walls naturally drop blocked sides → a pantry counter is approachable **360°**, a recessed appliance only from its open front. Matches the real world.
- **Seat kinds (Couch/MeetingSofa/MeetingStand) pass through their `pos` unchanged** — the seat cell is intentionally blocked (the sprite sits *on* the furniture; A\* snaps the walk adjacent). Unchanged behavior.
- No A\*, no terminal deps — lives in core so the stateless and stateful paths share it.

Wired into **all four** consumers with the same `origin = home desk` so the walk destination and the render anchor can't disagree again:
`core::pose::idle_pose` · `tui::motion::pick_wander_dest` · `anchors::character_anchor` (label) · `pixel_painter` AtWaypoint sprite arm.

## Tests

- `approach.rs`: stand is walkable & off-center; follows the origin side (N/S/E desk → N/S/E stand); seat pass-through; walled furniture picks its only open side; **real-layout walkability guard across all 5 floor seeds**.
- `pose.rs`: regression — an agent's pantry walk destination is walkable (was the blocked center).
- Full preflight green (fmt + machete + deny + clippy `-D warnings` + workspace tests). Waypoint `pos` and the walkable mask never move, so layout/walkable goldens are unchanged.

## Verify

Visual: agents now descend straight to the pantry counter and stand at it from the desk-nearest side (GIF in the PR thread / shared with owner). Tagged `needs-human-verify` — please eyeball a live `pixtuoid run` before merge.


IvanWng97:
## Code review (7 finder angles + verify pass)

**1 real bug found and fixed; 1 maintainability improvement; 2 test gaps closed.** Low/latent items acknowledged below.

### Fixed
- **Occupancy overlay reserved the wrong cell** (`pixel_painter/mod.rs`). The diff moved the agent's standing position to the stand cell but left the per-frame occupancy reservation on the furniture *center* — so a standing agent's actual cell was unreserved and another agent's A\* could route straight through it (acute for the wide pantry, ~18px offset). On `main` these agreed (agent stood at the center). Now reserves the resolved stand cell. **This is the one behavioral regression the change introduced.**
- **Single source of truth for footprints.** `approach::half_extents` was a second hand-copy of the per-kind footprint sizes in `mask.rs`; a future resize of one would silently desync `stand_point`'s clearance. Extracted `approach::obstacle_footprint`, consumed by both the mask stamper and `stand_point`. Walkable goldens unchanged (behavior-identical).

### Tests added
- Motion-side **dest-mirror guard**: `ms.wander_dest` for a pantry trip must equal `stand_point` with the home-desk origin — fails CI if the core/tui walk-dest origins ever drift (the load-bearing invariant).
- **Real-layout side-direction guard**: a north desk yields a stand cell no lower than a south desk's (the whole point — top-down approach).

### Acknowledged, not changed (low/latent)
- **Out-of-range `desk_index` fallback** differs (motion uses origin `(0,0)`; pose/anchors early-return `None`). Not user-visible: such agents aren't rendered. Left as-is.
- **Euclidean-nearest side ≠ A\*-cheapest**: a desk diagonally behind a wall could still detour. Inherent to keeping `stand_point` A\*-free in core; rare and strictly better than the old center-targeting.
- **Rank-offset vs hit-box** (2nd+ co-located visitor): pre-existing, not introduced here.
- **Resize mid-dwell**: one-frame anchor adjustment; narrow and cosmetic.

Full preflight green (fmt + machete + deny + clippy `-D warnings` + workspace tests + goldens).

PR #84, opened 2026-06-03
fix(layout): footprint the meeting screen + bookshelf, tighten the sofa footprint

IvanWng97:
## What

Three walkable-mask footprint fixes in `crates/pixtuoid-core/src/layout/`, all verified with the `--debug-walkable` overlay.

### Floor-standing wall decor had no footprint → agents walked through them
- `Bookshelf`: `None` → `Some((8, 3))`
- `MeetingScreen` (the meeting "TV"): `None` → `Some((14, 3))`

Both stand on the floor against the back wall; their base dips below the window band into the room. The existing `wall_decor` loop in `mask.rs` already south-anchors a shallow base strip to the sprite bottom (`stamp_south_strip`) — adding the footprint auto-wires it. The panel/shelves above overhang the base and sit in the already-blocked window band (invariant #6).

`BulletinBoard` (not placed in the layout) and `ExitSign` (hung high in the band, no floor contact) correctly keep `None`.

### Tightened the meeting sofa footprint
- `MeetingSofaBody`: `Some((16, 7))` → `Some((16, 3))`

Height 7 expanded to **11** blocked rows once the ±2px `OBSTACLE_PAD` was applied, over-blocking walkable floor off the sofa's front and back. `3` makes both axes land exactly on the 20×7 sprite after the pad (width `16+4=20`, height `3+4=7`). Width stays 16 on purpose — a literal 20 becomes 24 with pad and disconnects the narrowest meeting room.

## Verification
- `--debug-walkable` overlay (diffed vs clean render): screen + bookshelf bases now blocked; sofa footprint hugs the sprite.
- Connectivity report: **0 disconnected** pixels across all buffer sizes + floor seeds.
- 96×70 walkable **golden snapshots** regenerated and eyeballed (the footprints land as a shallow strip off the north wall band; room interior stays open).
- Full `just preflight` green (fmt + clippy + machete + deny + 600+ tests).

## Human-verify
Static overlay checks pass; needs a live `./target/release/pixtuoid run` glance to confirm sprites read right and agents route around the screen/bookshelf.


IvanWng97:
Addressed the MEDIUM finding in `0deeb76`.

Root cause was a pre-existing inaccuracy: `MeetingSofaBody.visual` was `(20, 8)` but the real `meeting_sofa.sprite` is **20w × 7 data rows**. Rather than just soften the comment, corrected `visual` → `(20, 7)` so it matches the sprite — which also makes the footprint coverage genuinely exact (`footprint 3 + 2·OBSTACLE_PAD = 7 = visual.h`).

Verified safe: the sofa render centers on the actual sprite frame (not `def.visual`), so rendering is unchanged; the only `visual(MeetingSofaBody)` consumer is the hover hitbox, now 1px more accurate; invariant `footprint ≤ visual` still holds (16≤20, 3≤7); mask is unchanged so the golden snapshots don't move.

Also ran a local 2-agent review (reviewer + architecture trace) — no further findings.

PR #112, opened 2026-06-04
feat(site): Claude-style Astro landing page + Pages deploy

IvanWng97:
## What

A terminal-pixel-art marketing site for pixtuoid, built with **Astro** and deployed to GitHub Pages → **https://ivanwng97.github.io/pixtuoid/**. Self-contained `site/` subtree — **no Rust/workspace changes**.

## Design — "Cozy Terminal"
- Warm **"Coworking" palette**: a cream lifted from the office carpet (`#8A765F`) + Claude coral — analogous warm harmony so the office art sits *into* the page (60-30-10, coral = the 10% accent). Day = cream, night = after-hours dark.
- Type: **Pixelify Sans** (display) · **JetBrains Mono** (UI/code) · **Lora** (body).
- Sections: hero · activity ticker · theme switcher · feature bento · how-it-works · install · footer.

## Interactive FX (all `prefers-reduced-motion`-safe)
day/night toggle · **click-to-retint theme chips** (whole page recolors, like the app's `t`) · cursor glow · 3D tilt · headline shimmer · CRT power-on · pixel-dust · **boot intro** (once/session) · **typewriter → CRT** hero · **live activity ticker** · `dracula` **easter egg** (type it, or `?theme=dracula`).

## Dynamic themes
`site/src/themes.json` is the single source of truth — the switcher chips, the live "N built-in themes" count, the page retint, and the render script all read it. **Adding a theme = one line + `gen-demos.sh`.**

## Demo art
Generated from the pixtuoid binary via `site/scripts/gen-demos.sh` (per-theme shots, day/night, the hero `.mp4`) — never hand-placed, so it can't drift.

## Quality / process
- Gates (all green locally): `prettier --check` · `eslint` (TS + Astro + a11y) · `astro check` (0 errors) · `astro build` · `shellcheck`.
- `npm run verify` one-shot · the **`site`** CI workflow runs the same on PRs · **`pages.yml`** deploys via `withastro/action@v6` on merge.
- `just site-{setup,dev,check,fmt,demos}` from the repo root.

## ⚠️ One-time after merge
**Settings → Pages → Source: GitHub Actions** (so the deploy workflow can publish).

## Review
AI-built — please run the review process + **visually verify** before merge (`just site-dev` → http://localhost:4321/pixtuoid/). Tagged `needs-human-verify`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #113, opened 2026-06-04
feat(dev): one-command release bump + route all build/lint/format/test through just

### LINKED ISSUE #110: Guard: assert pixtuoid→pixtuoid-core dep version tracks the workspace version
## Problem

`crates/pixtuoid/Cargo.toml` pins the path dependency as a **hardcoded version string** that duplicates `[workspace.package] version`:

```toml
pixtuoid-core = { path = "../pixtuoid-core", version = "0.5.0" }
```

Cargo offers no inheritance for a dependency *constraint* string — `version.workspace = true` only applies to a package's own `version` field, not to a dep's `version = "…"`. So this must be hand-bumped on every release.

It already drifted once: it sat at `version = "0.4.0"` through the entire 0.4.1 release and only didn't break because `^0.4.0 ⊇ 0.4.1`. For 0.5.0 (PR #102, commit `ac0dd70`) it had to be bumped by hand or `cargo check` and the crates.io publish would fail (`^0.4.0` excludes 0.5.0). `cargo-semver-checks` does **not** catch this — it diffs the published API surface, not the dep-constraint string.

## Why deferred

Not blocking PR #102 — the constraint is correct for 0.5.0 right now. This is future-drift prevention.

## Fix sketch

Add a check to `release.yml`'s `check` job (next to the existing tag==version validation) and/or a fast local `just` recipe:

```bash
WS=$(grep -m1 '^version' Cargo.toml | cut -d'"' -f2)
DEP=$(grep 'pixtuoid-core.*version' crates/pixtuoid/Cargo.toml | sed -E 's/.*version = "([^"]+)".*/\1/')
[ "$WS" = "$DEP" ] || { echo "::error::pixtuoid-core dep version ($DEP) != workspace ($WS)"; exit 1; }
```

Surfaced by the PR #102 pre-merge review (versioning-strategy lens).

IvanWng97:
Two related single-source wins: a one-command release bump, and **every** build/lint/format/test call-site routed through the justfile.

## `just bump X.Y.Z` (release flow)
Cutting a release was 4 hand-coordinated edits — workspace version, the `pixtuoid`→`pixtuoid-core` path-dep requirement (the classic missed one, #110), the `version.rs` `release_notes()` arm, and `Cargo.lock`. Now one command:

- `cargo set-version --workspace` rewrites the version + path-dep + `Cargo.lock` together (closes #110 at the source; the guard test stays as a backstop)
- drafts the `release_notes()` arm from `git log <last-tag>..HEAD`
- runs `just preflight`, commits on `release/vX.Y.Z`
- **stops before the tag** — pushing the tag is the irreversible crates.io publish, so a human owns it

## justfile = the one definition of build / lint / format / test
Previously ci.yml mostly called `just` recipes but release.yml inlined raw `cargo fmt/clippy/test`, and both ci.yml + release.yml had raw `cargo build`/`cargo deb`/`cargo llvm-cov`. Now every call-site goes through a recipe:

- **new recipes:** `just build *args`, `just build-target TARGET CROSS` (per-target cross release build), `just deb TARGET`, `just coverage`
- **ci.yml:** coverage → `just coverage`, smoke build → `just build --release --bins --examples`
- **release.yml:** check job → `just fmt-check`/`clippy`/`test`; build matrix → `just build-target`; deb → `just deb`
- only the irreversible `cargo publish` stays inline
- `just --list` now groups recipes **check / build / release / docs** — one command per intent
- `cliff.toml` skips the `chore(release)` commit in its own changelog
- README / CONTRIBUTING / CLAUDE.md document the flow; fixed a stale preflight description (missing `hack`)

## Verification
- `just bump 0.5.1` end-to-end on a throwaway branch: `set-version` rewrote all three version spots incl. the path-dep; generated `version.rs` compiles + the `current_version_has_release_notes` guard passes; committed exactly the 4 expected files.
- `cargo nextest run -p pixtuoid version` → **30 passed**
- shellcheck clean on the recipe bodies; `just --list` renders the groups; ci.yml + release.yml + cliff.toml validated.
- The ci.yml smoke (`just build`) + coverage (`just coverage`) paths run on this PR; the release.yml `build-target`/`deb` paths only run on a tag (verified locally instead — recipes mirror the prior inline commands).

## Related
The companion **docs → site auto-deploy fix** (`site.yml`/`pages.yml` path filters) is on #112.


IvanWng97:
Addressed both review findings in `8fa5601` (+ a second-pass multi-agent review):

**Finding 1 (awk anchor / `justfile`)** — adopted your suggestion: `just bump` now injects after a dedicated `// [bump-inject-here]` marker in `version.rs` (single-line, with the rationale as a comment above it), and the awk anchor is `/\[bump-inject-here\]/` — whitespace-independent, no longer coupled to the `match` brace's indentation. Verified end-to-end: the new arm injects cleanly after the marker.

**Finding 2 (release `check` parity / `release.yml`)** — chose the "honest comment" option rather than `just preflight`: the comment now documents that the release check is *intentionally* lighter (fmt/clippy/test) because the tag comes from `main`, which already passed the full CI matrix — and deliberately omits `just lint`/`just hack` so the irreversible tag path doesn't pull in the network-dependent `cargo deny` advisory fetch.

Second-pass review also fixed: a SIGPIPE-under-`pipefail` false-negative in the publish-retry classifier (now uses `grep <<<"$out"` herestrings, no pipe), the bump cleanup now `git restore --staged --worktree` + drops an orphan release branch on failure (previously left the bump *staged* if the commit step failed), tightened the #110 path-dep match, and `[doc()]`'d the last recipe that rendered a comment-fragment in `just --list`.

PR #131, opened 2026-06-05
refactor(state): remove the never-used Activity::Thinking variant (#94)

### LINKED ISSUE #94: refactor(state): remove never-used Activity::Thinking variant
**Problem:** `Activity::Thinking` is defined but never produced or read — genuinely dead.

**Why deferred:** the tempting bundle (also delete `Reading`, rewrite the 4 `tests/reducer.rs` literals to `Typing`) is unsafe — those `Reading` literals are deliberate distinct-value discriminators for the hook-wins-dedup + event→state value-fidelity tests; collapsing them makes those assertions vacuous.

**Fix sketch:** delete **only** `Activity::Thinking`. Keep `Reading` as the JSONL-vs-hook discriminator. The serde derive is for the v2-daemon split, not a variant-count-pinned wire format, so deleting Thinking breaks no serialization contract. (Optional follow-up: remove the whole write-only `activity` field once a non-discriminator way to pin hook-wins-dedup exists.)

---
_Surfaced by the whole-codebase refactor-discovery workflow (run `wf_4023ba70-fb8`); deferred from the Batch 1+2 DRY sweep on `refactor/codebase-dry-sweep`. Verified real by adversarial review._

IvanWng97:
Closes #94.

## Problem
`Activity::Thinking` was defined but **never constructed or matched** anywhere — genuinely dead (`Typing` is produced 12+ times; `Reading` is used in tests; `Thinking` zero).

## Fix
Delete only the `Thinking` variant. `Reading` stays — it's the deliberate JSONL-vs-hook discriminator for the hook-wins-dedup + event→state value-fidelity tests. The `Serialize`/`Deserialize` derive is for the v2-daemon split, not a variant-count-pinned wire format, so removing `Thinking` breaks no serialization contract.

`just preflight` green (pre-push hook).

### DIFF
--- a/crates/pixtuoid-core/src/source/mod.rs
+++ b/crates/pixtuoid-core/src/source/mod.rs
@@ -64,7 +64,6 @@ pub enum Transport {
 pub enum Activity {
     Typing,
     Reading,
-    Thinking,
 }
 
 /// Structured tool detail. Replaces the free-form `Option<String>` so the


IvanWng97:
**Parked for the 0.6.0 breaking batch — do not merge at 0.5.x.**

CI's `semver-checks` correctly fails with `enum_variant_missing`: `Activity` is public API in the published `pixtuoid-core` 0.5.0, so removing `Thinking` requires a major (0.6.0) bump. The code itself is verified dead (sole remaining ref is the declaration; re-checked post-#136 — the registry/custom-decoder work added no use).

Merge sequencing when cutting 0.6.0: merge this immediately before the `just bump 0.6.0` PR so main's semver gate is red only briefly. Batch-mate: the `STALE_CODEX_IDLE_TIMEOUT` rename deferred in #136 (noted in a code comment at the const).

PR #134, opened 2026-06-05
feat(reasonix): hook-only integration — payload decoder, cwd-keyed sprites, install-hooks target

zirs97:
## Summary

Adds **DeepSeek-Reasonix** (the Go line, releases v1.0.0+ / `reasonix@next`) as a supported agent CLI — **hook-only**: `pixtuoid install-hooks --target reasonix` registers the shim in the global `~/.reasonix/settings.json`, and `rx·project` sprites get live per-tool activity, approval-Waiting, Delegating (subagent dispatch), clean `SessionEnd` exits, and a `UserPromptSubmit` resurrect path.

The original JSONL-watcher approach was rewritten by the maintainers after verifying the wire format against upstream `@v1.2.0` — see the [takeover comment](https://github.com/IvanWng97/pixtuoid/pull/134#issuecomment-4636170465) below for the full evidence (wrong sessions dir, flat ToolCall shape, full-rewrite-per-turn save model, no hook↔JSONL coalescing key). Contributor commits are preserved in the branch history.

## Related issue

Follow-up tracked in #135 (replace the hand-constructed fixture with a sanitized live capture).

## Type of change

- [x] New agent-CLI integration (hook-only — no `Source` adapter; see the updated invariant #3 in CLAUDE.md)

## How I tested it

- `just preflight` (fmt + machete + deny + clippy -D warnings + hack + 1032 tests) — green
- 30 new tests: 17 decoder units, install merge/idempotency/e2e round-trip, hook-only fixture + insta golden, 5 reducer lifecycle pins (`/new` rotation resurrect, Delegating stale carve-out, turn-end Waiting resolution + Codex race protection)
- `scripts/check_upstream_drift.py` extended to Reasonix and smoke-tested live against upstream (no drift)
- Two multi-agent adversarial reviews (30 + 22 confirmed findings, all addressed)
- ⏳ Live verification against a real Reasonix install pending — see below

## Visual verification

- [ ] Pending: `rx·` sprite screenshot from a live Reasonix session (author or maintainer — see takeover comment)

## Checklist

- [x] Read the relevant `CLAUDE.md`s (and updated them: hook-only is now invariant #3's documented exception)
- [x] New behavior has tests (TDD)
- [x] No `unwrap()` in non-test code
- [x] No new `println!`/`eprintln!` on production paths
- [x] Docs updated in the same commits (README, 3× CLAUDE.md, ARCHITECTURE, CONTRIBUTING, copilot-instructions, fixtures README, drift watch)
[...]

PR #136, opened 2026-06-06
refactor(source): one SourceDescriptor registry row per CLI + capability-driven lifecycle policy

IvanWng97:
## Summary

Adding agent-CLI #4 (PR #134) required restating "this source exists" across five core registry sites. This consolidates every cross-source fact into **one `SourceDescriptor` row** in `source/registry.rs` (const fn-pointer table — the `install::Target` pattern), and replaces the reducer's source-name gates with **capability flags**:

- `decoder.rs` → registry-driven dispatcher: a source's optional custom hook decoder runs *before* the shared field requirements (alien envelopes decode in their own module); AgentId keying comes from `HookDecoding::id_key`. Codex's subject-changing `SubagentStart/Stop` moved to `codex::decode_codex_hook_custom`.
- `reducer.rs` → `stale_threshold` split into lookup + policy halves; the Codex 5-min idle carve-out is now the derived policy `SourceCaps::short_idle_reap() = !has_exit_signal && resurrects_on_prompt` (pinned: fires for Codex **only** — antigravity lacks the signal too but can't resurrect, hence its long window). The Delegating hook-silent arm is in place for #134, pinned with a synthetic-caps test.
- Fixture harness: decoder + hook-only-ness come from the row (`line_decoder: None` ⇒ transcript-less scenario required, and only then).
- `#[doc(hidden)]` registry: pub only for integration tests; cargo-semver-checks ignores it, so future caps fields don't burn a breaking bump.

**Adding a CLI is now**: source module + ONE registry row + the name in `REGISTERED_SOURCES` (bridge-tested) + runtime wiring. CONTRIBUTING.md gained the 9-step guide; ARCHITECTURE.md the registry seam bullet.

## Behavior

Preserved for cc/codex/antigravity — identical event streams and lifecycle policy; all pre-existing tests pass unchanged (1008 total, preflight green). One deliberate, pinned narrowing: `SubagentStart/Stop` stamped by a **non-codex** source now bails instead of minting a never-coalescing phantom child (unreachable via our own installs).

## Sequencing

- Independent of #133 (no install/ changes) — merge order free.
- **#134 rebases onto this**: its reasonix code collapses into one `REASONIX` row (`line_decoder: None`, `custom: Some(decode_rx_hook_payload).map(Some)` adapter, `delegations_are_hook_silent: true`, `has_exit_signal: true`). Full rebase runbook (merge-tree-verified: one reducer conflict region, harness/mod/decoder resolutions specified) in the review artifacts.

## How I tested it

- `just preflight` green (fmt, clippy -D warnings, hack feature-powerset, 1008 tests)
- TDD: 13 new tests (registry shape + bridge, synthetic-caps Delegating policy, claim-fully guards, non-codex Subagent bail, unknown-source version-skew pin)
- Two review rounds: rust-skills idiomatic-pattern review (8 findings — const-table choice validated vs enum/trait-object/inventory; `type_complexity` alias; the doc(hidden) semver treatment) + a 4-dimension adversarial multi-agent review (26 confirmed findings, 0 refuted — all addressed; the 2 majors became implemented features rather than re-tensed docs)

## Type of change

- [x] Refactor / chore (behavior-preserving)


IvanWng97:
Re the standing MEDIUM (`delegations_are_hook_silent` YAGNI): **consciously retained**, for the record:

1. The alternative re-opens a worse finding. Our round-1 review's major was "registry docs advertise behavior that doesn't exist" — the cures were implement-the-arm (chosen) or re-tense the docs. Re-tensing re-creates a merge-tree-verified rebase trap for #134: its name-gated Delegating arm **auto-merges cleanly** over this refactor, silently stranding the cap as dead data and resurrecting the exact name-gate pattern this PR deletes.
2. Not untested speculation: `stale_threshold` was split into lookup + policy halves precisely so `delegating_slot_with_hook_silent_caps_gets_waiting_window` pins the arm (positive + detail-gate negative) before any row flips the flag.
3. Zero behavioral surface today (all rows `false`).

YAGNI yields to "the consuming PR (#134) is already open and its rebase silently mis-resolves without this."

PR #141, opened 2026-06-06
refactor(coverage): isolate the untestable-headless residue (#103)

### LINKED ISSUE #103: chore(coverage): track the untestable-headless residue + optional isolation refactors
## Context

PR #101 raised project coverage to **97.32%** (codecov-adjusted) by writing tests for every *headless-reachable* gap and deleting dead code, with **zero `codecov.yml` bypass flags added**. A small residue (~125 raw lines / the ~51 "missing patch" lines codecov flagged) stays uncovered because it is **structurally unreachable by any headless test** — verified against BOTH the unit and the harness tiers in the PR #101 review (the adversarial-verify phase flipped ~30 over-eager "untestable" calls back to testable; what's left is genuine).

This issue tracks the **optional** refactors that would let those lines be removed from the coverage denominator *if* a clean 100%-on-those-files is ever wanted. **None are required** — `codecov.yml` has `project: informational: true`, and leaving the residue visible (rather than hiding it behind a file ignore) is the honest default. Filed only so the convention "track every deferred finding as a gh issue" is satisfied.

## The residue (all real-TTY / async-runtime / dead-by-construction glue)

| File | Lines | Why no unit/harness test reaches it | Isolation option |
|---|---|---|---|
| `pixtuoid/src/runtime.rs` | ~110 | `run`/`run_async`/`reducer_task`/`headless_loop`/`compute_boot_capacities` build a tokio runtime + `block_on` + spawn + `ctrl_c` + bind a socket | Extract them into `runtime/driver.rs`, then add **only that file** to `codecov.yml ignore:`. Keep `RunConfig`/`summarize`/the capacity helpers in `runtime.rs` (they ARE tested). |
| `pixtuoid/src/tui/renderer.rs` | ~12 | `setup_terminal`/`teardown_terminal` = `enable_raw_mode` + `EnterAlternateScreen` on real stdout | Move the two fns into a tiny `tui/terminal_lifecycle.rs` (or back into the already-ignored `tui/mod.rs`) and ignore that file. The rest of `renderer.rs` (`draw_scene`/`scene_rect`/`flush`) is well covered. |
| `pixtuoid-core/src/source/claude_code.rs` | ~4 | the `Source::run` `tokio::select!` abort/warn arm BODIES (both inner loops are infinite; forcing one to win is racy) | Accept as residue. A thin `wire()` helper could isolate them but it's not worth the churn. |
| `pixtuoid-hook/src/main.rs` | ~2 | the stdin read-error return + the `enrich_payload` call in `main()` — behaviorally tested end-to-end by `tests/shim.rs`, just uncaptured because the shim runs as a spawned subprocess | Accept as residue (a whole-file ignore would also drop the well-tested `enrich_payload`/`default_socket_path` from the denominator). |

## Also: hook env-test enforcement (claude[bot] finding on PR #101)

`pixtuoid-hook`'s `default_socket_path_branches` test mutates process-global env vars. There is **only one** such test in that binary today → no current race (`save`/`restore` is correct). If a second env-reading/mutating test is ever added to the hook crate, enforce mutual exclusion via a **crate-internal mutex** — the pattern PR #101 introduced in `pixtuoid/src/lib.rs` (`TEST_ENV_LOCK`) — rather than the `serial_test` dev-dependency, which this repo deliberately avoids.

## Why deferred / not done in #101

The bypass mechanism on stable Rust is file-level only (no per-fn `#[coverage(off)]`), so removing a *partial-file* residue requires the small isolation refactor above. Low priority — the residue is honest, the gates pass, and `codecov/project` is informational. Surfaced by the coverage-weighted multi-agent CR on PR #101.

IvanWng97:
Closes #103.

## What

Splits the structurally-untestable glue out of two well-covered files so codecov's denominator only counts headlessly-reachable code, following the existing "keeps the % honest" convention (`main.rs`, `tui/mod.rs`):

- **`runtime.rs` → `runtime/mod.rs` + `runtime/driver.rs`** — `mod.rs` keeps everything unit-tested (`RunConfig`, `SceneRx`, boot-capacity math, `summarize` + the 10-test suite); `driver.rs` takes the async glue (`run`/`run_async`/`reducer_task`/`headless_loop`/`compute_boot_capacities` — tokio `block_on` + `ctrl_c` + socket bind, unreachable by any headless test) and is codecov-ignored. `runtime::run` is re-exported, so `main.rs` is untouched.
- **`tui/renderer.rs` → `tui/mod.rs`**: `Term` + `setup_terminal`/`teardown_terminal` move into the already-ignored event-loop file — raw mode + alternate screen belong with the loop that owns them. No new ignore entry needed; `renderer.rs` (`draw_scene`/flush) now carries no real-TTY code.
- **Accepted as residue** per the issue: the `claude_code.rs` `select!` abort arms (forcing one infinite loop to win is racy) and the 2 uncaptured hook-shim lines (behaviorally covered by `tests/shim.rs` as a subprocess).

Behavior-preserving move — no logic changes.

## Review (10 agents: 3 dimensions × adversarial verify, 7 confirmed / 0 refuted, all fixed in-PR)

- **`claude-security-review.yml` paths filter was dead**: every entry except `pixtuoid-hook/**` was missing the `src/` segment, so the AI security review has *never* fired on its intended install/runtime/source changes (pre-existing since its introduction). Fixed; the runtime entry now targets `src/runtime/**`.
- Stale doc cross-refs updated: `registry.rs` module doc, `pixel_painter/mod.rs` header.
- **Semver note (advisory, no action)**: this deletes pub paths `pixtuoid::tui::renderer::{Term, setup_terminal, teardown_terminal}` from the published `pixtuoid` 0.5.0 lib target — technically major, but zero consumers exist, CI's semver gate covers `pixtuoid-core` only, and the lib surface is documented internal-facing. CLAUDE.md now records that only `pixtuoid-core` is a semver surface.

## Verification

- `just preflight` green (lint → clippy → hack → test), 1014/1014.
- Docs currency sweep: 4 CLAUDE.md files, CONTRIBUTING.md, add-source.prompt.md, codecov.yml, + the two in-code doc comments the review caught.


IvanWng97:
claude[bot]'s finding is valid and fixed in db36c54 — both workspace `CLAUDE.md` cross-refs now point at `runtime/driver.rs::run_async`. Also swept the two remaining same-class prose refs the review didn't flag (`docs/CONTRIBUTING.md:151`, `.github/prompts/add-source.prompt.md:20`). `grep -rn 'runtime::run_async' --include='*.md'` is now clean repo-wide.

Note on the red `security-review` check: that's the claude-code-action's anti-tamper guard — this PR modifies `claude-security-review.yml` itself (the paths-filter fix), and the action refuses to run a workflow file that differs from the default branch's version. The action's own error text says to ignore it in this case; it self-heals once merged.

PR #142, opened 2026-06-06
refactor(reducer): extract suppress_subagent_leak + track_active_tasks pre-passes (#90)

### LINKED ISSUE #90: refactor(reducer): decompose apply() — extract suppress_subagent_leak + track_active_tasks pre-passes
**Batch 3 (risk: HIGH).** `state/reducer.rs::apply()` is ~320 lines and is the single most-pinned fn in the crate (51 reducer tests). It opens with two cohesive cross-slot pre-passes (`suppress_subagent_leak` + `track_active_tasks`) that read cleanly as named helpers.

**Why deferred:** the obvious extraction is dangerous. The hook-wins dedup MUST stay *between* suppress and track_active_tasks — if dedup runs after track_active_tasks, a JSONL duplicate drains `active_tasks` and fires `cascade_exit` before being dropped. There's also borrow-checker friction (`&mut self` + `&mut scene` while reading `self.active_tasks`). Needs an explicit ordering-pinned plan and a pure mechanical move, green-to-green.

**Fix sketch:** extract `suppress_subagent_leak(&self, ...)` and `track_active_tasks(&mut self, ...)` as named methods, preserving the exact sequence suppress → **dedup** → track_active_tasks. Keep b1 completed-subtree cascade ordering, hook-wins dedup, and `gated_before_waiting` restore byte-identical. Run the full 51-test reducer suite before and after.

---
_Surfaced by the whole-codebase refactor-discovery workflow (run `wf_4023ba70-fb8`); deferred from the Batch 1+2 DRY sweep on `refactor/codebase-dry-sweep`. Verified real by adversarial review._

IvanWng97:
Closes #90.

## What

Decomposes `apply()`'s two cohesive cross-slot pre-passes into named private methods on `Reducer` — `suppress_subagent_leak` + `track_active_tasks` — leaving the hook-wins dedup **inline between the two call sites**, exactly as the issue's ordering constraint demands. The handled flags travel via a private `TaskTracking` struct (destructured at the call site, per the PR #89 named-bundle idiom). Cross-slot correlation maps stay in `reducer.rs` per the Layer-A/Layer-B invariant; `fsm.rs`/`scope.rs` untouched.

## The ordering pins (the issue's whole premise)

The pre-pass order is now documented at the one reorderable call site and pinned by **two forcing-function tests**, each validated by mutation (red under its broken ordering, green on both old and new code):

| Leg | Invariant | Pinning test | Mutation that turns it red |
|---|---|---|---|
| (1) | suppress before the dedup **record** | `suppressed_parallel_task_dispatch_jsonl_copy_survives_dedup_and_tracks` | hoist `recent_hook_tool_uses.insert` above the suppress call |
| (2) | dedup before task tracking | `jsonl_duplicate_task_end_inside_dedup_window_does_not_fire_cascade` | move `track_active_tasks` above the JSONL dedup-drop |

Leg (2) is the issue's stated hazard (a JSONL duplicate draining `active_tasks` → premature `cascade_exit`). Leg (1) was surfaced by review: a parallel **second** Task dispatch arrives as a hook `ActivityStart` while the first is in flight → suppressed as a leak; if it recorded its `tool_use_id` first, dedup would also drop its own JSONL copy (same parent `AgentId` + tuid) — the only transport left to track it — and the first Task's drain would cascade-exit the still-working subtree one Task early. Before this PR, that reorder passed all 1015 tests silently.

## Review

Multi-agent review (3 lenses — mechanical equivalence, correctness, architecture — findings adversarially verified, majority-vote for major+):

- **Equivalence: PROVEN** — token-level diff of old vs new; the only deltas: `match &event` → `match event` on the `&AgentEvent` param (identical binding modes), the early `return` → returned bool with the `Transport::Hook` gate short-circuiting at the call site, b1 `cascade_exit` relocated inside `track_active_tasks` at the same flow position, comments reworded.
- **Correctness:** no behavioral regression (statement-level comparison + mutation testing); both legs of the new ordering comment verified true against the actual code paths.
- **Architecture:** layering invariant respected; no doc-currency impact (internal shape only, no public surface). 1 minor (leg-1 unpinned) + 1 nit (pre-pass ordinal wording) — **both fixed in this PR**; 1 finding refuted.

## Verification

- Baseline 377/377 → after 379/379 (`pixtuoid-core`), green-to-green
- `just preflight` (lint → clippy → hack → test): **1016/1016**, exit 0
- Both pinning tests mutation-validated red→green

### DIFF
--- a/crates/pixtuoid-core/src/state/reducer.rs
+++ b/crates/pixtuoid-core/src/state/reducer.rs
@@ -439,13 +439,14 @@ impl Reducer {
         id: AgentId,
         now: SystemTime,
     ) -> bool {
-        let in_task = self.active_tasks.get(&id).is_some_and(|s| !s.is_empty());
+        let tasks = self.active_tasks.get(&id);
+        let in_task = tasks.is_some_and(|s| !s.is_empty());
         let suppress = match event {
             AgentEvent::ActivityStart { .. } => in_task,
             AgentEvent::ActivityEnd { tool_use_id, .. } => {
                 let is_task_self_end = tool_use_id
                     .as_ref()
-                    .is_some_and(|t| self.active_tasks.get(&id).is_some_and(|s| s.contains(t)));
+                    .is_some_and(|t| tasks.is_some_and(|s| s.contains(t)));
                 in_task && !is_task_self_end
             }
             _ => false,
@@ -463,13 +464,11 @@ impl Reducer {
             // work, so the gate resolved: restore Active(Delegating) instead
             // of leaving a stale "permission?" Waiting until the 60-min
             // stale-sweep. Then drop the spurious display update.
-            let task_tuid = self
-                .active_tasks
-                .get(&id)
-                .and_then(|s| s.iter().next())
-                .map(|t| Arc::<str>::from(t.as_str()));
             if let Some(slot) = scene.agents.get_mut(&id) {
                 if matches!(slot.state, ActivityState::Waiting { .. }) {
+                    let task_tuid = tasks
+                        .and_then(|s| s.iter().next())
+                        .map(|t| Arc::<str>::from(t.as_str()));
                     fsm::enter_delegating(slot, task_tuid, now);
                     self.gated_before_waiting.remove(&id);
                 }
--- a/crates/pixtuoid-core/src/state/reducer.rs
+++ b/crates/pixtuoid-core/src/state/reducer.rs
@@ -127,6 +127,19 @@ fn source_label_prefix(source: &str) -> &str {
         .unwrap_or(source)
 }
 
+/// Outcome flags from [`Reducer::track_active_tasks`], consumed by `apply`'s
+/// main event match.
+struct TaskTracking {
+    /// An `ActivityEnd` drained a tracked Task: the general ActivityEnd arm
+    /// must be skipped — otherwise it would redundantly re-arm
+    /// `pending_idle_at` or arm it while tasks are still in flight.
+    handled_by_task_tracking: bool,
+    /// An `ActivityStart` dispatched a Task (already applied as
+    /// Active(Delegating) by the pre-pass): the general ActivityStart arm
+    /// must be skipped.
+    handled_by_task_start: bool,
+}
+
 #[derive(Debug, Default)]
 pub struct Reducer {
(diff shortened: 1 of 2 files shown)


IvanWng97:
## xhigh multi-agent review (post-PR)

9 finder angles × adversarial 3-state verification → 35 candidates → 22 deduped → 19 survived (6 refuted) → 13 after same-mechanism merge.

**Verdict on the extraction itself: clean.** Mechanical equivalence re-proven token-by-token; no behavioral regression; layering invariant respected (correlation maps stay in `reducer.rs`).

**Fixed in this PR** (`ea77a36`, `ba3949f`):
- Both ordering pins now derive offsets from `HOOK_WINS_WINDOW` — previously a window retune ≤100ms turned leg-2 red with a misleading diagnosis, and ≤60ms left leg-1 passing **vacuously** under the exact mutation it pins
- `TaskTracking.handled_by_task_start` doc no longer claims Delegating was applied in the Task-before-SessionStart race (no slot → nothing applied)
- Leg-2 comment notes its guarded window is synthetic-narrow (a real CC JSONL Task END only exists at completion) — the pin freezes the refactor ordering, not the dedup keying
- `delegating_pair` + `#[track_caller] assert_delegating` helpers replace the 14th/15th scaffold copies and the 4th match-assert copy
- `suppress_subagent_leak` lookup consolidation as its own commit (3 same-key `active_tasks` probes → 1 binding; `task_tuid` alloc moved inside its only consumer's gate) so the extraction commit stays token-identical to main

**Pre-existing edges surfaced (NOT this PR's regressions — code moved verbatim) → filed:**
- #150 — hook-wins dedup is event-kind-blind: a hook Task START record can swallow the genuine JSONL Task self-END (permanent `active_tasks` leak; b1 disabled for the parent's session). Deterministic trace; needs a rare double failure to trigger
- #151 — `active_tasks` count drift under transport latency (b1 early-cascade on a suppressed parallel dispatch; late JSONL replay re-insert)
- #152 — stale `gated_before_waiting` entry when the Task self-END drains while parent is Waiting (plausible; harm needs tuid reuse CC doesn't do — investigate or wontfix)

Both ordering mutations re-validated red against the final code (each kills exactly its own pin, leaves the other green). `just preflight`: 1016/1016.

PR #147, opened 2026-06-07
perf: configure lld linker for linux targets in CI

IvanWng97:
This PR adds a workspace-level cargo configuration that uses the fast `lld` linker for Linux targets in CI/CD environments, significantly speeding up build and test runs. It also includes commented instructions for local macOS developers to configure `lld` or `zld`.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,22 @@
+# Use faster linkers for Linux targets in CI/CD environments.
+# On ubuntu-latest GHA runners, `lld` is pre-installed.
+[target.x86_64-unknown-linux-gnu]
+rustflags = ["-C", "link-arg=-fuse-ld=lld"]
+
+[target.aarch64-unknown-linux-gnu]
+rustflags = ["-C", "link-arg=-fuse-ld=lld"]
+
+# For local macOS developers:
+# To use a faster linker locally, install one of the following and uncomment:
+#
+# 1. Install lld via Homebrew: `brew install lld`
+# [target.aarch64-apple-darwin]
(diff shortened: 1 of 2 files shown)


IvanWng97:
Rebased onto main + one substantive amendment: **dropped the `[target.aarch64-unknown-linux-gnu]` section**.

That target is never linked natively in CI — it's built exclusively through `cross` Docker containers (release.yml build + deb matrices), whose images ship a gcc cross-toolchain **without lld** ([Dockerfile](https://github.com/cross-rs/cross/blob/main/docker/Dockerfile.aarch64-unknown-linux-gnu), [common.sh](https://github.com/cross-rs/cross/blob/main/docker/common.sh)). cargo reads the project's `.cargo/config.toml` inside the container, so the flag would have failed the link exactly as demonstrated in [cross-rs/cross#1200](https://github.com/cross-rs/cross/issues/1200) — at tag time, on the irreversible publish path. Zero CI benefit, release-breaking risk.

Also: removed the zld suggestion ([archived Feb 2023](https://github.com/michaeleisel/zld), author's note: "consider using lld instead"; Xcode 15+'s default linker is already fast), and documented the new Linux `lld` prerequisite in CONTRIBUTING.md.

Scope notes for reviewers:
- `cargo install` users are unaffected — `.cargo/` sits at the workspace root, outside every packaged crate.
- The coverage job gets no speedup either way: `cargo-llvm-cov` sets the `RUSTFLAGS` env var, which overrides config-file target rustflags entirely.
- First post-merge CI run rebuilds all caches once (rustflags change every fingerprint); steady state benefits code-change PRs that relink the ~24 test binaries.

PR #148, opened 2026-06-07
test: implement automated visual regression screenshot checking in CI

IvanWng97:
This PR adds a visual regression checking step in the CI smoke test. It builds a Python-based Pillow image diff script that compares the rendered dusk snapshot against a checked-in reference screenshot. It fails the build if the pixels differ by >0.01%, and uploads both the candidate snapshot and the red diff/delta image to GHA artifacts for easy visual inspection.


IvanWng97:
Rebased onto main + substantive fixes from a multi-agent review (all findings empirically executed against the real images, not just read):

**1. The check was a complete no-op.** `convert("RGBA")` + Pillow ≥10's `getbbox()` (defaults `alpha_only=True`) means the diff of two opaque images has an all-zero alpha band → `getbbox()` returns `None` for **any** two same-size images. A reference with its entire top-left quarter inverted passed with "SUCCESS: Images are visually identical". The green smoke run on this PR was not evidence of anything. Fixed: compare in RGB; comparator now tamper-verified (1-px corruption → exit 1 + red diff overlay).

**2. The baseline was timezone-baked.** `--now-hour` builds the clock via `chrono::Local`, so the epoch — and the 10-minute weather slot hash + city-twinkle phase — depends on the generating machine's TZ. The committed reference (PDT) renders **Clear**; CI's UTC render is **Windy** — 47.77% pixel divergence that the no-op comparator was hiding. Fixing #1 alone would have turned every PR red forever. Fixed: references render under `TZ=UTC` with `--weather` forced (explicit scene choice, not a hash accident).

**3. Threshold 0.01% → 0.** The render is byte-deterministic (verified run-to-run); ~157 allowed pixels would waive a 1–2 logical-px sprite regression — exactly what the gate exists to catch. The "float coordinate differences" rationale had a refuted premise.

**4. Single-source-of-truth restored.** `docs/images/` is owned by `gen-docs-images.py` (`just demo`); the references are now generated there. CI calls the new `just visual-check` recipe (params in one place, runnable locally). Workflow for intentional visual changes documented in CLAUDE.md + CONTRIBUTING: run `just demo`, commit the regenerated references in the same change.

**5. Enrichment (panel-recommended, capped at one extra frame):** a second baseline — `cyberpunk` at 03:00 — covers three otherwise-dark axes in one render: non-normal theme + `recolor_frame` RGB-equality substitution, night lighting, and `ThemeKind::Dark`-gated halos.

Refuted along the way: the "every version bump breaks the reference" worry — the PNG rasterizer paints text cells as identical 6×10 boxes regardless of glyph, so same-length bumps change 0 px (length changes are covered by the documented `just demo` workflow).

The CI run on this push is the first time the linux-render-vs-checked-in-reference equality is actually exercised; if cross-platform drift exists, this run will surface it honestly.

PR #149, opened 2026-06-07
feat: add automated entity and pet cropping to snapshot example

IvanWng97:
This PR adds automated entity, pet, and furniture cropping functionality to the snapshot example script.

### DIFF
--- a/.claude/skills/beautify-decoration/SKILL.md
+++ b/.claude/skills/beautify-decoration/SKILL.md
@@ -27,6 +27,9 @@ A repo-specific iteration loop for visually redesigning a decoration in `pixtuoi
 3. ./target/release/examples/snapshot --cols 192 --rows 80 /tmp/snap.png
    ↓
 4. .venv/bin/python3 scripts/crop-snapshot.py /tmp/snap.png --scale 3 -q <quadrant>
+   (or skip the quadrant guessing: snapshot --crop-furniture pantry|couch|vending|
+   printer|meeting|sofa|desk OR --crop-agent <label> renders a 40x24-cell window
+   already centered on the target — no Python step)
    ↓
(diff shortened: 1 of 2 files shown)


IvanWng97:
Rebased onto main + fixes from empirical review (every crop mode rendered and visually inspected):

**1. Coordinate-space bug — every crop landed in the top-left corner.** Layout waypoints / `PoseHistory` positions are *logical half-block buffer* pixels (1 px per cell across, 2 px per cell down — the very `buf_w`/`buf_h` this function feeds `compute_with_seed`), but the code divided by the PNG rasterization factors (8×16, which only apply later in `save_backend_as_png`). `--crop-furniture pantry` returned the wall display + windows. Fixed (`cell_x = p.x`, `cell_y = p.y / 2`); pantry / meeting / vending / agent crops verified centered.

**2. `--crop-pet` removed — it was unimplementable as wired.** `DrawCtx.last_pet_pos` is an output populated only when a pet is drawn; pets exist only on the renderer-driven `--gif` path (`--pets` has `requires = "gif"`), while the crop applies only to the static-PNG path. The flag could never produce a crop — only the "no visual position" warning. If pet close-ups are wanted, that's a renderer-path (gif) feature.

**3. Fail loudly on typos** (repo precedent: `--theme`/`--weather`): an unknown furniture name or agent label now bails listing the valid options / live scene labels, instead of warning and silently writing the full uncropped PNG. `--crop-agent` and `--crop-furniture` are now mutually exclusive at the clap level.

**4.** Six copy-pasted waypoint match arms collapsed; rustfmt failure from the original push fixed; sprite-iteration SKILL.md documents the new flags as the no-Python alternative to quadrant cropping.

PR #143, opened 2026-06-06
feat: native Windows support (#128)

IvanWng97:
Tracking PR for #128 — native Windows support.

## Scope (to be refined by the architecture spec)

- [ ] **Architecture spec** — Windows transport (UDS replacement), path strategy, CC/Codex-on-Windows wire facts (`docs/superpowers/specs/`, local)
- [ ] **Hook transport** — replace `std::os::unix::net::UnixStream` shim↔daemon channel with a Windows-capable equivalent; keep the 200ms / always-exit-0 contract
- [ ] **Paths** — socket/pipe path, config dir, CC `~/.claude/projects` + Codex sessions-root resolution on Windows
- [ ] **install targets** — Claude Code / Codex settings on Windows (path quoting, no symlink-resolve assumptions)
- [ ] **TUI on Windows Terminal** — crossterm/ratatui half-block rendering, console subsystem (no window flash)
- [ ] **CI** — Windows build + test job
- [ ] **release.yml** — `x86_64-pc-windows-msvc` (+ ARM64 if feasible), flat `.zip` layout, SmartScreen note in README
- [ ] **Docs** — Windows install snippet

Blockers called out in #128: UDS at `/tmp/pixtuoid-{uid}.sock`, POSIX hook-invocation assumptions, `.deb`-only hook packaging.

Closes #128.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #128, opened 2026-06-05
Add Windows support (build target + binary in releases of exe)


smdssz:
### What's the problem or use case?

Hi! Love the project — a pixel-art office for AI agents is a great idea. I'd like to run it on Windows but ran into a wall.

**What I found**

- The latest release (`v0.5.0`) ships binaries only for `aarch64-apple-darwin`, `x86_64-apple-darwin`, `x86_64-unknown-linux-musl`, and `aarch64-unknown-linux-gnu`. No Windows target.
- The `release.yml` workflow only produces Unix artifacts.
- No open issue or PR discusses Windows.

**Why it might already mostly work**

The workspace is pure Rust on cross-platform crates (`ratatui 0.30`, `crossterm 0.29`, `tokio 1`), so compilation on Windows should be feasible with a recent stable toolchain.

**Potential blockers I noticed**

1. The architecture diagram shows a Unix domain socket at `/tmp/pixtuoid-{uid}.sock` between the hook and the TUI. Windows supports UDS since Win10, but the path/scheme likely needs Windows-aware handling.
2. The `pixtuoid-hook` shim is invoked by agent CLIs — invocation semantics on PowerShell/cmd may differ from the current POSIX assumptions.
3. The hook packaging ships as `.deb` only; Windows users would need a `.zip` (or `winget`/`scoop` would be a nice bonus).

### Proposed solution

Please ship a proper Windows binary in the release. Specifically:

- **Two `.exe` binaries** in a single `.zip` asset (mirror the Unix layout):
  - `pixtuoid.exe` — the TUI binary (GUI subsystem is fine, but it must work when launched from Windows Terminal / `cmd` / PowerShell).
  - `pixtuoid-hook.exe` — the agent hook shim, invoked by Claude Code / Codex / etc. on session events.
- **Targets**: `x86_64-pc-windows-msvc` for both x64 and ARM64 (via the same MSVC build; ARM64 Windows runs x64 emulated for unsigned binaries, or include `aarch64-pc-windows-msvc` if feasible).
- **No console window flash**: the TUI binary should use the Windows console subsystem correctly (no extra console window popping up when launched from a shortcut). The hook shim should *not* be GUI subsystem, since it needs to be silent and CLI-invokable.
- **Code signing**: out of scope for now, but please document that Windows SmartScreen will warn on first run and how to bypass (`Unblock-File` or right-click → Properties → Unblock).
[...]

PR #153, opened 2026-06-07
feat: cross-platform pre-work for Windows support (#128)

IvanWng97:
PR 1 of the Windows-support arc (#128, tracking PR #143, spec approved 2026-06-06). **Zero behavior change on Unix** — every change is either a helper-equivalent refactor or a Windows-only-visible fix.

## What

- **`user_home()` helpers** (USERPROFILE-first on Windows, HOME on Unix, empty=unset): `pixtuoid-core/src/platform.rs` (returns `String`, per-platform fallback) + `pixtuoid/src/install/io.rs` (returns `Option<String>` — call sites keep their own fallbacks). All 10 production home-dir reads rewired (`claude_code`/`codex`/`antigravity` `default_paths` via `PathBuf::join`, `home_relative`, `config_path`, `resolve_pack_dir`, `embedded_pack::xdg_pack_dir`, `crash_log_path`, `log_file_path`, snapshot example). On native Windows, HOME is unset — these sites silently watched `/tmp/.claude/projects` / wrote config into CWD / hard-errored the log path.
- **`EXE_SUFFIX`** in the hook-binary sibling lookup (`hook_sibling_name()`).
- **`KeyEventKind::Press` guard** in the event loop — Windows delivers Press AND Release per keystroke; every key double-fired there (`p` paused then instantly unpaused). No-op on Unix.
- **Teardown ordering** (`teardown_terminal` + panic hook): `DisableMouseCapture` must run while raw mode is still on — on Windows it restores the input mode snapshotted at Enable time (raw-era), so the old order left the shell echo-less on exit. Order-indifferent on Unix.
- **6 symlink tests** gated `#[cfg(unix)]` (they call `std::os::unix::fs::symlink`).

One documented rare-path deviation: the no-HOME/no-XDG crash-log fallback moved from literal `/tmp` to `std::env::temp_dir()`.

## Verification

- `just preflight` green: lint → clippy → hack → **1025/1025 tests**.
- Each task TDD'd (red→green) + two-stage subagent review (spec compliance, then code quality).
- Windows branches of both helpers are pure functions unit-tested on macOS; live Windows verification comes with PR 3's CI job + the #128 community tester.

Part of #128 — next: PR 2 (workspace compiles for `x86_64-pc-windows-msvc`, named-pipe hook transport).

### DIFF
--- a/crates/pixtuoid-core/CLAUDE.md
+++ b/crates/pixtuoid-core/CLAUDE.md
@@ -55,6 +55,8 @@ src/
 │                       walk_arrived, speed_mult, pause_ms_for; constants: V_CRUISE_COMMUTE=0.36,
 │                       V_CRUISE_SNAPBACK=0.65, V_CRUISE_WANDER=0.25, WALK_ACCEL=6.5e-4,
 │                       WALK_ACCEL_SNAPBACK=2.0e-3, SPEED_MULT_MIN/MAX, PAUSE_MS_MIN/MAX
+├── platform.rs         cross-platform home-dir resolution (user_home(), USERPROFILE-first on
+│                       Windows — HOME is unset there and Git Bash's HOME is POSIX-form)
 ├── pose/               pure state→pose derivation + wander state machine (no terminal deps);
 │                       mod.rs (production) + tests.rs sibling.
 │                       ENTRY_ANIMATION_MS is demoted: not a duration knob, only the spawn-window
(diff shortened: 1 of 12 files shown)


IvanWng97:
Both findings addressed in 2700f21: the test now pins the per-platform literal (`pixtuoid-hook` / `pixtuoid-hook.exe`) instead of re-computing EXE_SUFFIX, and `platform.rs` has its layout entry in `crates/pixtuoid-core/CLAUDE.md`. (#155 rebased onto the fix.)

Issue #152, opened 2026-06-07
reducer: gated_before_waiting entry left stale when the Task self-END drains while the parent is Waiting


IvanWng97:
## Problem

When a parent is `Waiting` while delegating and the gate was recorded from the Task's own tuid (parent was Active(Delegating){tuid: task-T} when the permission `Notification` fired → `gated_before_waiting[parent] = task-T`), the Task self-END for task-T takes the task-tracking drain path (`handled_by_task_tracking`), which deliberately skips the main ActivityEnd arm — including its `resolves_wait` consult/clear of `gated_before_waiting`. The entry goes stale: task-T is gone from `active_tasks` but the gate still holds it.

## Assessment (honest framing)

Mechanism is real (verified by trace in PR #142's review — PLAUSIBLE verdict); harm is theoretical: the stale tuid can only false-match a later `ActivityEnd` carrying the SAME tuid, and CC tool_use_ids are unique, so no known trigger fires. The entry is also bounded — the next `ActivityStart` removes it, the next `Waiting` overwrites it, and slot eviction reclaims it (both eviction sites).

## Fix sketch / next step

Trace whether the skipped `resolves_wait` is intentional for this interleaving (the drain path's comment only covers the pending-idle half: "a Task drain must NOT arm the idle-resolve"). If intentional, clear the now-unreachable gate entry on the drain path for hygiene; if the trace finds no harmful interleaving at all, close as wontfix with the trace attached.

Surfaced by the PR #142 review.

Issue #151, opened 2026-06-07
reducer: active_tasks count drifts under transport latency (b1 early-cascade on suppressed parallel dispatch; late JSONL replay re-insert)


IvanWng97:
Two pre-existing edges of the same family — `active_tasks` tracks Tasks from whichever transport survives the pre-passes, and transport latency can make the count drift. Both surfaced by the PR #142 xhigh review (CONFIRMED by adversarial verification); both untouched by that PR (the code moved verbatim).

## A. Under-count: b1 early-cascade window on a suppressed parallel dispatch

A parallel SECOND Task dispatch arriving via hook is (correctly) suppressed as a leak and is tracked ONLY via its JSONL copy — that's the leg-1 ordering invariant `suppressed_parallel_task_dispatch_jsonl_copy_survives_dedup_and_tracks` pins. But the pin covers the happy path where the JSONL copy has already been delivered. During watcher latency (notify delay, or the 60s `scan_root` poll backstop after a missed event), `active_tasks` under-counts: if the FIRST Task's hook END drains the set inside that window, `set.is_empty()` fires one Task early → `cascade_exit` evicts the second Task's still-working subtree.

## B. Over-count: late JSONL replay re-inserts a drained tuid

A late JSONL replay of a Task START arriving >`HOOK_WINS_WINDOW` (500ms) after its hook copy (e.g. delivered by the 60s poll backstop) passes the window-deep dedup and re-inserts an already-drained tuid. The parent flips back to Active(Delegating) and its hook events are suppressed until the JSONL END in the same batch re-drains (redundant `cascade_exit` no-ops on the already-GC'd child). Transient one-batch mislabel; self-healing because START/END arrive in the same transcript read.

## Why deferred

Both need design thought about per-tuid bookkeeping (e.g. tracking suppressed-dispatch tuids as expected-from-JSONL, or per-tuid timestamps so a drain can't fire while a known dispatch is still unconfirmed), which touches the same load-bearing keying as #150 — they should be considered together. A is a rare race with a real (but self-limiting) eviction; B is cosmetic and self-healing.

Surfaced by the PR #142 review.

PR #154, opened 2026-06-07
fix(reducer): asymmetric hook-wins dedup — Task self-END can no longer be eaten by its dispatch record (#150)

### LINKED ISSUE #150: reducer: hook-wins dedup is event-kind-blind — a hook Task START record can swallow the genuine JSONL Task self-END (permanent active_tasks leak)
## Problem

`recent_hook_tool_uses` keys on `(AgentId, tool_use_id)` with no event-kind discrimination (`event_tool_use_id` extracts the tuid from both `ActivityStart` and `ActivityEnd`), so a hook Task **START** record dedup-drops a JSONL **END** carrying the same tuid inside `HOOK_WINS_WINDOW` (500ms).

Failure sequence (deterministic at the reducer API; verified by trace in PR #142's review):

1. Hook `PreToolUse` for a Task dispatch records `(parent, "task-T")` and inserts into `active_tasks`
2. The Task fails/returns in **<500ms** (instant dispatch error / trivially fast subagent)
3. The hook `PostToolUse` drops — within contract: the shim is best-effort, 200ms write timeout, exit-0
4. The JSONL `tool_result` END lands inside the window → dedup-dropped at the JSONL-drop `return`
5. `active_tasks[parent]` never drains; there is no per-tuid expiry — only whole-slot eviction recovers

Consequences, session-long for that parent: `in_task` stays true so **every** later hook ActivityStart/non-self-End is suppressed (parent degrades to JSONL-only updates); `set.is_empty()` can never fire again so the **b1 completion cascade is disabled for all of the parent's future delegations** (completed subagents linger to the 30-min sweep — the exact failure b1 exists to prevent); and the suppress block's Waiting-restore will false-clear the parent's own genuine permission `Waiting`.

The irony: JSONL is the documented fallback for dropped hooks, but the kind-blind dedup defeats the JSONL fallback for precisely the Task self-end. The only real-stream event the START→END cross-kind drop ever suppresses is a genuine completion (a real `tool_result` for an in-flight Task cannot exist 100ms after dispatch).

## Why deferred

The fix changes the load-bearing hook-wins keying — it needs its own ordering analysis (PR #142 pinned the current pre-pass order with two mutation-validated tests; any keying change must re-derive those pins), not a ride-along on a mechanical-move PR. Trigger needs a rare conjunction (sub-window Task + dropped hook END), so severity is low, but it is silent and permanent when it hits.

## Fix sketch

Either:
- **Exempt Task self-ends from the JSONL-END dedup drop**: skip the drop when the tuid is present in `active_tasks[id]` — mirroring `suppress_subagent_leak`'s `is_task_self_end` exemption (smallest change, preserves all other dedup behavior), or
- **Key the map on `(AgentId, tuid, event-kind)`** so a START record can never drop an END. Ride-along if the key shape changes anyway: the JSONL-drop check currently allocates `tuid.to_string()` per event just to build the tuple key for `contains_key` (no `Borrow` bridge for `(AgentId, &str)`).

Update the leg-2 ordering pin (`jsonl_duplicate_task_end_inside_dedup_window_does_not_fire_cascade`) alongside — its synthetic stream shape depends on the cross-kind drop.

Surfaced by the PR #142 xhigh review (4 finder angles independently, CONFIRMED by adversarial verification).

IvanWng97:
Closes #150.

## The bug

The hook-wins dedup map was **event-kind-blind**: keyed on `(AgentId, tool_use_id)`, with `event_tool_use_id` matching both `ActivityStart` and `ActivityEnd`. A hook Task START's record therefore dropped the genuine **JSONL END** for the same tuid inside `HOOK_WINS_WINDOW` (500ms). When the PostToolUse hook also dropped (best-effort shim, 200ms timeout, exit-0 — documented-real), that JSONL END was the *only* completion signal — eating it leaked `active_tasks[parent]` **permanently** (no per-tuid expiry): every later parent hook event suppressed as a subagent leak, b1 completion cascade disabled for all future delegations, and the suppress block's Waiting-restore could false-clear genuine permission prompts. JSONL is the documented fallback for dropped hooks; the kind-blind dedup defeated the fallback for exactly the event that needed it.

## The fix: asymmetric drop matrix, kind in the VALUE

`recent_hook_tool_uses: HashMap<(AgentId, String), (SystemTime, ToolEventKind)>` — drop a JSONL event unless `recorded == Start && event == End`:

| record \ JSONL event | Start | End |
|---|---|---|
| **Start** | drop | **pass** ← the fix |
| **End** | drop | drop |

Kind-in-the-**value** (not the key) is load-bearing: the hook End *overwrites* its tool's Start entry, so one End record covers the tool's whole lagged JSONL pair — a `(id, tuid, kind)` key would orphan the lagged JSONL START after the Start entry's expiry and regress exactly the path it claims to handle (two review verifiers empirically proved that mutant survives the old test timings; the pin now reaches it).

Chosen by a 3-agent adversarial design panel (advocate-per-option + risk auditor) over the surgical `active_tasks` exemption — unanimous, after the panel itself discovered the naive-symmetric-key regression.

## What this also fixes (same mechanism)

- ordinary-tool stuck-Active ghost: sub-window tool + dropped hook END → JSONL END now arms the idle debounce instead of being swallowed (previously: mislabeled until next event, or wrongful 10-min stale-sweep of a live CC session that cannot resurrect)
- ghosted permission Waiting whose gated tool's END dropped → `resolves_wait` now fires

No-op for Codex (tuids `None` on both transports, wire-verified) and Antigravity (no hook transport).

## Behavior-preservation notes

- Starts byte-for-byte unchanged; only the END drop narrows
- Retires `jsonl_duplicate_task_end_inside_dedup_window_does_not_fire_cascade` — its synthetic premise (a real JSONL Task END mid-delegation) cannot occur on the wire (`decode_cc_line` emits `ActivityEnd` only from `tool_result` blocks, written at completion); the behavior it froze IS the bug. Its #90 ordering duty is taken over by a new wire-plausible pin (below)
- Leg-1 ordering pin (#90, suppress-before-record) is same-kind — passes unchanged, re-mutation-validated

## Verification

- TDD: both leak repros red pre-fix → green (`jsonl_task_self_end_drains_when_hook_end_drops`, `jsonl_ordinary_tool_end_drains_when_hook_end_drops`)
- **4 mutants killed**: kind-blind always-drop (both repros red), symmetric value-matching (`late_batched…` red), kind-in-the-KEY (verified empirically by two independent review agents at the fixed offsets), pre-pass order swap (`jsonl_task_start_duplicate_does_not_clobber_waiting_parent` red)
- Multi-agent review (11 agents, adversarial 3-state verification): full interleaving-matrix trace — no cell worse than pre-fix; all five wire claims verified against `claude_code.rs`/`codex.rs`/`antigravity.rs`/shim/captured fixtures; 2 findings confirmed → fixed in `7dbbd4b`; 1 refuted
- `just preflight`: **1024/1024**, exit 0

#151 (active_tasks drift under transport latency) and #152 remain open — same design space, separate PR.


IvanWng97:
## Live repro (real transports, not unit tests)

Triggered the bug against a running daemon: real `pixtuoid run --headless`, events fired through the real `pixtuoid-hook` shim over the real Unix socket, JSONL appended to a watched transcript (real FSEvents). Script: hook `SessionStart` → hook `PreToolUse` Agent dispatch (`task-LAB`) → **PostToolUse never sent** (the documented best-effort drop) → JSONL `tool_result` END appended +50ms (inside `HOOK_WINS_WINDOW`) → ordinary hook `PreToolUse` Bash +1s.

**pre-fix (main):**
```
agents=[cc·demo-150@0:idle]
agents=[cc·150@0:active(Delegating)]     ← stuck here forever; Bash Pre suppressed
```

**post-fix (this PR), same script byte-for-byte:**
```
agents=[cc·demo-150@0:idle]
agents=[cc·150@0:active(Delegating)]
agents=[cc·150@0:active(Bash: ls)]       ← END drained active_tasks; suppression released
```

The post-fix run is also the control: it proves the watcher delivered the END inside the window, so the pre-fix stuck state is attributable only to the kind-blind swallow.

Lab note: the repro initially split into two agents because FSEvents reports canonical `/private/tmp` paths while the synthetic hook payloads said `/tmp` — the transports coalesce on exact `transcript_path` string equality. Real CC is immune (`~/.claude/projects` has no symlink), but it confirms how load-bearing that path parity is.

PR #155, opened 2026-06-07
feat: workspace compiles for x86_64-pc-windows-msvc — named-pipe hook transport (#128)

IvanWng97:
PR 2 of the Windows-support arc (#128, tracking PR #143, spec approved 2026-06-06). **Stacks on #153** — review the last 7 commits until that merges; GitHub will collapse to the delta after.

The whole workspace now **compiles for `x86_64-pc-windows-msvc`** (libs, bins, examples, every test target), enforced permanently by a new ubuntu CI job. Unix runtime behavior byte-identical (Unix arms are verbatim moves; suite green throughout).

## What

- **Shim transport seam** (`pixtuoid-hook/src/transport.rs`, the only platform split in the shim): `cfg(unix)` arm = today's UnixStream + 200ms write timeout moved verbatim; `cfg(windows)` arm = named-pipe client via `fs::OpenOptions` (CreateFile — **zero new shipped deps**) + a 200ms watchdog thread bounding the *whole* connect+retry+write phase (strictly stronger than the Unix write-only bound; `Builder::spawn` so thread exhaustion degrades to a dropped event, never an abort), `ERROR_PIPE_BUSY` retry. Invariant — never block CC, always exit 0 — holds on every path.
- **Endpoint scheme**: `\\.\pipe\pixtuoid-{USERNAME}` on Windows (name = namespacing; the DACL is the security boundary — owner-only SDDL hardening lands in PR 3 with the Windows runner). `PIXTUOID_SOCKET`/`--socket` carry a pipe name verbatim; shim and daemon branches mirrored, pinned per-cfg by `socket_path_parity.rs` (same `#[path]` include, no dep edge).
- **Daemon listener split** (`pixtuoid-core/src/source/hook/{mod,unix,windows}.rs`): `handle_conn` generified over `AsyncRead` → **one framing/decode path feeds both transports** (pinned by a new platform-neutral `tokio::io::duplex` unit test); `unix.rs` = byte-move; `windows.rs` = tokio first-party named-pipe accept loop (`first_pipe_instance` squat detection, `reject_remote_clients` pinned on both builders, byte mode, 1MiB in-buffer ≥ the shim's stdin cap, create-next-instance-before-handoff). Public API unchanged.
- **`libc`** → `[target.'cfg(unix)'.dependencies]` in both crates.
- **CI forcing function**: `just check-windows` (cargo check, no linking → ubuntu runner) + the `windows-check` job — cfg(windows) compile rot now fails every PR for pennies. `hook_socket.rs`/`shim.rs` integration tests gated `#![cfg(unix)]` (their pipe twins are PR 3).
- **CLAUDE.md currency** for the new module structure, same PR.

## Verification

- `just preflight` green: **1030/1030** (+5 new tests vs PR 1).
- `just check-windows` clean — full workspace, `--all-targets`.
- Every task TDD'd + two-stage subagent review; two holistic reviewers (correctness + silent-failure hunt) ran over the full range.
- Runtime behavior on real Windows is exercised starting PR 3 (Windows CI runs the suite + pipe test twins) and live-verified by the #128 community tester at PR 5.

Part of #128 — next: PR 3 (Windows runner runs the full suite + `hook_pipe.rs`/`shim_pipe.rs` twins + SDDL hardening).


IvanWng97:
All review findings addressed:
- **connect-error stale instance** + **parity branch 3**: already fixed in 5b71542 (the error arm recreates the instance — propagating if recreation fails too, rather than warn-and-proceed without one; the `default` fallback is parity-pinned).
- **Unsanitized USERNAME** → a7a6bd5: `.replace('\\', "-")` on BOTH parity sides + a new parity branch pinning `USERNAME=CORP\alice` → `\\.\pipe\pixtuoid-CORP-alice` + the shim unit twin.
- **Unbounded per-line read** → a7a6bd5: `handle_conn` now wraps the stream in `take(MAX_CONN_BYTES)` (2× the shim's 1MiB stdin cap) — bounded on both platforms, one shared code path.

PR #160, opened 2026-06-07
feat: one CC session = one sprite on Windows — normalize_path_key (#128)

IvanWng97:
PR 4 of the Windows-support arc (#128, tracking #143; spec §3). Small by design — the `/subagents/` half of path identity already landed in PR 3.

## What

- **`decoder::normalize_path_key()`**: identity on Unix (every existing AgentId/golden depends on byte-identity — full suite unchanged); on Windows `\`→`/` + casefold (CC emits backslash hook paths but mixes `\`/`/` forms of one file, and NTFS is case-insensitive). Pure core unit-tested cross-platform.
- Applied at **exactly the two AgentId keying sites**: the hook decoder's `transcript_path` (session_id UUID fallback untouched) and the watcher's `default_id_from_path`. Without this, the hook key and watcher key hash to two AgentIds on Windows → every CC session renders as **two sprites** and hook-wins dedup + permission-Waiting silently die.
- **End-to-end coalescing pin**: `hook_and_watcher_keys_coalesce_for_one_file` drives the REAL seams on both sides (public `decode_hook_payload` vs a real `JsonlWatcher` emitting `SessionStart` for the same file) — no inline re-simulation. Plus a cfg(windows) mixed-separator/case twin exercised on the `windows-test` job.
- **Windows-slug label pin**: `cc_derive_label`'s project-dir fallback verified to yield `cc·bar` from the `C--Users-foo-bar` slug shape (format upstream-verified; the real Windows hook-payload fixture lands once the #128 tester captures one at PR 5).
- CLAUDE.md sharp-edge entry (same-PR docs rule).

## Verification

- `just preflight` 1043/1043 (Unix byte-identical — the identity arm is load-bearing and unit-pinned)
- `just check-windows` clean; the `windows-test` required check runs the cfg(windows) pins on a real VM

Part of #128 — next: PR 5 (CC exec-form install + VT gate + release zips → first pre-release to the tester).


IvanWng97:
Security-review findings addressed in 3752832:
- **Finding 1 (HIGH)**: `walk_jsonl`'s `transcript_path_str` now folds through `normalize_path_key` at the chokepoint — per-line event ids match the SessionStart key on every platform (identity for Codex's lowercase-UUID extraction). New pin: `watcher_emits_session_start_then_activity_for_tool_use` now asserts SessionStart and per-line events share ONE AgentId.
- **Finding 2 (HIGH)**: was already fixed in 16d2b60 (the review raced the push) — `detect_parent_id` folds the rebuilt parent key; the cfg(windows) test asserts the normalized form via `default_id_from_path` itself.
- **Finding 3 (MEDIUM)**: docs now enumerate all **four** keying sites (CLAUDE.md sharp edge + the fn doc).

PR #203, opened 2026-06-09
fix(cc): derive agent identity from session UUID so subagent→parent links survive a worktree cwd-split

### LINKED ISSUE #200: Subagents render as roots when cwd changes mid-session (worktree split): parent transcript under a different project dir
## Symptom
With many subagents (e.g. a Workflow spawning ~20), they all appear as **first-level roots** in the agent dashboard (PR #199) instead of nested under their parent session — defeating the fold-the-workflow goal. The scope tree (liveness↑/cascade↓/b1) also no-ops for them.

## Root cause (verified against live `~/.claude/projects`)
When a session's **cwd changes mid-session** (e.g. entering a git worktree), CC splits the transcripts across two project dirs:
- **Main conversation transcript** under the *original* cwd's project dir: `…/projects/-Users-…-ascii-agent-nosync/<uuid>.jsonl` (exists).
- **Subagent transcripts** under the *worktree* cwd's project dir: `…/projects/-Users-…-agent-dashboard/<uuid>/subagents/[workflows/wf_…/]agent-*.jsonl`.

`detect_parent_id` (`source/jsonl.rs`) rebuilds the parent key from the **subagent's** path → `<worktree-project>/<uuid>.jsonl`, which **doesn't exist** (the real transcript is under the *main* project dir). `session_id` is filename-derived (`jsonl.rs:351`), so parent (`<uuid>`) and subagent (`agent-<id>`) don't share it either. Parent and child are keyed by full *path* → different project dirs → never match → orphan → dashboard's orphan-as-root rule renders each as a root.

Normal single-cwd sessions are fine: main transcript + `subagents/` share one project dir, `detect_parent_id` matches, subagents nest.

## The link that survives the split
Both the subagent and its parent carry the **same session UUID** — recoverable from the subagent two ways: the dir immediately before `subagents/` in its path, **and** its JSONL `sessionId` field (both `c0f7fb3f…`). The parent session's filename-derived `session_id` equals that UUID.

## Fix sketch
1. Carry the **parent session UUID** to the reducer — a new field on `AgentEvent::SessionStart` (or re-purpose the subagent's `session_id`), extracted in `walk_jsonl`.
2. **Reducer reconciliation**: when a subagent's path-derived `parent_id` resolves to no watched agent, match the parent UUID to the session-root slot whose UUID equals it, and re-point `parent_id`.

Touches a **workspace-wide contract** (`AgentEvent` → `agent_id()` + the four `reducer.rs`/`e2e.rs`/`transport/socket.rs`/`watcher.rs` test areas per CLAUDE.md "When refactoring") **plus the scope-tree reducer** (liveness/cascade/b1 sharp edges) **plus a new cwd-split fixture**. Own PR, full TDD.

Surfaced by PR #199 (agent dashboard). Carved out so #199 ships as the dashboard; this is the follow-up.

IvanWng97:
## Summary

Claude Code agent identity was derived from the transcript **file path**. Under a git-worktree cwd-split (the session's cwd changes mid-run), CC writes the **main** transcript under one `~/.claude/projects/<project-dir>/` and its **subagent** transcripts under a *different* one (the project-dir name is a pure function of cwd). The old `detect_parent_id` rebuilt the parent's key from the subagent's path → it landed under the wrong project dir → the link pointed at a phantom `AgentId` → every subagent rendered as a **top-level root** (a ~20-subagent workflow flooded the dashboard). Closes #200.

**Fix:** CC now keys identity on the **session UUID**, mirroring Codex. The UUID is recoverable cwd-independently from the transcript filename stem (`cc_id_from_path`, the sibling of `codex_id_from_path`) and, for the parent link, from the `<parent-uuid>` directory component (the segment before `subagents/`). The four CC keying sites flip in lockstep:

- **hook** → `IdKey::SessionId` (keys on `session_id`)
- **watcher first-sight** → `.with_id_deriver(cc_id_from_path)`
- **per-line** → `decode_cc_line` uses `cc_id_from_path`
- **parent link** → `detect_parent_id` returns `from_parts(source, <parent-uuid dir component>)`

`normalize_path_key` / `default_id_from_path` / `from_transcript_path` stay — Antigravity is now their sole production user (it remains path-keyed). The reducer / scope / dashboard layers are **id-agnostic** and unchanged: the fix is contained to the decode layer, and the parent link is now correct *at first-sight*, so the dashboard nests the subtree the moment the parent slot appears.

Commits: `cc_id_from_path` helper → lockstep flip + tests → docs → rustfmt → review-fix. Rebased onto current `main` (clean coexistence with #202's `\\?\` `normalize_path_key` hardening, which now serves Antigravity).

## Review

- Spec-compliance + code-quality review (2 agents) on the core change — both passed; one Minor (a redundant unit test) fixed.
- Adversarial review **could not break it**, verified against **2483 real CC subagent transcripts** on disk: hook `session_id` == filename stem == `cc_id_from_path`, and the `<parent-uuid>` dir component == the parent's id. Two theoretical-only nits (Windows-uppercase asymmetry; nested-`subagents`) are non-exploitable in real CC (ids are lowercase hex; exactly one `subagents` component) and match Codex's accepted assumptions.

## Test Plan

- [x] `just preflight` green on the rebased branch — **1158/1158** (lint → clippy → hack → test).
- [x] New regression: `detect_parent_id_survives_cwd_split` + `watcher_links_subagent_across_project_dirs` (parent + subagent under different project dirs → same parent link). Genuine RED on the old code.
- [x] Lockstep guard: `decode_cc_line_keys_agent_id_on_cc_id_from_path` + the rewritten `hook_and_watcher_keys_coalesce_for_one_file` pin hook == watcher == per-line id.
- [x] Conformance snapshot rebuilt; Contract-2 (hook+JSONL → one id) holds.
- [ ] **Live verify (human, `needs-human-verify`):** run a CC multi-agent workflow inside a git worktree (so the cwd-split actually occurs), open the Tab dashboard, and confirm the subagents nest under their parent instead of appearing as separate top-level roots. Windows-only test branches run on the `windows-test` CI job (can't execute on macOS).


IvanWng97:
Re: the MEDIUM on `cc_id_from_path` (empty-stem coalescing) — **WONTFIX, intentional**, after verifying the premise:

- **Unreachable in practice.** The JSONL watcher only processes files whose extension is `jsonl` (`x.jsonl` → stem `x`), so a watched path always has a non-empty stem. The empty branch is reachable only by a direct test call.
- **Deliberately tested.** `cc_id_from_path_empty_for_no_stem` passes `Path::new("")` and asserts `""` on purpose — the suggested `debug_assert!(!stem.is_empty())` would panic that test.
- **Mirrors Codex.** `codex_id_from_path` has the identical un-guarded empty-stem behavior; guarding only CC breaks the symmetry we deliberately adopted.

No exploit (the bot agrees), and the consequence is at most a dashboard mis-aggregation for a path that can't occur via the watcher. Leaving as-is.

Issue #85, opened 2026-06-03
Flaky: watcher_skips_recent_file_with_session_end_marker — 250ms rescan bypasses the session_end gate


IvanWng97:
## Symptom
`pixtuoid-core::jsonl_watcher::watcher_skips_recent_file_with_session_end_marker` fails intermittently under the **full parallel test suite** (`just preflight` / `cargo nextest run --workspace`). It passes 5/5 in isolation and usually passes in the full run, so it's a low-frequency, load-dependent flake — not a real regression. Most recently observed during the `feat/pet-naming` preflight (a change touching zero `pixtuoid-core` files).

## Root cause (the asymmetry, not just "timing")
`check_session_ended` is called in **exactly one place** — `initial_seed_walk` (`crates/pixtuoid-core/src/source/jsonl.rs:220`). When an ended file is seen there, its cursor is seeded at EOF (`:222`) so no `SessionStart` is emitted.

The **250ms post-startup rescan** (`run` → `scan_root` → `walk_jsonl`, `:122-124` → `:242` → `:270`) does **not** call `check_session_ended`. `walk_jsonl` reads from the file's cursor, defaulting to `0` (`:311`), so a file it sees for the first time is read from the top and its `session_start` line emits `SessionStart`.

Flake sequence:
1. Test writes `ended.jsonl` (with a `session_end` marker), then starts the watcher.
2. `initial_seed_root`'s `read_dir` **misses the freshly-created dir/file** (APFS metadata-propagation race, amplified by full-suite CPU load) → the file is never seeded; no cursor entry.
3. 250ms later `scan_root` finds it → `walk_jsonl` (cursor=0) reads from the top → emits `SessionStart`.
4. The test's fixed 300ms window catches that `SessionStart` → assertion fails.

## Why the earlier fix didn't hold
The 250ms rescan *was* the earlier fix — added to solve the **opposite** race (live files the initial `read_dir` missed never appearing). But it was wired to `walk_jsonl`, which has **no session_end/mtime gate**, so it re-introduces exactly the `SessionStart`-for-an-ended-file that `initial_seed_walk` is designed to suppress — whenever the initial `read_dir` loses the race. The two seeding paths have divergent session_end handling; prior work only tuned the initial-seed mtime window, never unified the rescan path.

## Fix options
[...]

PR #205, opened 2026-06-09
fix(watcher): register large transcripts on first-sight instead of waiting for their next write (#204)

### LINKED ISSUE #204: Large transcripts (>1MB) are invisible until their next write — oversized-tail guard suppresses first-sight SessionStart
## Problem

When `pixtuoid` attaches to a session whose transcript is already large (> 1 MB), that session's agent does **not** appear until the session writes its next line. With a delegating parent, its subagents (small transcripts) show up **flat as roots** because the parent has no slot to nest under; the tree only forms once the parent next writes.

Observed live: a ~7.4 MB main-session transcript stayed invisible on `pixtuoid` launch; subagents rendered as top-level roots; the parent + nesting appeared only when the main session became active again.

## Root cause

`source/jsonl.rs::walk_jsonl`, the oversized-pending-tail guard:

```rust
const MAX_PENDING_BYTES: u64 = 1 << 20; // 1 MB
...
if file_len - cursor_now > MAX_PENDING_BYTES {
    warn!("{} has > {} pending bytes with no newline; skipping to end", ...);
    cursors.insert(path, file_len);   // jump cursor to EOF
    return;                            // returns BEFORE the is_first SessionStart block
}
```

On first-sight of a large transcript the whole file is a >1 MB pending tail, so this branch jumps the cursor to EOF and `return`s **before** the `is_first` block that emits the synthesized `SessionStart` — so the agent never registers. It only registers on the *next* append (small tail < 1 MB -> guard skipped -> `is_first` fires).

Two sub-issues:
- The check is **size-only** despite the `"no newline"` wording — a valid multi-line 7.4 MB transcript is skipped identically to a corrupt no-newline blob.
- Skipping the backlog is fine; **suppressing agent registration is not**.

## Fix sketch (approach A — register-on-skip)

When the oversized guard fires on **first-sight**, still emit the first-sight `SessionStart` + `Rename` (register the agent) before seeding the cursor to EOF — skip replaying the backlog, but make the agent appear immediately. Derive `cwd`/label from a **bounded head read** (the first line carries `cwd`) rather than the whole file. Non-first-sight oversized appends keep the current skip-to-EOF behavior.

Fixes large **solo** sessions too (not just parents), and makes "a live subagent's parent is shown" fall out naturally (parent registers -> subagents nest). Distinct from #203 (identity/cwd-split). Discovered while live-verifying #203.

IvanWng97:
## Summary

When `pixtuoid` attaches to a session whose transcript is already large (> 1 MB), that agent stays **invisible until the session writes its next line**. A delegating parent (large transcript) then has no slot, so its subagents render as **flat top-level roots** — the tree only forms once the parent next writes. Observed live while verifying #203: a ~7.4 MB main-session transcript didn't show on launch; its subagents scattered as roots; the parent + nesting appeared only when the main session became active again. Closes #204.

**Root cause** — `source/jsonl.rs::walk_jsonl`, the oversized-pending-tail guard:

```rust
if file_len - cursor_now > MAX_PENDING_BYTES {   // 1 MB
    cursors.insert(path, file_len);   // jump cursor to EOF
    return;                            // returned BEFORE the first-sight SessionStart block
}
```

On first-sight the whole file is a >1 MB pending tail, so it jumped to EOF and returned **before** registering the agent. (The check is size-only despite the old `"no newline"` wording — a valid multi-line 7.4 MB transcript was skipped like a corrupt blob.)

**Fix (approach A — register-on-skip):** on **first-sight** of an oversized tail, still emit the synthesized `SessionStart` + `Rename` (register the agent) before seeding the cursor to EOF. The giant backlog is **not** replayed; `cwd`/label come from a **bounded ≤1 MB head read** (CC writes `cwd` on the first line), never the whole file. The first-sight emission is extracted into one `emit_first_sight` helper shared by the normal and oversized paths, so they register identically. Unchanged: stale/ended files stay gated by `should_seed_at_eof` (runs first — no #85 regression); mid-session oversized appends still skip to EOF without re-registering.

Fixes large **solo** sessions too (not just parents), and "a live subagent's parent is shown" falls out for free. Independent of #203 (identity/cwd-split) — different code region, branched off `main`.

## Test Plan

- [x] `just preflight` green — **1150/1150** (lint → clippy → hack → test).
- [x] New RED→GREEN test `watcher_registers_large_transcript_on_first_sight`: a recent >1 MB multi-line transcript now emits a `SessionStart` (cwd/label from the head) and does **not** replay the backlog (`activity_count` stays low). Genuine RED on old code.
- [x] `watcher_skips_oversized_pending_tail` reconciled (strengthened): first-sight oversized now registers + skips backlog; the post-skip valid line still decodes.
- [x] Stale-gating preserved — `watcher_skips_session_start_for_stale_files_on_startup` + `…_recent_file_with_session_end_marker` + `stale_file_emits_session_start_when_written_to` all pass.
- [ ] **Live verify (`needs-human-verify`):** rebuild + relaunch `pixtuoid run` against a session with a >1 MB transcript (e.g. this one) — the main session agent should appear on attach, with its subagents nested, instead of staying invisible until the next write.

### DIFF
--- a/crates/pixtuoid-core/src/source/jsonl.rs
+++ b/crates/pixtuoid-core/src/source/jsonl.rs
@@ -238,11 +238,13 @@ async fn walk_jsonl(path: &Path, decoders: SourceDecoders, ctx: &WatchCtx<'_>) {
         tx,
         window,
     } = *ctx;
+    // `derive_label` / `id_derive` are consumed inside `emit_first_sight` (off
+    // `decoders` directly); only the per-line decoder and the end-checker are
+    // used directly here.
     let SourceDecoders {
         decode_line,
-        derive_label,
         check_ended,
-        id_derive,
+        ..
     } = decoders;
     let meta = match tokio::fs::metadata(path).await {
         Ok(m) => m,
@@ -295,10 +297,23 @@ async fn walk_jsonl(path: &Path, decoders: SourceDecoders, ctx: &WatchCtx<'_>) {
     }
     if file_len - cursor_now > MAX_PENDING_BYTES {
         warn!(
-            "{} has > {} pending bytes with no newline; skipping to end",
+            "{} has > {} pending bytes; skipping backlog to end",
             path.display(),
             MAX_PENDING_BYTES
         );
+        // #204: on FIRST-sight of an oversized tail (a recent, not-yet-ended
+        // large session — stale/ended ones were already gated by
+        // should_seed_at_eof above), still REGISTER the agent before seeding the
+        // cursor to EOF. Otherwise a >1 MB transcript stays invisible until its
+        // next small append (a long session, or a delegating parent whose
+        // subagents then render as flat roots). The giant backlog is NOT
+        // replayed; cwd/label come from a BOUNDED head read (CC writes `cwd` on
+        // the first line), never the whole 7.4 MB file. A mid-session oversized
+        // append (`known`) is unchanged — skip to EOF, no re-registration.
+        if !known {
+            let head_cwd = read_head_cwd(path, MAX_PENDING_BYTES).await;
+            emit_first_sight(path, source, decoders, seen, tx, head_cwd).await;
+        }
         cursors.lock().await.insert(path.to_path_buf(), file_len);
         return;
     }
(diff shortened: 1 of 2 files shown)


IvanWng97:
Resolved in the latest commit: the oversized branch now seeds the cursor to EOF **before** the head-read + `emit_first_sight`, matching the normal tail-read path — so a concurrent rescan/notify sees `known` and can't re-enter. Verified the consequence was only ever a redundant head read (never a duplicate `SessionStart` — the `seen` guard ensures that); the reorder closes the window.

Issue #195, opened 2026-06-09
Codex-on-Windows: hook fails when pixtuoid-hook.exe path contains spaces


IvanWng97:
## Problem
The Windows Codex hook install (branch \`worktree-codex-windows\`, the upcoming Codex-on-Windows PR) writes the **bare** command form into \`~/.codex/config.toml\`:

\`\`\`
command = "C:\\Users\\ada\\.cargo\\bin\\pixtuoid-hook.exe --source codex"
\`\`\`

Codex spawns hooks via \`Command::new(cmd.exe).arg("/C").arg(&handler.command)\` (verified in codex-rs \`hooks/src/engine/command_runner.rs\` @ rust-v0.135.0 — regular \`.arg()\`, not \`raw_arg\`). That means:
- We **cannot quote** the path: \`Command::arg\` escapes an embedded \`"\` to \`\"\`, which \`cmd.exe /C\` mangles → the hook silently never fires.
- The **bare** form is correct only when the exe path has **no spaces**. If \`%USERPROFILE%\` contains a space (a username with a space → \`C:\\Users\\John Smith\\.cargo\\bin\\...\`), \`cmd.exe\` splits the path at the first space and the hook fails to launch.
- Codex injects **no per-hook \`env\`** we could use to drop the \`--source codex\` arg (config-layer hooks get \`env: HashMap::new()\` — verified in \`discovery.rs\`), so we can't fall back to a quoted-path-with-no-args (which \`cmd /C\` rule-1 would preserve).

## Why deferred
Windows Codex support is shipped as **experimental**; the space-free install paths (\`%USERPROFILE%\\.cargo\\bin\`, npm prefix) are the common case and work. A spaced username is the uncovered edge.

## Fix sketch (pick one)
1. **DOS 8.3 short path** at install time: resolve the exe to its short form via \`GetShortPathNameW\` (space-free, e.g. \`C:\\Users\\JOHNSM~1\\...\\PIXTUO~1.EXE\`), then emit the bare form. Caveat: 8dot3name can be disabled per-volume (\`fsutil 8dot3name\`), in which case the API returns the long path unchanged.
2. **Install-time warning**: detect a space in the resolved hook path on Windows and warn the user (surface via the existing install warning sink) that Codex hooks may not fire and to install to a space-free location.
3. **Upstream**: file a codex issue asking for a hook \`args\` array or a config-settable per-hook \`env\` (either makes a spaced path trivially robust).

[...]

PR #206, opened 2026-06-09
refactor(install,platform): consolidate cross-platform logic — one hook-command seam + one home/config resolver

IvanWng97:
## Summary

Behavior-preserving consolidation of the unix/windows split, from a workflow that mapped every platform-conditional site + mined the past Windows bug tickets (#195/#198/#201/#202…). The codebase was already ~80% on the right patterns (the `source/hook/` sibling-module split + the "inject `cfg!(windows)` into a pure core fn" idiom); this targets the two spots that genuinely deviated. No new dependencies; `pixtuoid-core` stays terminal-dep-free.

### A — one hook-command seam (kills per-CLI cmd-quoting duplication)
`codex::hook_command` and `reasonix::hook_command` were byte-identical per-OS-fork twins (Unix env-prefix vs Windows bare form) differing only in the source name — every new cmd.exe-shelling CLI re-paid the fork (the recurring **Class A** bug from #195/#198).
- `io::shell_hook_command(path, source)` is now the single OS fork; codex/reasonix are one-liners (the `#[cfg(unix)] use shell_single_quote` warts gone).
- Quoting primitives moved into `install/hook_cmd/{mod,unix,windows}.rs`, mirroring the gold-standard `source/hook/` split — `unix.rs` (POSIX single-quote), `windows.rs` (cmd-safety guard + DOS-8.3 FFI), with the injected-`short_path` pure core kept so it unit-tests on macOS.
- **Claude untouched** (exec-form, opposite strategy, already clean).

### B — one home/config resolver
- The two USERPROFILE-vs-HOME resolvers (core `platform.rs` + `install/io.rs`) could drift (**Class C**, #168/PR#153). The rule now lives once in `platform.rs::resolve_user_home_opt`; `install::io::user_home` is a thin adapter keeping its `Option`/empty-is-unset contract (tests unchanged).
- `reasonix::user_config_dir` was a runtime `cfg!(target_os)` if-else whose macOS arm couldn't be tested on Linux → extracted to a pure `platform::resolve_user_config_dir(os, appdata, xdg, home)` (OS + env injected at one call site), with all three OS arms unit-tested on any host.

API change to `pixtuoid-core` is **additive only** (`pub mod platform` + 3 new `pub fn`s; no item removed or changed).

## Test Plan
- [x] `just preflight` green — **1166 tests**, lint/clippy/`hack` (feature-powerset, 6/6).
- [x] Behavior-preserving, pinned by the **unchanged** byte-string tests: codex/reasonix hook-command (`PIXTUOID_SOURCE=… '…'` + Windows bare/8.3), `io::user_home` Option/empty semantics, reasonix `user_config_dir` per-OS dirs.
- [x] New pure-fn tests cover the macOS/Windows/Linux arms of `resolve_user_config_dir` + `resolve_user_home_opt` on this host.
- [x] Windows cross-lint (`x86_64-pc-windows-msvc -D warnings`, both crates) clean.
- [ ] CI semver (additive `pixtuoid-core` API — expect pass) + windows-test.

Discovered + scoped via the cross-platform rethink workflow; verdicts it produced and this PR honors: **keep** the `\\`→`/`+lowercase AgentId-key munge (it's a key string, not an fs path; `dunce`/`std::path` can't replace it), **don't** merge the socket/pipe parity twins (invariant #5), and use `cfg` (not `std::env::consts::OS`) for behavior since binaries are per-target.


IvanWng97:
Fixed in the latest commit. Added an identifier-allowlist guard on the Unix arm of `shell_hook_command` (`source` must be `[A-Za-z0-9_-]`), mirroring the Windows arm's guard — used an allowlist rather than a metachar denylist so it can't miss a character. The path is already single-quoted, so only `source` needed it. Valid sources keep the byte-identical `PIXTUOID_SOURCE=<src> '<path>'` form (pinned by a new test), and a shell-unsafe source now bails. Also repointed 5 stale `io::windows_bare_hook_command` doc refs to `hook_cmd::` (caught by the two-agent review).

Issue #209, opened 2026-06-09
Type-design: newtype the desk-index / floor-index domain quantities (global vs floor-local)


IvanWng97:
From the whole-codebase review (2026-06-09), finding #16 — deferred as the one large-effort item out of the `chore/review-followups-2026-06-09` bundle (every other confirmed finding was fixed in that PR).

## Problem
`AgentSlot.desk_index` is a raw `usize` whose 5-line doc comment warns it is a **GLOBAL** index and that indexing `layout.home_desks` (which wants a **FLOOR-LOCAL** index) is a bug. `floor_idx` shares the same `usize` type for a different index space; `active_ms` / `tool_call_count` are also bare integers. The global→floor-local remap in `build_floor_scene` is an unguarded subtraction. The invariant ("don't mix the two index spaces") is enforced by prose, not the type system.

No current bug — but it's a latent footgun: a future edit that passes a global index where a floor-local one is expected compiles fine and mis-seats agents.

## Why deferred
A faithful rollout introduces `GlobalDeskIndex(usize)` / `FloorLocalDeskIndex(usize)` newtypes and threads them through `pixtuoid-core::state` AND the binary's `tui/floor.rs` layout code — a cross-crate, many-call-site change (effort: large). Out of scope for the small/surgical review-followups bundle.

## Fix sketch
- Introduce `GlobalDeskIndex(usize)` and `FloorLocalDeskIndex(usize)` newtypes in `pixtuoid-core::state`.
- `SceneState::floor_local_desk(...)` returns `FloorLocalDeskIndex`; `Layout::home_desks` lookups accept only `FloorLocalDeskIndex`. The global→floor-local bridge (`build_floor_scene`'s remap) becomes the **only** legal conversion, type-checked.
- Optionally wrap `active_ms` (a `Duration`-ish ms count) and `tool_call_count` while in the area, but the desk/floor index split is the load-bearing part.

Location: `crates/pixtuoid-core/src/state/mod.rs` (`AgentSlot.desk_index` / `floor_idx`), `crates/pixtuoid/src/tui/floor.rs` (`build_floor_scene`).

PR #210, opened 2026-06-09
Review-followups: fix the whole-codebase review's confirmed findings (1 medium + lows + tests)

### LINKED ISSUE #211: tui: footer width/padding should measure display columns, not bytes (#12 — needs baseline regen)
From the whole-codebase review (2026-06-09), finding #12. Implemented then **dropped from PR #210** (reverted in f2831e7) because it requires regenerating the entire visual-baseline corpus — better to fold into a future PR that already changes the office visuals.

## Problem
`hud.rs::status_segments` selects the footer tier (full→medium→min) and computes the right-flush padding using **byte length** (`.len()`). The footer carries single-column multi-byte glyphs — `·` (separator), `×` (tool count, e.g. `Bash×2`), `↑↓` (floor indicator) — each 2–3 UTF-8 bytes but 1 display column. So byte length over-counts width → the padding is short → `[q]uit` doesn't reach the right edge (off by the multi-byte-glyph count). The source-warning branch in the same fn already correctly uses `chars().count()`.

Cosmetic only (never overflows; just a few columns short). Verified by both review agents as a real, low-severity correctness issue.

## Fix (trivial, ~2 lines)
In `status_segments`: `let q = quit.chars().count();` and `let stats_len: usize = tier.iter().map(|(s, _)| s.chars().count()).sum();` — matching the warning branch. (Optionally `unicode-width` if 2-column glyphs ever appear, but the current glyphs are all 1-column.)

## Why it needs care (the reason it was deferred)
Changing the rendered footer padding shifts the bottom status row of **every full-frame still** (screenshot, dashboard, reference-\*, day, night, and the whole theme×weather matrix — ~20 PNGs in `docs/images/` + `site/public/demos/`). The `smoke` job's `gen-check` pixel-diffs these at threshold 0.0%, so the fix MUST be landed together with a `just gen` regeneration of all baselines (per the documented "intentional look change → regenerate" workflow in CLAUDE.md). Doing that regen standalone for a cosmetic fix is disproportionate churn — so it should ride a PR that's already regenerating the office visuals.

## When to do it
Bundle with the next intentional visual/office change (or a dedicated `just gen` refresh). Re-add the column-width fix + its test (`status_segments_pads_to_full_column_width_with_multibyte_glyphs`) + the tui/CLAUDE.md footer doc note, run `just gen`, commit the regenerated `docs/images/` + `site/public/demos/`.

IvanWng97:
## Review-followups: fix the confirmed findings from the whole-codebase review

Acts on the 2026-06-09 multi-agent whole-codebase review (16 reviewers → adversarial design-intent + reproduction verify → synthesis), which surfaced **0 critical / 0 high**, 1 medium, 21 low confirmed findings (26 candidates refuted against documented sharp edges). This PR fixes the medium + the trivial/small correctness, security, and validation lows, each with a regression test.

Closes #211.

### The two that matter
- **🟠 reachable per-frame panic** — `tui/widgets/tooltip.rs`: the hover/disambiguation label sliced `&session_id[..4]` guarded only by a **byte** check. A Reasonix `session_id` IS the raw cwd path, so byte 4 landing inside a multi-byte UTF-8 codepoint (`/naïveté/app`) panicked the render loop when two agents shared a label. Now char-safe (`chars().take(4)`).
- **🟠 terminal-escape injection** — `runtime/`: the headless `summarize` fed untrusted label / tool-detail / Notification reason (uncapped) straight to `println!`. A crafted ANSI/OSC escape reached the user's terminal. Now control chars are stripped (`sanitize_line`).

### The rest (low)
| # | area | fix |
|---|------|-----|
| 12 | `tui/widgets/hud.rs` | footer tier/padding measured in display columns, not bytes (·/×/↑↓) — `[q]uit` right-flushes to the edge. Visual baselines regenerated (`just gen`). |
| 4 | `config.rs` | `update_config` fsyncs the tmp file before the atomic rename |
| 6 | `install/{claude,reasonix}.rs` | refuse a valid-JSON-but-non-object settings doc (was silently coerced to `{}`); codex is TOML so immune |
| 3 | `state/reducer.rs` | resurrect-in-place folds the in-flight Active span into `active_ms` (via `fsm::resurrect_in_place`) |
| 5 | `tui/tui_renderer/mod.rs` | zero `last_popup_scale` on a footer-only `Ok(None)` frame |
| 7 | `tui/tui_renderer/mod.rs` | evict per-agent `MotionState` across **all** floors, not just the current one |
| 10 | `sprite/format.rs` | reject non-hex palette bytes before `from_str_radix` (which accepts `+`) |
| 11 | `sprite/format.rs` | final-frame parse error carries line context like in-loop frames |
| 9 | `source/antigravity.rs` | skip a negative `step_index` (would mint an unmatchable `ag--1-0` id) |
| 13 | `sprite/mod.rs` | `RgbBuffer::get/put` gain a bounds `debug_assert` (release unchanged) |
| 8 | `source/jsonl.rs` | the known-file oversized-skip branch tail-scans for a buried `/exit` SessionEnd |

Test gaps added: resurrect folds active_ms (#3); an exiting **subagent** does not resurrect (#19); a capacity-dropped unknown-cwd session consumes no ghost ordinal (#22); motion eviction on a non-current floor (#20).

### Deliberately deferred (out of scope for this PR)
- **#16** desk-index/floor-index newtypes — large cross-crate change → **#209**.
- **#14** `Transport #[non_exhaustive]` — correct forward-compat, but it's a one-time breaking change and `just semver` baselines against crates.io, so landing it outside a release would red `semver-checks` on *every* PR until the next release (masking new breaks meanwhile). Deferred to land WITH the next `just bump` → **#212**.
- **#21** sweep_stale pass-2 write-once skip — un-pinnable (`mark_exiting` is already write-once, so removing the skip changes no observable state → a regression test can't fail-first). Documented, not a vacuous test.

### Verification
`just preflight` green (fmt / machete / deny / clippy / hack / full test suite, 1196/1196); `gen-check` reports every committed artifact in sync (24 footer-affected stills regenerated for #12; non-deterministic demo videos are presence-only and untouched); 2-agent diff review (correctness **APPROVE** + adversarial **ship-it**). All CI checks green. AI-authored — `needs-human-verify`.


IvanWng97:
### CI note — the single `semver-checks` red is expected & intentional (kept on purpose)

This PR adds `#[non_exhaustive]` to `pixtuoid-core`'s `Transport` enum (review finding #14), making it consistent with its event-channel siblings `AgentEvent` / `ToolDetail` / `SourceDeath` (all already `#[non_exhaustive]`) so that adding a future transport (e.g. the documented v2 daemon split) is a **non-breaking** change.

Adding `#[non_exhaustive]` to an *existing* public enum is itself a one-time breaking change, so `cargo-semver-checks` correctly reports **"requires new major version" (195 pass / 1 fail)**. This is the tool working as intended, not a regression:

- `semver-checks` is **non-required** (outside the enforced gate set); every other check is green.
- The break is **consumer-free** — `pixtuoid-core` has no external crates.io consumers, and the only in-workspace consumer (`pixtuoid`) uses `Transport` via `==`, never an exhaustive `match`, so nothing actually breaks. (Both review agents grepped this and confirmed.)
- We pay this break **now** — pre-1.0 with no consumers is the cheapest possible moment; deferring it to "once there are consumers" would make it a genuinely painful break. The next `just bump` (a minor: 0.6.x → 0.7.0) releases it and `semver-checks` returns green.

Kept deliberately under the project principle: *make the cheap one-time forward-compat break now, not the painful one later.*

IvanWng97:
### Update — dropped #12 + #14 to keep CI clean now *and* on future PRs

Both findings are correct, but each reds a **non-required** gate that would stay red beyond this PR — and a perpetually-red gate can't catch the next real regression. So they're deferred to the moment they land cleanly:

- **#14 `Transport #[non_exhaustive]`** → `just semver` baselines against crates.io 0.6.1 (no `--baseline-rev`), so the one-time break reds `semver-checks` on *every* PR until the next release. Moved to **#212**, to land with the next `just bump` (the break releases with 0.7.0).
- **#12 footer column-width** → shifts the bottom status row of ~20 full-frame visual baselines (`smoke`/`gen-check`). Moved to **#211**, to ride a future visual-change PR where `just gen` regenerates the baselines anyway. The tooltip panic fix (#1, separate file) is unaffected and stays.

Reverted in f2831e7. `gen-check` now reports every artifact in sync; all required checks green.

IvanWng97:
Update — **#12 restored** (it's a correct fix; a correct change is worth the visual-baseline regeneration). Reinstated the column-width footer + regenerated the 24 footer-affected stills via `just gen`; `gen-check` reports every artifact in sync. Closes #211. (#14 still deferred to the next release bump, #212.) All CI green.

PR #236, opened 2026-06-11
fix: env-contract + install + signal-teardown hardening

IvanWng97:
## What

Fixes the 7 config/install/runtime findings from the 2026-06-10 whole-codebase review.

1. **empty `XDG_CONFIG_HOME`** — treated as unset (was: CWD-relative `./pixtuoid/config.toml`, settings silently 'reset' + orphan configs scattered per launch dir); same fix applied to the two `XDG_STATE_HOME` log-path sites.
2. **`PIXTUOID_HOOK` bypass** — the env override now flows through the SAME absolutize-and-warn arm as `--hook-path` (drive-relative bail, cwd-unreadable hard error, nonexistence warning) instead of being embedded verbatim; and it now **embeds like the flag** (review round: `resolve_hook_binary_from` returns `(path, explicit)` — writing the bare PATH name discarded the user's explicit binary choice on Claude/Unix).
3. **lock-target re-resolution** — `ConfigLock` gained `read()/backup_once()/remove_backup()` pinned to the lock-time symlink resolution; install/uninstall/update_config can no longer split a round across two files on a symlink retarget. Correct by construction + pinned by a retarget-while-locked test.
4. **drive-relative `C:foo.exe`** — hard actionable error (the cwd-join silently no-opped); `is_drive_relative` verified against the full Windows `Prefix` taxonomy (UNC/verbatim/8.3 from #201/#202 unaffected).
5. **`CMD_UNSAFE` delimiters** — `tab ; , =` added (cmd.exe first-token delimiters, legal in filenames); 8.3-substitute-or-loud-reject contract pinned for all four.
6. **TUI signal teardown** — SIGINT/SIGTERM now ride the frame-pacing select and exit through the existing `teardown_terminal` path; an external kill previously left the terminal in altscreen + raw mode + mouse reporting.
7. **headless `ctrl_c` Err arm** — a failed signal-handler registration now logs and disarms (pending-future swap) instead of silently exiting 0; new `headless_loop_with_signal` seam + 2 paused-clock tests.

Review round also: the three locate-failure tests are env-immune again (ambient `PIXTUOID_HOOK` inverted their assertions — reproduced, then verified green with the var exported), empty/whitespace `--hook-path` is a hard clap error, and ONE emptiness semantics workspace-wide (`io::nonempty(_env)` + core `platform.rs` aligned to trim — `XDG_CONFIG_HOME="   "` can no longer be unset for the app config but set for the CLI config-dir resolution).

## Tests

1351/1351 (run WITH `PIXTUOID_HOOK` exported) + full preflight green; Windows cross-lint clean.

## Review

2 independent agents, both **APPROVE-WITH-NITS**; all actionable findings applied (the two reviewers independently converged on the hermeticity + embed-semantics issues).


IvanWng97:
Bot finding addressed in 193d74f: `update_config` now reads through `ConfigLock::read` (the full fix, not the rename mitigation) — missing-file handling rides `lock.read()`'s `""` contract into the existing empty-document arm, so the NotFound branch collapses cleanly. 772/772 + clippy green.

Issue #246, opened 2026-06-11
Codex multi-turn child fires SubagentStop per TURN — we end it after turn 1, re-registration loses the parent link


IvanWng97:
## Problem (from the 2026-06-11 upstream-alignment research, codex-rs @ main)

`codex-rs/core`'s `run_turn_stop_hooks` fires SubagentStop for every ThreadSpawn child **at each turn end**, and SubagentStart fires only on thread STARTUP — so a multi-turn child (parent `send_input`) emits Start, Stop, Stop, Stop... We decode the first Stop as SessionEnd: the child exits, then re-registers on its next rollout append (seen un-claim self-heal) but **parentless** (no second Start ever fires; the rollout first-sight is flat/parentless).

Related upstream facts from the same read: a RESUMED child thread fires NO Start at all (Stop-only is a designed code path, not just reorder); CC's bg agents likewise get Stop without Start (anthropics/claude-code#44075, closed not-planned); and 34.7% of real-world SubagentStops are orphans (anthropics/claude-code#59719).

## Why deferred

Needs a small design: the child→parent binding must survive the SessionEnd so a re-registration re-links. Interacts with the #244 ended-as-child tombstone design — likely the same reducer-private map (child id → parent id, TTL'd) serves both.

## Fix sketch

Keep a TTL'd child→parent correlation (written on SubagentStart decode / first parented registration, NOT evicted on SessionEnd). The SessionStart arm consults it to re-link a parentless re-registration of a known child id. Pin with a multi-turn Codex scenario (Start, Stop, rollout-append re-register, assert parent restored).

IvanWng97:
Status update from the child-ledger PR's review (upstream-verified against codex-rs hook_runtime.rs @ main, 2026-06-11):

**Half of this issue ships in the child-ledger PR** — the reducer-private child→parent ledger re-links any re-registration that OCCURS (flat first-sights, post-un-claim revivals), so the 'loses the parent link' half is fixed.

[...]

Issue #242, opened 2026-06-11
Subagent Stop-before-Start hook reorder mints a sweep-cleared phantom (shared CC + Codex hole)


IvanWng97:
## Problem

Hook deliveries ride per-connection tasks and can reorder. For a sub-TTL-lived subagent, `SubagentStop` can be decoded BEFORE its `SubagentStart`: the unknown-id SessionEnd lands a tombstone (reducer SessionEnd pre-pass), but the tombstone gates only Activity-SYNTHESIS — the main SessionStart arm still registers the late-arriving Start, minting a slot whose End already passed. It lingers to the stale sweep.

Found during the issue #241 implementation's correctness review (confidence 80 on mechanism, 40 on real-world occurrence). **Pre-existing and symmetric: the Codex SubagentStart/Stop twin on main has the identical hole** — this is NOT introduced by the CC wiring.

## Why deferred

Any fix must be subagent-scoped: a blanket SessionStart-tombstone check would break Reasonix's documented SessionEnd→SessionStart resurrect on the same id. Needs a small design decision (e.g. tombstone consultation only for child registrations). The window is tiny (Start and Stop are lifetime-separated for any agent living longer than connection-task jitter) and the stale sweeps backstop it.

## Fix sketch

In the reducer's SessionStart arm, consult the SessionEnd tombstone ONLY when the registering event carries a `parent_id` (i.e. a child registration): a tombstoned child Start degrades to a no-op (warn + skip) instead of registering. Pin with both CC and Codex reorder tests; verify Reasonix's same-id resurrect path stays untouched (it registers parentless).

PR #249, opened 2026-06-11
fix(core): child ledger — ended-as-child tombstone windows + revival re-link

### LINKED ISSUE #244: Subagent end-tombstone residual windows — Codex parentless first-sight bypass + known-id post-GC late discovery
## Problem

#242's child-SessionStart tombstone gate (parent_id-scoped, 5s TTL) closes the main reorder hole but three bounded residual windows remain, characterized during its review:

1. **Codex flat-rollout bypass**: a Codex child's rollout first-sights as a PARENTLESS SessionStart (`detect_parent_id` is CC-layout-specific), so a tombstoned Codex child can still register as an orphan through the JSONL leg — and a late hook Start can then enrich the phantom with a parent link. Bounded by the 5-min `STALE_SHORT_IDLE_TIMEOUT`.
2. **Known-id post-GC window**: the tombstone mints only for an UNKNOWN-id hook SessionEnd. Start→Stop(known slot)→GC(4.5s)→late duplicate discovery (e.g. notify outage defers the transcript first-sight to the 60s poll) re-registers a phantom with no tombstone to consult. Pre-existing, strictly narrowed by #242.
3. **Post-TTL window**: any child Start >5s after a tombstoned end registers (deliberate — the TTL bounds the guard; pinned by `child_session_start_past_tombstone_ttl_registers`).

All three ride the documented sweep backstops; none regressed by #242/#243.

## Why deferred

Closing 1 properly needs the tombstone itself to carry the discriminator instead of relying on the event's parent_id: the cleanest seam (per the design review) is the decoder stamping the SubagentStop's SessionEnd as **ended-as-child**, so the gate can key on "this id ended as a child" rather than "this registration claims a parent" — which also closes window 2 if the tombstone mints for known-id child ends too. That touches the AgentEvent shape (or a parallel reducer map) and the Reasonix-exemption analysis — a design change, not a patch. Window 3 is the deliberate TTL bound.

## Fix sketch

Mint a child-end tombstone for EVERY SubagentStop-decoded SessionEnd (known or unknown id), keyed in a reducer-private map, and gate child-shaped registrations (CC: `agent-*` stems; Codex: ids seen in the SubagentStart/Stop correlation) on it regardless of the registering event's parent_id. Reasonix stays exempt (its ends never decode from SubagentStop). Extend the #242 test suite with the Codex-flat-rollout reorder scenario and the known-id post-GC scenario.

Surfaced by #242's two-lens review (PR pending).

IvanWng97:
## What (Closes #244; advances #246 — its remaining half stays open)

The #242 reorder gate left three residual windows (#244), and a re-registered child lost its parent link (#246's fixable half). One unified design:

- **`AgentEvent::SessionEnd` gains `as_child`** — stamped true ONLY by the two SubagentStop decode arms (CC + Codex); all six production constructors audited and pinned (shared hook arm, reasonix, negative-vouch/instant-exit, buried terminator → false).
- **Reducer-private child ledger** (`child id → {applied parent, ended_at}`): parent recorded on APPLIED links only; `ended_at` from as_child ends (slot or no slot) AND from sweep removal; pruned 90s after end (`CHILD_END_LEDGER_TTL` — sized past the 60s poll-discovery latency, same calculus as the drained-Task tombstone).
- **Gates** (compose with the #242 5s tombstone, judged on the WIRE parent before adoption): a parented Start of a freshly-ended child → skip (**closes #244-w2**, the known-id post-GC window); a parentless Start of a known child → **adopt the ledger parent** — then the #240 cycle filter runs on the adopted link — so revivals re-link and a dead child's late flat first-sight registers parent-LINKED instead of orphaned (**improves #244-w1**; rides the parent cascade / Codex 5-min reap).
- **Honest scoping (upstream-verified)**: an IN-FLIGHT multi-turn Codex child has NO SessionStart carrier at turn N+1 on either transport (codex-rs hook_runtime.rs: UserPromptSubmit fires only for direct user input; child-context events carry ROOT session_id; SubagentStart only at thread startup; the rollout stays seen-claimed). The re-link therefore engages for re-registrations that occur (flat first-sights, post-un-claim revivals); #246 stays open for the hook-End→seen-un-claim design that creates the missing carrier. Documented at the seam, in the sharp edges, and on #246.
- Reasonix's parentless same-id resurrect: exempt by construction, pinned twice.

## Tests

3 red-first behavior pins (+ red-proofs), enrichment-path adoption pin, independent #242-gate pin (proven by neutering the gate), white-box ledger bounding pin (cascade→sweep stamp + 90s prune), both-transport revival pin, full constructor-stamp pins. 626/626 core, 1418/1418 workspace, dual-target clippy, full preflight green.

## Review

2 independent agents, both **APPROVE-WITH-NITS**; all nits applied incl. the upstream-carrier honesty round (one finding triggered a live codex-rs source verification that CORRECTED this PR's original multi-turn claim before merge).


IvanWng97:
Both bot findings addressed: (1) the MEDIUM — agreed with the reviewer's own option-2 lean and went further on the why: a misbehaving `Source` already controls the entire event stream and can forge `SessionStart` parent links DIRECTLY (strictly stronger than ledger poisoning), so a structural guard would add friction to the documented extension seam without adding a trust boundary; the contract is now spelled out on the field doc. (2) the fallback-path `as_child` assertion — applied as suggested.

Issue #281, opened 2026-06-13
v2-daemon prereq: liveness probe layer is process-local


IvanWng97:
From the 2026-06-13 source-layer scalability review (verified real-liability). **Inherent, not a defect; only matters for a multi-host daemon — filed to track.**

**Problem.** The whole liveness ladder is same-host by construction: `kill(pid, 0)`, the proc-table / open-FD enumeration (`fd_probe`, `cc_probe`), and the `exit_watch` kqueue/pidfd all observe LOCAL processes. A daemon observing agent CLIs on remote machines could not vouch/negative-vouch/instant-exit for them.

**Why deferred.** The visualizer runs on the same host as the agent CLIs it watches; remote observation is not a current use case.

**Fix sketch (only if remote ever matters).** Define a transport for liveness facts from a remote probe agent (the probe `Option<ProbeSnapshot>` seam is already the right abstraction — it would just be fed over the wire), or accept that remote sessions fall back to the mtime/negative-vouch-by-silence path. Keep the local probe as the default.

IvanWng97:
Closing as **out of scope**.

The app runs on the **same host** as the agent CLIs it watches (single-machine, local), so the same-host liveness ladder — `kill(pid,0)`, fd/proc enumeration, kqueue/pidfd — is correct and complete. A `RemoteProbe` only has value once a multi-host v2 daemon **and** a per-host `ProbeAgent` exist, neither of which is planned, and standing it up now means a guessed wire format + an HTTP-client dep single-host users would carry, with no payoff.

The abstraction is already right: `LivenessProbe = Arc<dyn Fn() -> Option<ProbeSnapshot>>` is a clean drop-in seam and the negative-vouch / instant-exit / first-sight machinery is transport-agnostic — so this stays cheap to revisit. Reopen as part of a daemon epic if remote observation ever becomes a use case.

Issue #280, opened 2026-06-13
v2-daemon prereq: scene channel is single-client-shaped (one watch::Receiver)


IvanWng97:
From the 2026-06-13 source-layer scalability review. **Not actionable until a v2 daemon is built — YAGNI; filed to track the prereq.**

**Problem.** The reducer publishes scene state on a single `watch::channel` consumed by the one TUI renderer. A daemon serving many concurrent clients (multiple TUIs / a web view) needs a fan-out (broadcast, per-client cursors, or a pub/sub) — a `watch::Receiver` only gives every reader the latest value with no per-client backpressure/history.

**Why deferred.** Today there is exactly one consumer; `watch` is the correct, minimal choice for it.

**Fix sketch.** Introduce a fan-out layer between the reducer and clients (e.g. `tokio::sync::broadcast` or a subscription registry) only when a second consumer is real; keep `watch` for the single-client path until then.

IvanWng97:
Closing as **out of scope**.

The planned direction is a single-window, local app (a Swift frontend embedding `pixtuoid-core` via FFI, or at most one socket client) — there is **no multi-client / multi-view requirement**, so the `watch`→`broadcast` fan-out has no consumer. Today the scene `watch::Receiver` has exactly one reader (TUI *or* headless, never both), and `watch` is the correct minimal choice for it; adding a broadcast layer now would be a no-op refactor that trades latency + `watch`'s no-lag-drop reliability for a multiplex nothing uses.

If a true multi-client v2 daemon is ever pursued, reopen as part of that epic — the fan-out approach is recorded in the issue body above.

PR #287, opened 2026-06-14
feat(core): serializable SceneState + full-scene snapshot regression net (#279)

### LINKED ISSUE #279: v2-daemon prereq: SceneState tree is not serializable
From the 2026-06-13 source-layer scalability review (adversarially verified). **Not actionable until a v2 daemon is actually built — YAGNI; filed so the prereq isn't lost.**

**Problem.** The "v2-daemon-ready" framing (e.g. the `AgentSlot.state_started_at` sharp edge) is overstated: only the leaf `SystemTime` field is `Serialize`-able, not the `SceneState` tree as a whole. A persistent daemon that survives client reconnects / restarts needs to snapshot+restore state.

**Why deferred.** No daemon exists; the single-process visualizer rebuilds state from the live event stream on every launch, so persistence buys nothing today.

**Fix sketch (when a daemon lands).** Derive `Serialize`/`Deserialize` across `SceneState` + `AgentSlot` + the reducer's correlation maps (or define an explicit snapshot DTO), and add a round-trip test. Audit the reducer-private maps (`state/correlation.rs`) for non-serializable members first.

IvanWng97:
## What

Derive `Serialize`/`Deserialize` across the `SceneState` tree (`SceneState`, `AgentSlot`, `ActivityState`, `GlobalDeskIndex`) and add a **full-scene snapshot regression net**.

`Arc<str>`/`Arc<Path>` slot fields cross through owned `String`/`PathBuf` via three local `*_serde` adapters in `state/mod.rs` — serde has no blanket `Arc<T>` impl, and its `rc` feature wouldn't cover `Arc<Path>` (no `Box<Path>: Deserialize`). **No new dependency** (serde + serde_json already core deps; `AgentId` already derived serde).

## Why (the standalone, daemon-independent payoff)

`tests/reducer/snapshot.rs` drives a deterministic, fixed-timestamp script through the reducer and snapshots the **whole serialized scene** (insta YAML). Unlike the field-level reducer tests, this locks the tree's shape *and* the reducer's end-to-end output in one reviewable golden — the regression net for the recent core-layer (Fable-5) refactors. The golden already captures `cc·`/`cx·` labels, Delegating/Generic/Waiting states, parent linkage, desk allocation, and liveness-up-propagation (parent `last_event_at` refreshed by the child). Plus a JSON round-trip unit test for the serde impl itself.

The serialized shape is **debug/snapshot-oriented, not a stable wire contract** — adding a slot field is non-breaking (the golden just surfaces it for review).

## Scope

This is #279's now-actionable part. The v2-daemon snapshot/restore *consumer* — and the #280 multi-client / #281 remote-probe prereqs — stays deferred (those issues closed as out of scope for the single-window, embedded-frontend direction).

## Test plan

- `cargo test -p pixtuoid-core` — round-trip + full-scene golden green
- `just preflight` — machete / clippy `-D warnings` / cargo-hack / 1540 tests, all green

Closes #279


IvanWng97:
Addressed in c2d48fa (review round). The MEDIUM (Idle never serialized) is fixed: an `Idle` slot now appears in **both** the round-trip unit test and the insta golden, plus a winding (Active→Idle debounce) slot so the golden also pins a populated `Option<SystemTime>` (`pending_idle_at`). The golden now covers all four `ActivityState` shapes.

Also folded in a local three-lens review: reconciled a doc-currency contradiction the PR opened (workspace `CLAUDE.md` still said "v2-daemon-ready" after #279/#280/#281 were closed), softened the four surviving `v2 daemon` code doc-comments, and documented why `FloorLocalDeskIndex` deliberately lacks serde. Adjudication recorded in `docs/REVIEW-LEDGER.md` (R0614-11..15).

PR #288, opened 2026-06-14
fix(media): render real glyphs in demo snapshots (legible popup text)

IvanWng97:
## Problem

The website/README demos show all UI text as **colored blocks** (see the CH 04 DASHBOARD station). Root cause: the snapshot rasterizer in `examples/snapshot.rs` (`save_backend_as_png` + the GIF twin `cells_to_rgba`) only handled the `▀` half-block (office sprites); **every other cell was drawn as a small fg square** — "a tiny glyph box to represent where text lives." No actual letters. The **live terminal was always fine** (real font); only the generated media was unreadable.

## Fix

Render real glyphs. An 8×8 bitmap font (`font8x8`, **dev-dependency** — example/gen-media only, never in the shipped binary) blitted 1×2 into each 8×16 cell, via one shared `glyph8x8` + `blit_glyph_cell` used by **both** rasterizers (no drift). A glyph no font set covers (decorative `●○◐` status dots, arrow keys) falls back to the old centered block. Bit order verified non-mirrored against the regenerated demos.

**No TUI / product change** — the dashboard/connection panels keep their exact layout (`[cc] cc·proj …`); this is purely the demo-media renderer.

Before → after (dashboard popup): rows of colored blocks → `[cc] cc·pixtuoid  F1  ● Edit: reducer.rs` etc. Connection popup now reads `[cc] Claude Code · connected · 2 agents · 3s ago`.

## Scope

Fixes asks ① (connection demo) and ③ (dashboard demo) from the feature batch — both via this one media fix. Website reorg (②) + pets/meeting close-up crops (④) follow in a separate PR.

## Test plan

- `just preflight` — fmt / machete / **cargo-deny (font8x8 = MIT, allowed)** / clippy `-D warnings` / 1540 tests — all green
- `just gen` — regenerated 44 media files (docs/images + site/public/demos)
- `just gen-check` — green (every committed artifact in sync)


IvanWng97:
Addressed in 549567e.

**HIGH (no tests):** added two unit tests to the example's `mod tests` (which CI runs — the `[[example]]` has `test = true`, verified via nextest, now 1542 total):
- `blit_glyph_cell_lsb_is_leftmost_and_doubles_vertically` — pins LSB=leftmost (a font8x8 bit-order flip → fails here instead of silently mirroring 40+ media files), plus the 1×2 vertical double and the x0/y0 origin offset.
- `glyph8x8_covers_ascii_and_custom_dots_but_not_arbitrary_symbols` — ASCII + `·` resolve, the 3 status dots resolve (and are distinct bitmaps), an uncovered glyph (`✓`) and the empty string return None.

**LOW (state dots):** added `custom_glyph` with 3 hand-rolled 8×8 bitmaps (`●` filled disc / `○` ring / `◐` left-half), consulted before the font8x8 chain — the dashboard dots are now shape-distinct (green disc / gray ring / yellow half), not just color-distinct.

Ledger: R0614-16..18.

PR #316, opened 2026-06-15
feat: OpenClaw source (ok·) — wandering gateway mascot "Molty"

IvanWng97:
## OpenClaw — the 9th source (`ok·`), a gateway mascot

Adds **OpenClaw** as a supported source. OpenClaw is a local-first **daemon gateway** (`ws://127.0.0.1:18789`) that multiplexes many chat sessions to in-process agents. Its backend `claude-cli` coding sessions already render via `cc·`, so OpenClaw is shown as **one presence-gated wandering mascot — the lobster "Molty"** — *not* a desk agent. He appears only when a gateway is live, so the ~99% who don't run one see a normal office.

### Why a mascot, not a desk agent
OpenClaw has no per-session transcript and produces **zero `AgentEvent`s**; presence rides a sibling channel into `SceneState.source_presence` (the documented `presence_only` source class). Rendering it as a desk agent would be wrong (it's a daemon, not a coding session) and would double-count the backend `cc·` sessions.

### Motion encodes gateway state
Reuses the office-pet's stateless, no-flash A\*-on-static-mask wander (Molty lives in `source_presence`, not `scene.agents`):

| state | behaviour |
|---|---|
| gateway up | scuttles in from the elevator (the `entered_at` anchor) |
| idle | ambles the social spots & rests |
| busy | shuttles toward the backend desks + activity bubbles |
| down | walks out to the elevator and vanishes |

**Busy intensity keys on in-flight runs, not session count** — per upstream docs + a research sweep, the single-user default collapses all DMs into one persistent `main` session (so `active_sessions` sticks at 1 even at rest; a session-scaled bubble would linger while idle).

### Commits
1. `feat(core)` — presence-only source: `decode_openclaw_hook_payload` → `DaemonPresenceUpdate` on a sibling channel; `DaemonPresence` on `SceneState`; `apply_presence` + TTL sweep + pid `ExitWatch` (abrupt-down); runtime wiring.
2. `feat(tui)` — marigold `ok` badge across 6 themes (clears the 60-Manhattan hue floor) + `sources.json` manifest.
3. `feat(install)` — two-ownership target (the FIRST): a TS plugin **dir** (`extra_artifacts`) + a `ConfigLock` merge into `openclaw.json` (`load.paths` + `entries.pixtuoid.enabled` + the `allowConversationAccess` grant). Strict allowlist (timing/ids only, never message content); detached non-awaited shim spawn (never blocks the gateway, invariant #5).
4. `feat(tui)` — the render layer: the wandering Molty sprite + `entered_at` enter/leave anchor + hover tooltip; the run-intensity fix; the drift watch.

### Drift defense
`OPENCLAW_EVENTS` is the single source of truth, pinned plugin `HOOKS` ⇔ decoder arms ⇔ const by a consistency test; a decode breadcrumb fires on an unknown event; `check_upstream_drift.py` watches the 6 depended events in upstream `src/plugins/hook-types.ts` (one-directional, like opencode/cursor).

### Verification
- `just preflight` green (1647 tests), `just gen-check` green (Molty is presence-gated, so default baselines are unchanged).
- **Live-verified e2e** against a real `~/.openclaw` gateway: `gateway_start` → Molty idle; `kill <gateway-pid>` → walk-out (the pid `ExitWatch` / `PidExited` rung).
- Busy is the same proven path (test-covered); not yet eyeballed live.

### Resolved in this PR (no longer deferred)
- **#317 — the "degraded" 4th state** (gateway up but its model backend failing every run — the Apr-2026 Anthropic-ban failure mode): the signal already existed as `agent_end.success: boolean`, so `success:false` → `DaemonState::Degraded` (sickly-red, sluggishly-wandering Molty + tooltip "model error"); heals on a clean run / new run / restart. The plugin forwards only the `success` bool, never the error string.
- **#318 — mid-attach pid binding**: the plugin stamps `_pid` on every event; `PidSeen` adoption arms the instant abrupt-down rung for a daemon attached to after `gateway_start`.

Both closed by the follow-on commit; verified live end-to-end (real gateway + a claude-cli backend turn showed Molty + a `cc·` backend sprite in one frame).


IvanWng97:
## Thorough review complete + addressed

Ran a comprehensive multi-agent review (8 dimensions × 3 adversarial lenses — correctness-repro / design-intent-vs-CLAUDE.md / blast-radius): **17 confirmed, 6 refuted, 0 critical / 0 high**. All confirmed findings are now fixed in `63355f6`; adjudications logged in `docs/REVIEW-LEDGER.md` (§ 2026-06-15 OpenClaw).

**Merge-blocker (L5):** the branch predated #314's required `Target.verify_schema`. Merged `main`, authored `openclaw::verify_schema` (config-level enabled-entry + `load.paths` checks; `ShimRef::Unknown` since the shim lives in the plugin file), wired the field. Compiles + `verify_target_is_sound_after_a_real_install_for_every_target` covers it.

**Correctness (M1, the one real gap):** the presence side-channel **bypassed the connection gate** — a panel-disconnected gateway kept rendering Molty. Now: the presence arm is gated on `is_connected(openclaw)`, the sweep tick reconciles via `mark_presence_down` (walks him out), and `sweep_presence_ttl` GCs `Down` entries after the walk-out window (which also fixed **L1** — TTL→Down now re-anchors `last_seen` so the SIGTERM path animates the leave instead of vanishing).

**Docs/nits:** L2 (`extra_artifacts` "idempotent" doc → "rewritten verbatim"), N1 (documented the stateless Idle-spot walk-out reconstruction, citing the stateless-mascot invariant).

**Test hardening (M2/M3/M4/L4/N2/N3/TA-7):** busy-bubble harness test (runs ≠ sessions, idle-never-bubbles), `entered_at` resurrection re-anchor + `run_key` fallbacks, floor-0 projection, tooltip-text unit test, `handle_conn` presence-duplex routing test, and a byte-real conformance pin for the presence decoder (a field rename now fails loudly).

**Refuted (deliberate-design, all cite a sharp edge):** JS allowlist type-narrowing (Rust decoder is the sanitizer), drift deprecation signal (upstream is string literals), hardcoded presence decode (single-daemon YAGNI), in-process gateway pid (capture-verified, live kill→walk-out), CC owning `presence_tx` (the one shared socket, invariant #2).

**Deferred:** a 4th "degraded" state (gateway up but auth-broken — the Apr-2026 Anthropic-ban mode) → tracked as #317 (net-new scope; needs a run-failure signal).

`just preflight` green (1683 tests), `gen-check` green. **Live-verified e2e** against a real `~/.openclaw` gateway: idle ✓, abrupt-kill walk-out ✓.

PR #338, opened 2026-06-16
fix(install): resolve CodeWhale + OpenClaw config paths HOME-first on Windows

IvanWng97:
## Problem

A user reported CodeWhale "installed but no sprite appeared" on Windows. Root cause is a cross-binary path divergence: pixtuoid resolved `~` **USERPROFILE-first** (its generic `platform::user_home`), but **CodeWhale's own `effective_home_dir` is `$HOME`-first** (`$HOME ?? dirs::home_dir()`). On a `HOME`-exporting Windows shell (Git Bash / MSYS2 / Cygwin), pixtuoid wrote hooks to `%USERPROFILE%\.codewhale\config.toml` while CodeWhale reads `%HOME%\.codewhale\` → hooks installed, but the CLI never loads them.

Auditing **every** source against its CLI's actual Windows resolution surfaced a **second** instance: **OpenClaw** (`infra/home-dir.ts::resolveRawOsHomeDir` = `$HOME ?? $USERPROFILE ?? os.homedir()`, + honors `OPENCLAW_HOME`) had the same bug for its gateway plugin.

**The other six CLIs are safe** — they defer to a USERPROFILE-first stdlib home (Rust `dirs::home_dir`, Go `os.UserHomeDir`, Node `os.homedir`, all ignoring `$HOME` on Windows), which pixtuoid's generic resolver already matches. CodeWhale + OpenClaw are the only two that hand-roll `$HOME`-first.

## Fix

- **`platform::home_first_dir`** (generalized from `codewhale_home`) — the shared HOME-first→USERPROFILE-on-Windows resolver for the two hand-rolled CLIs.
- **`install/codewhale.rs`** — `default_config_path` routes through `home_first_dir` and honors `CODEWHALE_CONFIG_PATH` / `DEEPSEEK_CONFIG_PATH` first (CodeWhale's `resolve_config_path` order); `detect_installed` matches.
- **`install/openclaw.rs`** — `openclaw_home` mirrors OpenClaw's `resolveStateDir`: `OPENCLAW_STATE_DIR` → `OPENCLAW_HOME` → `home_first_dir()/.openclaw` (pure injectable core).
- **`doctor::home_split_advisory`** — a Windows-only safety net in `pixtuoid doctor` that fires when `HOME` ≠ `USERPROFILE` (the host condition for this whole bug class); cosmetic slash/case diffs don't trip it, a POSIX-form `HOME` vs native `USERPROFILE` does.
- **Regression guards** — `platform.rs` codifies the per-source home-basis contract (the audit) as an executable comment + a divergence test, so a future "simplify back to the generic home" re-break fails CI; pure-core tests for `resolve_openclaw_home` / CodeWhale `resolve_config_path`.

## Why this escaped local debug + CI
USERPROFILE-vs-HOME only diverges on **Windows with `HOME` set**; on Unix both sides use `HOME` (agree by construction), and CI's `check-windows` is compile-only (never runs the binary). It's a cross-binary contract mismatch no single unit suite spans — hence the new contract test.

## Verification
`just preflight` green — **1718 tests passed, exit 0** (lint → clippy → hack → test). All new resolvers are pure/injectable and unit-tested.

> Note: still needs the repo's **two-lens review** before merge (correctness/grounding + design/blast-radius).


IvanWng97:
Superseded by #343 — rebased cleanly onto current main (0.10.0) and recreated on a fresh branch to avoid a force-push. The online review here was carried over: the 1 MEDIUM (stale codewhale module-header doc) is fixed in #343; security review was clean. Same commits + a Reasonix %APPDATA% fix, a Cursor CURSOR_CONFIG_DIR/XDG fix, and the full per-axis re-verification.

PR #356, opened 2026-06-18
chore(lint): enforce pub(crate) for layer-internal items workspace-wide

IvanWng97:
## What

Mechanically enforce the encapsulation rule "inter-layer items don't leak as public API" workspace-wide.

- Add `unreachable_pub = "warn"` to `[workspace.lints.rust]` (root `Cargo.toml`) and opt every crate in via `[lints] workspace = true`.
- CI's `just clippy` (`cargo clippy --workspace --all-targets -- -D warnings`) turns the warn into a **hard gate** — a `pub` item in a private module tree (one unreachable as crate API) now fails the build and must be `pub(crate)`.
- Fix the 16 pre-existing offenders the lint surfaces.

## Why

This is the backstop for the encapsulation work: `install_target`/`uninstall_target` are already `pub(crate)` (`crate::sources` is the sole caller), but nothing *stopped* a future inter-layer seam from silently shipping as `pub`. `unreachable_pub` is that stop.

## The 16 fixes (all crate-internal → semver-neutral)

`cargo-semver-checks` never sees an unreachable item, so none of these touch the public API / semver surface:

| crate | items |
|---|---|
| `pixtuoid-core` | `decor::DESK_WALK_{X,Y}_OFF`, `hook::router::SOURCE_NAME`, the 6 `state/*_serde` adapter fns |
| `pixtuoid-hook` | `paths::default_socket_path`, `transport::WRITE_TIMEOUT` + both `send_line` arms |
| `pixtuoid` | `floating::window::{FloatingEvent, FloatingApp, FloatingApp::new}`, `tui::widgets::tooltip::paint_mascot_tooltip` |

## Verification

- `RUSTFLAGS="-W unreachable_pub" cargo build` → exactly these 16 before, **0 after**.
- Confirmed the gate fires *via the workspace config* (no RUSTFLAGS): reverting one item to `pub` reproduces the warning, restoring clears it.
- `just preflight` green (lint → clippy → hack → **1865 tests pass**).

## Scope

`unreachable_pub` only catches `pub`-in-private-module leaks. A deeper `cargo-public-api` audit (items that are reachable-`pub` but used by no other crate) is a separate, optional follow-up — not in this PR.

### DIFF
--- a/.github/prompts/impl-plan.prompt.md
+++ b/.github/prompts/impl-plan.prompt.md
@@ -51,6 +51,18 @@ Every section gets an answer; "n/a" counts only with a reason.
    the `beautify-decoration` loop. Verification steps are blocking plan
    items, not checkboxes — PR #61 shipped five walk regressions behind an
    unchecked "live run" checkbox.
+8. **Layering / orchestration boundary** — if the change adds or crosses a
+   layer seam (a mechanism/foundation layer with an orchestrator over it,
+   like `install` ← `sources`), name the ONE orchestration entry point and
+   design the lower layer's mechanism API as `pub(crate)`, never `pub`:
+   callers reach it ONLY through the orchestrator, never directly. Do NOT
+   expose an underlayer/foundation API as crate-public so something can call
+   it around the facade — that is a design leak `unreachable_pub` can't catch
+   once the module itself is reachable, so the plan is where the
+   single-gateway shape is decided (install/uninstall route SOLELY through
+   `crate::sources`; the plan states "no second caller" and which existing
+   direct calls collapse into the orchestrator). A new public seam on a lower
+   layer needs a one-line justification of why the orchestrator can't own it.
 
 ## The contract with review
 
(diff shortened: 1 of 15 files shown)


IvanWng97:
Added a second commit (`59e2bcb4`) folding the **design half** of this principle into the review + plan briefs (per maintainer note: "don't expose underlayer/foundation-layer API, force follow the orchestration design"):

- **`impl-plan.prompt.md`** — new item 8 "Layering / orchestration boundary": name the one orchestration entry point, keep the mechanism API `pub(crate)`, justify any new public seam on a lower layer.
- **`pr-review.prompt.md`** — new lens-2 item 7: flag a new `pub` exposing an underlayer seam, or a new call site reaching the mechanism around the facade.

Rationale: `unreachable_pub` (this PR's lint) is the *mechanical* half — it catches a `pub` in a **private** module tree. It can't catch a *reachable*-but-should-be-funnelled API; that's the design check the briefs now own. Single-gateway exemplar in both: install/uninstall are `pub(crate)`, `crate::sources` is the sole caller.

Prose-only change to the brief docs — no code/build impact, so the two passing lens reviews on the lint change still hold.

PR #359, opened 2026-06-18
feat(onboarding): cinematic first-run "move-in" overlay + headless `setup` (PR2 of 3)

IvanWng97:
PR 2 of the sources-core arc (after #355). A first-run onboarding "move-in" overlay over the `crate::sources` core, plus its headless twin.

## What the user sees
On the **first ever launch** (`setup::is_first_run` = config absent OR `[sources]` empty), the TUI plays a one-time overlay: a borderless card with a **typewriter** subtitle and a **staged roster reveal** — each detected agent CLI fades in with its colored `[xx]` badge + a `[x]`/`[ ]` checkbox. `↑↓/jk` move, `space` toggles, `Enter` connects the checked, `Esc` skips. It's modal and the **top of the precedence chain**.

## Architecture (over the same core, no new IPC)
- **Pure model** `tui::welcome::{WelcomeUi, WelcomeRow}` (roster checklist + nav) + `OnboardingFrame` (the bundled render snapshot) — mirrors `tui::dashboard`/`tui::connection`.
- **Painter** `widgets::welcome::paint_welcome` — typewriter + staged reveal driven by `OnboardingFrame.elapsed_ms` (the model holds no clock); painted topmost via the shared `paint_overlays` (both draw paths).
- **Apply** `sources::apply_choices(cfg, &[(id,bool)])` — connect checked / disconnect unchecked, **scoped to the detected roster**. Deliberately NOT the plan's declarative `reconcile_to`: that's over *every* registered source and would disconnect the complement, incl. `antigravity` (connected-by-default, never in `detect()`). `Esc`/skip persists a *freeze* of current state so the gate closes without touching hooks.
- **Headless twin** `pixtuoid setup [--yes]` (Raycast/CI) — dry-run preview by default, `--yes` connects the detected set via the same apply path.
- **Threading** `main::is_first_run → RunConfig.first_run → TuiSession → run_tui`.

## Two refactors folded in (maintainer-approved)
- `refactor(tui)`: bundle `run_tui`'s 11 positional args into a `TuiSession` struct (it was about to become 12 — transposition hazard).
- The onboarding render-frame's 4 fields bundled into one `OnboardingFrame` rather than 4 parallel `onboarding_*` fields (the better shape; the older `connection_*`/`dashboard_*` mirrors stay flat — separate cleanup).

## Edge handled (caught in review)
[...]

PR #360, opened 2026-06-18
ci(smoke): make ffmpeg install resilient to apt-mirror stalls

IvanWng97:
## Problem
The smoke job's `Install ffmpeg` step:
```yaml
run: sudo apt-get update && sudo apt-get install -y ffmpeg
```
Ubuntu's apt mirrors on GitHub-hosted runners intermittently **stall**. With no per-fetch timeout, the step **hangs until the job timeout fires** → GitHub **cancels** the whole smoke job (a red `smoke` check that never even reached `gen-check`). It hit **~3× in one session** (#356, #359, the #359 post-merge `main` run) — each "fixed" only by a manual re-run.

## Fix
Bound + retry at the **apt level** (where the stall is), no new dependency:
- `Acquire::http::Timeout=30` / `Acquire::https::Timeout=30` — a dead mirror fails **fast** (≤30s/fetch) instead of hanging.
- `Acquire::Retries=3` — apt retries transient fetch errors itself (often a healthier mirror).
- A **3× outer retry** with backoff. Safe because apt fails on its *own* timeout (releasing the dpkg lock) — no `SIGKILL`/lock-leak.
- `--no-install-recommends` trims the install.

## Why not a `setup-ffmpeg` action
Keeps it self-contained (no third-party action / supply-chain pin), and targets the actual failure mode (apt fetch stalls) rather than swapping the dependency for another downloadable one.

## Verification
- `shellcheck` clean on the run-block; YAML validated.
- The real test is **this PR's own smoke run** — it executes the new step (GitHub uses the PR branch's workflow for `pull_request`).


IvanWng97:
**Two-lens review done.**
- **Correctness lens: SOUND** — verified the bash under GHA's `set -e -o pipefail` (the `if … && …; then` correctly exempts a failed attempt from `set -e`), the `apt_opts` array expansion (shellcheck-clean), the arithmetic backoff, the apt config keys, and the retry-after-own-timeout lock safety (no SIGKILL). The dropped `| head` SIGPIPE hazard is gone.
- **Blast-radius lens** flagged `--no-install-recommends` (VP9/libvpx for clips). **Premise verified & refuted for smoke**: `just gen-check` runs `gen-media.py --check`, which **skips clip/VP9 regen entirely** ("vp9 encoding blew the CI timeout"); its only ffmpeg op is a PNG→PNG `crop` (core codecs). So the flag was safe — but I **dropped it anyway** (it isn't part of the fix, and keeping it would make a future gen-check op silently depend on the recommends set). Ledger R0620-01.

The fix is **self-validating**: this PR's own smoke run executes the new step (GitHub uses the branch's workflow for `pull_request`). If the resilient install works here, it works on `main`.

PR #367, opened 2026-06-19
feat(contract): generate the Raycast --json type from the Rust serde source

IvanWng97:
Closes the one honest gap `docs/PARALLEL-DELIVERY.md` named: the `pixtuoid sources --json` `SourceStatus` shape was typed by hand in **both** Rust (`sources.rs`) and the Raycast TS (`pixtuoid.ts`) — a hand-mirror that can silently drift. Now it's **codegen-from-one-source**: the Rust serde type is the single source of truth, the TS type is generated, and a producer shape change is a **consumer compile error**. pixtuoid now practices what the doc preaches.

## The chain
```
Rust SourceStatus ──schemars──▶ committed JSON Schema ──json2ts──▶ generated src/contract.ts
   (source of truth)        (contract/source-status.schema.json)   (pixtuoid.ts re-exports it)
```

## Three gates lock drift out (all mutation-proven to fire)
1. **struct ↔ schema** — a `schemars` golden test (`source_status_schema_matches_the_committed_contract`) fails in `just test` if the struct drifts from the committed schema. *(teeth-test: added a field → EXIT 101.)*
2. **schema ↔ TS type** — the raycast CI regenerates `contract.ts` and `git diff --exit-code`s it; a schema change not propagated to the type fails. *(teeth-test: edited the schema → diff EXIT 1.)* Airtight by path-filter: a `sources.rs` change forces a schema regen (else gate 1 reds on every PR in `ci.yml`, which has no path filter), and the schema lives under `integrations/raycast/**` so it triggers the raycast workflow.
3. **TS type ↔ usage** — `tsc --noEmit`; a renamed/removed field the extension uses is a compile error. `deny_unknown_fields` keeps `additionalProperties: false` in the schema so the generated type has no `[k: string]: unknown` escape hatch (a typo is caught, not silently `unknown`).

## Details
- `schemars` is a **dev-dependency** with `#[cfg_attr(test, derive(JsonSchema))]` — zero production-binary impact (the only reference is `schema_for!` inside `#[cfg(test)]`).
- `just gen-contract` regenerates the whole chain after a `SourceStatus` change; `src/contract.ts` is eslint/prettier-ignored (generated, never hand-edit).
- `OutcomeRow` stays hand-typed — a `{id, outcome: string}` token where a schema adds no drift protection. The byte-shape test (`source_status_json_shape`) stays.
- Docs flipped: `PARALLEL-DELIVERY.md` gap → applied/dogfooded; `integrations/raycast/CLAUDE.md` hand-mirror → generated.

## Verification
- `just preflight` **1901 green** (machete + deny pass the new dev-dep; `cargo hack --no-dev-deps` confirms `cfg_attr(test)` gating).
- raycast `tsc --noEmit` 0, `eslint .` 0; schema/type regen byte-deterministic.
- Local 3-lens review (correctness + design + contract-integrity/teeth) → 0 blocking; both NITs fixed in the review-round commit.


IvanWng97:
Thanks — re the MEDIUM on `health` being `health?: string | null`: I verified both the premise and the suggested fix, and the fix is **refuted** (it makes things worse), so `health` stays optional with a documented WHY (1d7c1686).

`#[schemars(required)]` on an `Option<String>` doesn't just add the field to `required` — it **strips the `| null`** (the schema `type` becomes `"string"`, not `["string","null"]`), so the generated type is `health: string`. That asserts non-null, but the wire **can** be null (pinned by `source_status_json_shape`: `…,"health":null`). That's the exact "mis-specified nullability / garbage-in-garbage-out" pitfall this PR's own `PARALLEL-DELIVERY.md` names — strictly worse than the status quo.

`health?: string | null` is a safe **superset** of the wire: the wire always emits `health` (so the `?` is harmless, never wrong), and the sole consumer access is `if (s.health)` — identical for optional vs required. **Nullable** is the part that matters for safety, and it's preserved. So the precision gap is real but cosmetic (optional-vs-required is a no-op here), while the proposed fix introduces a real one (a non-null assertion the wire violates).

Net: kept optional, added the trace at the field. The separate index-signature MEDIUM from the local review (a typo'd field was `unknown`, not an error) was the genuinely-impactful one and is fixed via `deny_unknown_fields` (195433f7) — that one closes the guarantee.

IvanWng97:
Round 2 — agreed, that's a real gap. Added the ledger trace in `188a03f3` (`docs/REVIEW-LEDGER.md`, R0619-06..09): the index-signature MED (CONFIRMED→FIXED via `deny_unknown_fields`), the health-required MED (REFUTED-wrong-fix, with your round-2 confirmation noted), the 3-gate chain (CONFIRMED-with-teeth), and this ledger gap itself. Every adjudication now has a cross-PR audit row.

PR #369, opened 2026-06-19
chore(lint): gate shell formatting with shfmt

IvanWng97:
## What

Adds **shfmt** as a gated shell formatter, pairing a formatter with the existing shellcheck house rule (shellcheck lints `.sh`, shfmt formats).

## Design — fold in, don't add entry points

Per the "minimize test/lint entry points so an agent can't forget" goal:

- `just shfmt-check` joins the **parallel `just lint` block** → `just preflight` (and the pre-push hook) cover it for free. No new top-level command to remember.
- A **single lightweight CI `hygiene` job** runs it — and is the designated home for the other non-Rust linters (actionlint, link-check) as they land, so **CI grows by one shared job, not one per tool**.

## Changes

- `just shfmt-check` / `just shfmt-fix` recipes (`-i 4`, no `-ci` to match the prevailing un-indented `case` style → minimal churn).
- One-time reformat of the two non-conformant scripts (`openclaw-cc-backend-e2e.sh`, `replay-fixture.sh`); `.githooks/*` + `openclaw-live-e2e.sh` already conformed. **shellcheck still clean** on the reformatted scripts.
- `setup-tools` installs shfmt (brew on macOS; else points at docs so `just lint` can't *silently* skip it — a skipped lint is the failure class we're avoiding).
- CI `hygiene` job installs shfmt via the runner's pinned Go (`v3.13.1`) and runs `just shfmt-check`.
- `CLAUDE.md` shell convention updated.

## Verification

- `just shfmt-check` ✓ clean · `just lint` ✓ (now 5 checks incl. shfmt) · `actionlint .github/workflows/ci.yml` ✓ valid.

First of the lint/test tooling wave (actionlint, lychee, knip, snapbox, proptest, … to follow as separate PRs).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #372, opened 2026-06-19
chore(site): find dead code/deps with knip

IvanWng97:
## What

Adds **knip** to the Astro site's gate — finds unused **files / exports / dependencies** that `tsc` + `eslint` (within-module only) can't see. The site is component- and data-file-heavy, exactly where orphans accumulate silently.

## Design — fold in, site's own gate

- `knip` joins the site's **`verify` script** (→ covered by `just site-check`) and the **`site.yml` CI steps**. No new top-level entry; this is the **Node project's** gate (not the Rust `ci.yml`).

## 3 findings, each verified before acting (didn't blind-trust knip)

1. **`WEATHERS` / `WeatherShot`** (`src/consts.ts`) — exported but used **only inside `consts.ts`** (line 103). Grep-confirmed nothing external (src / `astro.config.mjs` / `scripts/`) imports them → dropped the `export` (module-local now). Real cleanup; `astro check` still 0 errors.
2. **`mermaid` devDependency — a FALSE POSITIVE, kept + ignored.** It's a **transitive build-time dep** (`rehype-mermaid` → `mermaid-isomorphic` → `mermaid`) that renders the `/architecture` ```mermaid``` block to inline SVG at build (the reason site CI installs Chromium — see `site/CLAUDE.md`). It's declared **directly to pin the version** that produces that SVG (which feeds the visual gates), so it's never `import`ed in our source and knip can't trace it through the rehype plugin. **Removing it would break the diagram** → `knip.jsonc` `ignoreDependencies` documents exactly why.

## Verification

- `npm run knip` ✓ clean · `npm run check` ✓ 0 errors · `format:check` ✓ · `lint` ✓ · `just actionlint` ✓ (the `site.yml` edit).

PR-4 of the lint/test tooling wave (after #369 shfmt, #370 actionlint, #371 lychee). Next: snapbox → proptest → …

### DIFF
--- a/.github/workflows/site.yml
+++ b/.github/workflows/site.yml
@@ -47,6 +47,10 @@ jobs:
       - run: npm run format:check
       - run: npm run lint
       - run: npm run check
+      # knip: unused files / exports / dependencies in the Astro project (catches
+      # an orphaned component or stale dep tsc/eslint can't see). Mirrors the
+      # `verify` script; runs before build for fast feedback.
+      - run: npm run knip
       - run: npm run build
       # Lighthouse CI: gate a11y / SEO / best-practices (perf advisory) against
       # the built pages — serves dist via `astro preview` (honors the /pixtuoid
--- a/site/README.md
+++ b/site/README.md
@@ -18,11 +18,12 @@ npm run dev        # http://localhost:4321/pixtuoid/   ·  just site-dev
 ## Quality gates
 
 ```sh
-npm run verify     # format:check → lint → astro check → build  (== site CI)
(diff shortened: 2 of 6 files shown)


IvanWng97:
Bot-findings-adjudicated: finding 1 (MEDIUM, stale `site/README.md` verify description) — CONFIRMED → FIXED in 6f1a19c5 (added knip to the verify chain + the individual-commands list). Reviewer independently confirmed the WEATHERS de-export and the mermaid false-positive ignore; no other findings.

PR #400, opened 2026-06-21
fix: truecolor preflight false positives (terminfo Tc/RGB + PIXTUOID_NO_TRUECOLOR_WARN)

### LINKED ISSUE #397: Truecolor preflight: terminfo (Tc/RGB) detection + suppression hatch to cut false positives
Follow-up from the CLI-best-practices change (feat/cli-best-practices, two-lens review LOW).

The new truecolor preflight (`crates/pixtuoid/src/term.rs` + `main.rs`) warns on a non-headless `run` tty when `$COLORTERM` doesn't advertise `truecolor`/`24bit`. It is intentionally **warn-only** (never gates) and pre-altscreen, but `$COLORTERM` is widely unset by genuinely-truecolor setups (tmux that strips it, some SSH sessions, terminals that signal truecolor via terminfo `Tc`/`RGB` rather than `$COLORTERM`), so it can false-positive nag a legitimate user with no escape hatch.

**Refinements to consider:**
- Also consult terminfo truecolor capability (`Tc` / `RGB`, ncurses 6.0+) before warning — the formal fallback signal the S-Lang convention names. (This is the real fix; it's why this is deferred rather than done in-PR.)
- And/or a one-line suppression env var (e.g. `PIXTUOID_NO_TRUECOLOR_WARN`) for users who know their terminal is fine.

Non-urgent: the warning is harmless (warn-only, scrolls away under the alt-screen). Revisit if false-positive reports come in.

mvanhorn:
## Summary

The truecolor preflight nags legitimate users. It warns on a non-headless `run` tty whenever `$COLORTERM` doesn't advertise `truecolor`/`24bit`, but that variable is widely unset on genuinely-truecolor setups: tmux strips it, some SSH sessions drop it, and a class of terminals signal 24-bit color through the terminfo `Tc`/`RGB` capability instead of `$COLORTERM` at all. Those users get a warning they can't act on and can't silence, which is the false-positive class this issue describes.

This teaches the warning to consult terminfo before firing, and adds a documented escape hatch. The check now stays silent when `$COLORTERM` advertises truecolor, when terminfo exposes `Tc`/`RGB` for the current `$TERM`, or when `PIXTUOID_NO_TRUECOLOR_WARN` is set to a truthy value, and the warning copy itself now names that env var so a nagged user has a way out. The behavior that matters is preserved: it stays warn-only and never gates on Unix, prints pre-altscreen, and Windows continues to hard-gate VT separately.

The policy lives in one pure function that takes its env values and an injected terminfo lookup, so every branch is unit-testable without touching the process environment, matching the existing style in this module. The real terminfo probe shells out to `infocmp` and degrades gracefully: any failure (no binary, no entry, unreadable database) is treated as not advertised, so a host without terminfo behaves exactly as it does today.

One subtlety worth calling out for review is that the probe must pass the extended-capabilities flag, because `Tc` and `RGB` are user-defined capabilities that the default output omits, which would otherwise drop the very signal this fix relies on. The `pixtuoid doctor` terminal verdict was also updated to read the same signals, so the diagnostic and the startup warning can no longer disagree about whether a terminal is truecolor.

## Related issue

Fixes #397

## Type of change

Bug fix.

## How I tested it

New unit tests cover every scenario the issue lists: the COLORTERM happy path, the terminfo fallback that is the real fix, the suppression hatch firing for truthy values but not for empty or falsy ones, the non-truecolor case where the warning is correctly preserved, and graceful degradation when terminfo is unavailable. One test drives the real `infocmp` seam to pin the extended-capabilities requirement, skipping when the host has no entry advertising it so it never goes flaky on CI. Locally I ran format, clippy, build, and the full library test suite for the crate, and all 650 tests pass.

Not a visual change. This touches terminal-capability detection and the `doctor` diagnostic only, with no sprite or pixel-painter edit.

## Checklist

I read the root and crate `CLAUDE.md`, added tests for the new behavior (TDD-first), kept non-test code free of `unwrap()` with errors propagating via `anyhow`, added no new production-path `println!`/`eprintln!` (the existing pre-altscreen warning is preserved, not added), and updated the crate guide's term entry in the same commit. I ran the per-step equivalent of `just preflight` (format, clippy, build, and the crate test suite); a pre-existing `collapsible_match` clippy warning elsewhere in `pixtuoid-core` is unrelated to this change.

## AI assistance

This PR was written with AI-agent assistance.

### DIFF
--- a/crates/pixtuoid/CLAUDE.md
+++ b/crates/pixtuoid/CLAUDE.md
@@ -38,11 +38,16 @@ src/
 ├── term.rs             truecolor preflight (BOTH fns PURE + unit-tested — main.rs is the untestable presenter,
 │                       so the policy lives here; `doctor::run` returns its report String so it's tested too):
 │                       `colorterm_is_truecolor($COLORTERM)` (the
-│                       S-Lang/terminfo `truecolor`/`24bit` convention) + `terminal_diagnostic_row(term, colorterm)` (the
-│                       `doctor` `terminal:` line; env values sanitized). main.rs WARN-ONLY (never gates on Unix — many
-│                       truecolor terminals omit COLORTERM; the env read is INLINED at the excluded main.rs call site, no
+│                       S-Lang/terminfo `truecolor`/`24bit` convention) + `terminal_diagnostic_row(term, colorterm, terminfo_caps)` (the
+│                       `doctor` `terminal:` line; env values sanitized; verdict shares the COLORTERM-OR-terminfo signals with
+│                       the warn policy so doctor + the startup warning agree on a TERM-only-truecolor terminal) + the WARN POLICY `should_warn_no_truecolor(colorterm,
+│                       term, suppress_env, terminfo_caps)` (#397: suppresses on $COLORTERM truecolor OR terminfo Tc/RGB for
+│                       $TERM OR a truthy $PIXTUOID_NO_TRUECOLOR_WARN — every input passed in / terminfo lookup INJECTED as a
+│                       closure, so the policy is pure + unit-tested; `terminfo_boolean_caps` is the real `infocmp` seam,
+│                       None on any failure = graceful degrade). main.rs WARN-ONLY (never gates on Unix — many
+│                       truecolor terminals omit COLORTERM; the env reads are INLINED at the excluded main.rs call site, no
 │                       untestable wrapper to mutate) on a non-windows Run-TUI tty. Windows hard-gates VT separately
-│                       (tui/mod). `floating` is exempt (softbuffer = real RGB px). #397 = terminfo Tc/RGB follow-up
+│                       (tui/mod). `floating` is exempt (softbuffer = real RGB px).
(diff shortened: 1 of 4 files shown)


IvanWng97:
Thanks for digging into this, @mvanhorn — your terminfo `Tc`/`RGB` investigation here was what kicked off the rethink.

We ended up taking a different route in #413 (now merged): instead of reading terminfo via `infocmp`, pixtuoid **asks the terminal directly** with a `DECRQSS` SGR query — set an unlikely 24-bit background and check whether the exact RGB triple echoes back. That sidesteps the cases where `infocmp` is absent (musl/alpine/Nix) or stale (stock macOS `tmux-256color` has no `Tc`), with no subprocess and no terminfo dependency. Your `$PIXTUOID_NO_TRUECOLOR_WARN` suppression hatch idea carried over verbatim.

Closing as superseded by #413 — but genuinely appreciate the report and the investigation that pointed the way. 🙏

PR #426, opened 2026-07-02
feat(site): OPEN FLOOR — the live office IS the page

IvanWng97:
## What

PR B of the live-office effort (#422 shipped the engine). It landed in two arcs:

**Arc 1 — the live backdrop.** The real pixtuoid office (the wasm module driven by the real `Reducer`) runs `position: fixed` behind the entire page. Poster-first progressive enhancement: reduced-motion / no-JS / no-wasm / fetch-failure all stay on the still; rAF pauses on tab-hide; ~30fps cap; every FFI contract from #422 honored (`Date.now()` calendar time, ptr/len re-read per frame). CSP gains `'wasm-unsafe-eval'` (narrower than `'unsafe-eval'`). **ONE continuous office** (user-pinned): the hero's Dense variant, seed 3, one timeline that never resets while you scroll — the earlier per-floor–swap design in this PR's history was superseded.

**Arc 2 — the OPEN FLOOR redesign** (full-page, user-licensed "bold redesign with live in background"):

- **The dimmer** — scrolling is the light switch. One rAF controller slaves a fixed `--bg` sheet (0 → 0.86) to viewport-center proximity of `[data-lit]` blocks; `data-lit="fade"` blocks rise with the darkness as ONE gesture. Compositor-only (a single `style.opacity` write), no `backdrop-filter` anywhere (the flicker class — twice removed, third instance found hiding in the nav).
- **The statusline** — the app's status bar as the page's one piece of fixed chrome, on `--screen`, never dimmed: floor LED + scrollspy (absorbed the Elevator, incl. keys 1–6), the build-time real-merged-PR feed (absorbed the Ticker; bots filtered), `lights %` / local clock / `● LIVE` truth-light.
- **The title card** — "Your agents / work here█" monumental Jersey-10, bottom-left film-title style; corner nav, no bar.
- **Observation gaps** — three 100svh holds where the dim releases to 0 and the office alone is the content. Gap 2 is the **frameless-window beat**: the real floating window is an opaque chrome-less rect of office pixels (softbuffer XRGB, no alpha), so the beat shows the committed day/night office still at window scale on a deep shadow — and swaps still + caption word with the viewer's clock.
- **Sections rebuilt as floors** (6F→1F): channel dial (JS contract preserved), features ledger, three monumental how-steps, tools strip with per-CLI badge chips, "Move them in." install climax, one-line footer.
[...]

PR #428, opened 2026-07-02
ci: raise the claude-review turn cap 30 → 60

IvanWng97:
## What

One line: the review bot's `--max-turns` goes 30 → 60. PR #426 (a redesign-scale 12-commit diff) exhausted 30 turns mid-analysis — the action errored "Reached maximum number of turns (30)", a red check with zero findings delivered. `--max-budget-usd 10` is unchanged and remains the real cost bound.

## Why its own PR

The action's anti-tamper token exchange requires a PR's workflow file to match the default branch's content, so a PR that edits `claude-review.yml` can never bot-review itself (401) — the bump only takes effect for other PRs once it lands on main. Expect the `review` check on THIS PR to be red for exactly that reason (the job still triggers; the 401 happens inside the action); every other gate applies normally.

## Note

The `review` check is not in main's required set — this is quality-of-life for big diffs, not an unblocking change.

https://claude.ai/code/session_0121vF6GrV7TEXC3H8eR5wBw

### DIFF
--- a/.github/workflows/claude-review.yml
+++ b/.github/workflows/claude-review.yml
@@ -47,4 +47,4 @@ jobs:
           prompt: ${{ steps.prompt.outputs.content }}
           track_progress: true
           show_full_output: true
-          claude_args: '--max-turns 30 --max-budget-usd 10 --allowedTools "mcp__github_inline_comment__create_inline_comment,Bash(gh pr comment:*),Bash(gh pr diff:*),Bash(gh pr view:*),Read,Glob,Grep"'
+          claude_args: '--max-turns 60 --max-budget-usd 10 --allowedTools "mcp__github_inline_comment__create_inline_comment,Bash(gh pr comment:*),Bash(gh pr diff:*),Bash(gh pr view:*),Read,Glob,Grep"'


IvanWng97:
## Two-lens review — both APPROVE

**Correctness (95):** actionlint-clean; the failed 30-turn run took <3 min wall time, so 60 turns extrapolates to ~6 min against the 15-min job timeout — the timeout and the unchanged `--max-budget-usd 10` remain the hard bounds. Verified live that `review` is not in main's 12 required contexts. One prose nit (the check goes *red*, not absent — job `if:` still fires) — PR body corrected.

**Design (no nits):** right lever — the review prompt's turn-costly requirements (Read-before-cite, tree-wide DRY search, beyond-diff escalations) are incident-backed quality policy; slimming them trades findings for turns. A turn-capped run's failure mode is a silently truncated review of exactly the PRs that most need one, which is the owner's per-push-review policy failing at its purpose. 60 = evidence-proportioned 2×; alternatives (per-PR override, 100) rejected as YAGNI. Blast radius clean — nothing in the tree cites "30 turns"; the three claude workflows' caps are independent per-job budgets.

PR #429, opened 2026-07-02
refactor(scene): hoist the shared headless render seam into pixtuoid-scene

IvanWng97:
Closes #423.

## What

The "scene → RgbBuffer + bookkeeping epilogue" existed in THREE places kept in sync only by convention (TUI floor-slide, floating `OfficeRenderer`, web `Office`) — and the convention failed twice for real (the TUI's dropped-carriers bug; the web loop-2 teleport). `pixtuoid-scene` now owns the frame:

- **`floor::render_floor`** — prologue (buffer sizing / layout / router zone) + the pixel pass + the epilogue (coffee-carrier persistence, door-anim clamp), compiler-owned. Returns the layout (callers cache it for labels/hit-testing) or `None` on an unlayoutable size.
- **`floor::CoffeeState`** — holders + fetched_at with `evict_missing`/`record` (first-insert-wins steam stamp, the TUI's edge semantics).
- **`FloorCtx::evict_missing`** — the cache/history/motion trio.

The three headless consumers are thin callers; `render_transition_floor` is deleted. The TUI's normal `draw_scene` path stays on raw `render_to_rgb_buffer` **by design** (needs the full `PixelPassResult` + holds immutable coffee borrows mid-flush) and routes its bookkeeping through the same shared types — documented as the one exception in the scene CLAUDE.md.

## Byte-identity

`just preflight` 1948/1948 (incl. the ~65-test TUI harness and both prior epilogue-bug pinning tests) · `just gen-check` pixel-diff green on every committed still · 2 new scene pin tests.

## Three deliberate deltas (each more correct, pixel-invisible)

1. **Floating now evicts per-agent state** — it previously never did (slow leak; gone agents were never painted, so no visual change).
2. **Floating's `last_layout` resets on a failed too-small layout** — labels no longer anchor against a stale layout over a cleared buffer.
3. **Web hero's dust motes seed from the layout seed** like every other consumer (was ambient seed 0 over a seed-3 layout — same density, different deterministic positions).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_0121vF6GrV7TEXC3H8eR5wBw

PR #437, opened 2026-07-02
test(site): Playwright e2e smoke suite — the runtime-contract gate

IvanWng97:
## What

A **13-test Playwright smoke suite** pinning the page's **runtime contracts** — the window globals, custom events, and data-attribute wiring where tsc/eslint/knip/build (and the Rust bridge + pixel gates) are all blind.

Two tiers:
- **7 regression pins** for bug classes a human review actually caught this arc: the `pix:onair` race, the `is:inline` parse-position trap, the `nav--floating` docs leak (#426), the `__pixHire` wiring, the dimmer mechanic, the reduced-motion parallel design — plus a console-error watchdog per distinct code path.
- **6 audit-driven tests** (+3 line-level extensions) from a 7-surface sitewide interaction audit that catalogued **91 runtime interactions** and adjudicated each: boot intro first-visit + session gate, the full theme chain (saved/URL/toggle/Escape/system-dark — every step wall-clock-proof), install tabs + both clipboard branches, the showcase studio (deep-links, dial, hydration, chips, play policy), nav menus + docs TOC scrollspy + 404 + mobile burger, and the landing's fixed chrome (floating nav, statusline readouts, floor popover, gap-2 day/night agreement). 49 interactions deliberately skipped with recorded reasons (wall-clock flake, ops-only, easter eggs, covered-elsewhere).

Per the official Astro posture: runs against the **production build** via `astro preview` (never the dev server, whose Vite cache has twice masqueraded as a site bug here).

## Two real catches before it even merged

1. **A production CSP bug (day-one watchdog catch)**: Vite's default 4KiB inlining turned small @fontsource subsets into `data:` fonts — silently blocked by the hand-rolled `font-src 'self'` in production (24 of 154 woff2 files were affected). Fixed strict-side: `assetsInlineLimit: 0` (assets stay files; CSP stays `data:`-free; post-fix dist has zero `url(data:)`).
2. **The orphaned-preview trap (caught by the suite's own teeth test)**: `npm run preview` as the webServer leaves an orphaned astro child squatting the port when Playwright kills the wrapper; `reuseExistingServer` then silently tests the orphan's **stale build**. The config runs the astro bin directly + pins `reuseExistingServer: false` — a squatted port fails loud.

## Wiring

`just site-e2e` (build + test) · `just site-setup` installs the Chromium browser · site.yml runs e2e after the build step · artifacts gitignored · window contracts declared in `env.d.ts` (load-bearing: without it the spec fails 4× TS2339) · gates documented in site/README.md, site/CLAUDE.md + the workspace map.

## Review

Two-lens local review (correctness/flake-audit + design/blast-radius briefs): both **APPROVE-WITH-NITS**, all 9 findings fixed in `e90342c3` — dispositions in the PR thread.

## Verification

| Check | Result |
|---|---|
| Playwright suite (×2 consecutive — no orphan, no port clash) | 13/13 both, ~8s |
| Teeth (mutate `__pixHire` in dist) | contracts test reds |
| `just site-check` | exit 0 |
| `just preflight` (pre-push hook) | 1951/1951 |

https://claude.ai/code/session_0121vF6GrV7TEXC3H8eR5wBw

### DIFF
--- a/.github/workflows/site.yml
+++ b/.github/workflows/site.yml
@@ -52,6 +52,13 @@ jobs:
       # `verify` script; runs before build for fast feedback.
       - run: npm run knip
       - run: npm run build
+      # e2e smoke suite vs the PRODUCTION build just produced (astro preview,
+      # the official Astro testing posture): pins the page's runtime contracts
+      # — wasm office goes live, statusline truth-light, digit-key scrollspy,
+      # dimmer release, Copy→hire wiring, docs-nav variant, reduced-motion
+      # fallback — plus a console-error watchdog (it caught the CSP-blocked
+      # inlined fonts on day one). Chromium is already installed above.
+      - run: npm run e2e
       # Lighthouse CI: gate a11y / SEO / best-practices (perf advisory) against
       # the built pages — serves dist via `astro preview` (honors the /pixtuoid
       # base) and audits each. Replaces the dropped eslint-plugin-jsx-a11y with a
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -78,7 +78,7 @@ scripts/             gen-media.py + media.json (the ONE manifest-driven driver f
                      account/gateway footprint, NOT a CI test),
                      check_upstream_drift.py (weekly wire-format watch),
 site/                Astro landing page → GitHub Pages; self-contained Node project,
-                     own CI; `just site-{setup,dev,check,fmt}` → see site/README.md
+                     own CI; `just site-{setup,dev,check,fmt,e2e}` → see site/README.md
 integrations/raycast/  Raycast extension (TypeScript, self-contained Node project; NOT Rust):
                      `Manage Sources` (connect/disconnect over `pixtuoid sources|connect|disconnect
                      --json`) + `Start Floating` commands. A thin shell over the CLI `--json`
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -15,7 +15,8 @@ live in nested `CLAUDE.md` files**, auto-loaded when you touch those trees:
   - [`crates/pixtuoid/src/tui/CLAUDE.md`](crates/pixtuoid/src/tui/CLAUDE.md) — the terminal painter (over the `pixtuoid-scene` crate): draw_scene flush, harness, widgets, the theme-PICKER ui, Sources panel, dashboard, hit_test, version popup.
 
 The NON-Rust **consumers** of the `--json` contract have their own guides (their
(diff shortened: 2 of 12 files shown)


IvanWng97:
## Two-lens review round — closed

**Lens 1 (correctness/flake-audit): APPROVE-WITH-NITS · confidence 92.** Verified empirically: wall-clock/TZ independence of every assertion (the clock-dependent surfaces — theme init, statusline clock, gap-2 — are asserted by *shape or agreement*, never value); digit-keys serialization (no race); `expect.poll` margins vs measured runtime (~50×); the CSP fix end-to-end (24/154 @fontsource woff2 were under Vite's 4KiB inline default → `data:` fonts blocked by `font-src 'self'`; post-fix dist has **zero** `url(data:)`); webServer orphan behavior (two consecutive runs, no listener left); knip's built-in Playwright plugin auto-detection; and that the `env.d.ts` Window augmentation is **load-bearing** (without it the spec fails 4× TS2339).
→ **Fixed (e90342c3):** the reduced-motion 2s dead wait replaced with a deterministic network-idle + zero-`/wasm/`-requests assertion (test now 0.7s); `watchErrors` scoping rule documented.

**Lens 2 (design/blast-radius): APPROVE-WITH-NITS · confidence 92.** Confirmed each regression-pin's provenance against the git/PR record; adjudicated `assetsInlineLimit: 0` as a perf **win** with zero collateral (the un-inlined class is exactly the broken class); endorsed e2e staying out of `just verify` (Lighthouse precedent, double-build, setup gap).
→ **Fixed (e90342c3):** provenance comment `#429`→`#426`; docs-page console watchdog; `site-setup` installs Chromium; the false "4399 dev habit" comment replaced; README quality-gates + seams (`__pixHire`) currency; workspace CLAUDE.md `e2e` token.

## Sitewide interaction audit → extend round (3a046382)

A 7-surface fan-out catalogued **91 runtime interactions** and adjudicated each for the smoke tier: **6 new tests + 3 line-level extensions**, 49 deliberate skips (wall-clock flake / ops-only / easter eggs / covered-elsewhere), each with a recorded reason. Suite is now **13 tests, ~8s**, 13/13 twice locally; site-check green.

Notable adjudications: the wasm fetch-failure fallback stays OUT (structurally prevented by `gen-wasm-check`; proving the negative costs a fixed 2s every run); `visibilitychange` pause stays out (`document.hidden` monkey-patch vs a 30fps loop = the flakiest machinery in the catalogue); the mobile burger folded into the docs/nav test rather than its own project.

PR #455, opened 2026-07-03
fix(site): whole-site audit remediation — a11y, responsive, no-JS & cross-browser robustness

IvanWng97:
## Origin

Started from a reported hero glitch ("the office is wider on load, then adjusts"). Investigation found two *separate* things, then widened into a full re-audit:

1. **Hero reframe** — the live wasm canvas computed its buffer from the window aspect, so it only matched the fixed 16:9 poster at 16:9 windows; every other aspect reframed on the poster→canvas crossfade. After seeing both options rendered, we **kept native-fit** (best framing, no crop) and accepted the one-time load reframe — so this is intentionally *not* "fixed" here.
2. **A real mobile bug it exposed** — `[data-lit]::before` (the readability glow, `inset -8%`) made full-width `<section>`s ~8% wider than the viewport. `body{overflow-x:hidden}` hid it on desktop, but on **mobile it panned the page left-right**. Fixed with `overflow-x: clip` (commit 1).

That prompted a **6-lens re-audit** (responsive / a11y / JS-correctness / perf-CSP / cross-browser / links-security), since #453's "whole-site review" scanned its own diff, not the pre-existing surface. Commit 2 is the remediation.

## Fixes (commit 2)

**Responsive (WCAG 1.4.10)**
- Studio dial wraps to 2 cols <480px + `.studio` track `minmax(0,1fr)` — all 8 channels stay tappable on phones (03/06 were clipped & unreachable).
- Docs prose tables get their own `overflow-x:auto` + inline-code `overflow-wrap:anywhere` (config was 404px / kb 414px wide at 390px; now 0 page overflow).

**Accessibility**
- **WCAG 2.2.2**: Showcase demo clips now honor `#office-pause` (`pix:paused`). They auto-loop with no controls in normal motion, so in-view-gating didn't satisfy 2.2.2. Corrected the false `CLAUDE.md` claim + added an e2e assertion.
- **WCAG 1.4.3 (day theme)**: install active tab `--led` 1.4:1 → `--coral-deep` 5.2:1; "see it live"/"see how" links `--coral` 2.7:1 → `--coral-deep` 5.2:1; primary CTA gradient light stop `#c96a45` 3.7:1 → `#b5592f` 4.7:1.

**Robustness**
- No-JS content blackout: `<noscript>` keeps `.reveal` sections visible (were `opacity:0` until JS — blank page below the hero with JS off).
[...]

PR #459, opened 2026-07-03
fix(site): #office-pause governs ambient motion independent of wasm (#456)

### LINKED ISSUE #456: site: #office-pause stays hidden on wasm-failure path — auto-motion has no visible pause affordance (WCAG 2.2.2)
### Problem
`#office-pause` (the page-wide WCAG 2.2.2 pause button in `OfficeBackdrop.astro`) is unhidden **only** on the first painted wasm frame (`paint()`). On a **wasm-failure / no-wasm / fetch-blocked load** the office never goes live, so the button stays `hidden` (the documented "poster-only path").

But several >5s auto-motion elements keep animating on that path regardless of wasm:
- the statusline **feed ticker**
- the **hero dust**
- (as of #455) the **Showcase demo clips** (they auto-loop with `controls` removed in normal motion)

So a **non-reduced-motion** visitor whose wasm never loads gets looping auto-motion with **no visible pause affordance** — a 2.2.2 gap.

### Scope / severity
- **Pre-existing** for the ticker + hero dust; #455 extended the same page-pause pattern to the clips (not a regression — it aligns the clip with the existing pattern and strictly improves the common wasm-loads case).
- **Reduced-motion users are fully covered** already: the clips fall back to native `<video>` controls (wasm-independent), and the ambient motion is suppressed.
- Exposed population is narrow: non-reduced-motion **and** wasm never loads **and** wants to pause. Hence low priority, but a concrete completeness gap flagged by the PR #455 two-lens review (design lens).

### Fix sketch
Decouple `#office-pause` visibility from "first wasm frame." Unhide it whenever **any** governed >5s auto-motion is present (ticker / dust / clips), independent of wasm — e.g. unhide on `DOMContentLoaded` when a non-reduced-motion page has any `pix:paused` consumer, rather than inside `paint()`. Pause then still works (it dispatches `pix:paused`; the rAF office loop is simply a no-op when there's no live canvas).

### Refs
- PR #455 two-lens review, design lens, finding 2 (`Showcase.astro:152`, `OfficeBackdrop.astro` `paint()`).
- Prose caveated in `Showcase.astro` + `site/CLAUDE.md` in #455 to document the gap until this lands.

IvanWng97:
Closes #456.

## Problem
`#office-pause` (the WCAG 2.2.2 pause switch) was unhidden **only by the first painted wasm frame**. On the **no-wasm / fetch-failure** poster paths it stayed hidden — but the statusline feed ticker, hero dust, and (since #455) the Showcase demo clips keep auto-animating there, **independent of wasm**. So a non-reduced-motion visitor whose wasm never loaded had **no visible pause affordance** for that motion. Pre-existing for the ticker/dust; #455 extended the same page-pause pattern to the clips, which is why it was tracked as a follow-up rather than fixed inline.

## Fix — decouple the pause *control* from the office *canvas*
- The early return no longer bails on `typeof WebAssembly === 'undefined'`; only the office **render** path (`boot`/`paint`) is gated on a new `hasWasm`. `setPaused`/`start`/`stop` are no-ops without a live office, so the control is safe standalone.
- `syncPauseBtn()` shows `#office-pause` whenever motion runs = **NOT reduced-motion**, independent of wasm, and hides it under reduced-motion — one rule, called at init and on every reduced-motion change. The wasm-frame unhide is removed.
- Clicking it still fires `pix:paused`, which the ticker / dust / clips already consume, and the statusline reads `❚❚ PAUSED` — even with no live office.
- Reduced-motion is unchanged (button hidden; clips keep native `<video>` controls).

## Verification
- **e2e teeth**: the wasm-fetch-failure test now asserts `#office-pause` is **visible** and a click fires `pix:paused{true}` + flips `aria-pressed` (it previously asserted *hidden* — which encoded the bug). The reduced-motion test still asserts hidden.
- **Rendered** the aborted-wasm poster: button shows bottom-right, click → statusline `○ STILL` → `❚❚ PAUSED`.
- `just site-check` EXIT=0 · `just site-e2e` **18/18**.
- Docs: `site/CLAUDE.md` updated — the old "hidden on every poster-only path" text and the #455 follow-up caveat are both replaced with the decoupled behavior.

### DIFF
--- a/site/CLAUDE.md
+++ b/site/CLAUDE.md
@@ -60,9 +60,7 @@ Mermaid diagram becomes an inline SVG at build via `rehype-mermaid`, which is
   The backdrop's pause switch (`#office-pause`, WCAG 2.2.2) lives in the same
   component: pause stops the rAF loop (frozen frame stays on the canvas) and
   resume subtracts the paused span from the sim clock (`pauseOffset`) so the
-  timeline doesn't lurch-jump; the button stays `hidden` on every poster-only
-  path (reduced motion / no wasm / fetch failure) and is unhidden by the
-  first painted frame. Pause is **page-scoped**: `setPaused` dispatches
+  timeline doesn't lurch-jump. Pause is **page-scoped**: `setPaused` dispatches
   `pix:paused` and every other >5s auto-motion listens (the statusline feed
   ticker, the hero dust) — one control governs the page's *ambient* motion,
   and the statusline reads `❚❚ PAUSED`. The Showcase demo clips ARE in the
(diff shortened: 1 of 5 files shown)


IvanWng97:

Two differentiated lenses — **correctness/grounding + state-machine** and **design/blast-radius + render-and-watch** (built the production site, drove Chromium across all six states). Both **APPROVE-WITH-NITS**. Adversarial verify on the one MEDIUM.

**No runtime bugs.** The state-machine change was traced clean across `{live, no-wasm, fetch-fail, reduced-motion, un-reduce, tab-hidden} × paused`:
- removing `WebAssembly===undefined` from the early return safely reaches the pause wiring; every post-return statement that could deref a null office is guarded (`start`/`paint` gated on `office`, `deferBoot`/`boot` on `hasWasm`, resize's paint on `live`, `__pixHire` only defined inside `boot`).
- `setPaused` is no-op-safe without an office (no NaN — `pausedAt` always set before `pauseOffset` reads it); pause-before-boot correctly freezes via `start()`'s `!paused` guard, no lurch.
- `syncPauseBtn` runs in both mq branches; `hasWasm`/`syncPauseBtn` have no TDZ issue; the e2e `pix:paused` listener is race-free.

**Rendered evidence (design lens):** wasm-fail → button visible (fixed right:10 bottom:42, 36×36, z-50, not clipped), click flips onair STILL→PAUSED + aria-pressed + freezes the `multi-floor.webm` clip (currentTime pinned across 700ms samples); reduced-motion → hidden + native controls; live → office up in 154ms, seamless pause/resume, **CLS 0.0000388** (button is `position:fixed`, the earlier reveal causes no shift); wasm-slow → pausing suppresses the office coming live then resume boots it (correct WCAG behavior); no dimmer/`__pixHire` interaction issues.

Every finding was one class — **stale comments the fix invalidated but I didn't sweep** (I skipped the repo-wide stale-phrase grep). All FIXED in `9a486046`:

| # | Finding | State |
|---|---------|-------|
| 1 | `Showcase.astro` `pix:paused` comment still said `#office-pause` "never unhides on wasm-fail / tracked separately" (MEDIUM, verify-**confirmed real**) | **FIXED** |
| 2 | `site/README.md` FX line "hidden whenever only the still poster is showing" (LOW) | **FIXED** |
| 3 | `smoke.spec.ts:140` live-path comment "unhidden by the first painted frame" (NIT) | **FIXED** |
| 4 | `smoke.spec.ts:230` reduced-motion comment conflated reduced-motion with the now-visible wasm-fail poster (**grep-caught — the review missed this one**) | **FIXED** |

The repo-wide grep for `never unhides|poster-only path|first painted frame|tracked separately|still poster is showing|nothing to pause` now returns **clean**.

**Gates on the new head (`9a486046`):** `just site-check` EXIT=0 · `just site-e2e` **18/18**.

PR #475, opened 2026-07-04
test(scene): reference shared disc visibility threshold

### LINKED ISSUE #473: sky.rs test hardcodes MIN_DISC_VIS=0.08 instead of the authoritative celestial::MIN_DISC_VIS
**Pre-existing DRY nit** surfaced by the #469 celestial extraction (PR #472) — non-blocking, deferred out of that move-only PR.

**Problem.** The disc-hide threshold `MIN_DISC_VIS = 0.08` is defined authoritatively in `crates/pixtuoid-scene/src/pixel_painter/background/celestial.rs` (it's `compute_disc`'s "thick cloud swallows the disc" gate). But the `thick_cloud_hides_the_disc_uniformly` test in `background/sky.rs` hardcodes a **second** local `const MIN_DISC_VIS: f32 = 0.08;` to assert that Overcast/Rain/Storm's `atmo().disc` values all sit below the gate. Two copies of the same domain threshold = latent drift: bump the gate in `celestial.rs` and the sky.rs test keeps asserting against a stale `0.08`, silently stopping guarding the real "atmo disc < hide threshold" relationship.

(This predates #469 — `MIN_DISC_VIS` lived in `background/mod.rs` before the move, and the sky.rs test already chose to hardcode rather than reference it. The move just relocated the authoritative copy to `celestial.rs`, where it is currently `private`.)

**Fix sketch (decide the cleaner direction first).** Two options, each with a wrinkle:
1. Have the sky.rs test read `super::celestial::MIN_DISC_VIS` — requires re-exposing it as `pub(super)` in `celestial.rs`, and creates a **test-only** `sky → celestial` back-dependency (prod dep is `celestial → sky::atmo`, so this is a reverse edge, though test-only and acyclic at the type level).
2. Move the threshold's *home* to `sky.rs` (atmo/disc-visibility is arguably sky's domain) and have `celestial::compute_disc` import it — no back-dep, but relocates a const that reads naturally as a disc concern.

Either way, pin the two sites together (one authority + one reference, or a `debug_assert!`/test that they match). Weigh which module should own the threshold before wiring.

Labels: tech-debt, priority: low.

zerone0x:
## Summary
- expose the celestial disc visibility threshold within the background module
- update the sky test to assert against that shared threshold instead of a duplicated literal

Fixes #473

## Test plan
- `cargo fmt --check`
- `git diff --check`
- `cargo test -p pixtuoid-scene thick_cloud_hides_the_disc_uniformly` (blocked: linker `cc` is not installed in this environment)

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/crates/pixtuoid-scene/src/pixel_painter/background/celestial.rs
+++ b/crates/pixtuoid-scene/src/pixel_painter/background/celestial.rs
@@ -67,7 +67,7 @@ const HORIZON_FRAC: f32 = 0.55; // horizon_y = top_wall_h * HORIZON_FRAC
 const ARC_RISE_FRAC: f32 = 0.80; // apex lifts top_wall_h * ARC_RISE_FRAC above horizon
 /// Below this atmo `disc` visibility, thick cloud swallows the disc entirely
 /// (no point painting a body no one can see through the murk).
-const MIN_DISC_VIS: f32 = 0.08;
+pub(super) const MIN_DISC_VIS: f32 = 0.08;
 
 /// This frame's disc placement, or `None` under thick cloud (`atmo(weather).disc`
 /// below [`MIN_DISC_VIS`]). `cx`/`cy` are absolute buffer coordinates derived
--- a/crates/pixtuoid-scene/src/pixel_painter/background/sky.rs
+++ b/crates/pixtuoid-scene/src/pixel_painter/background/sky.rs
@@ -885,7 +885,7 @@ mod tests {
         // show MORE of the disc than a thinner one (Rain), matching the
         // direct/diffuse ordering `storm_transmits_less_than_rain_overall`
         // already pins.
-        let min_disc_vis = super::celestial::MIN_DISC_VIS;
+        let min_disc_vis = crate::pixel_painter::background::celestial::MIN_DISC_VIS;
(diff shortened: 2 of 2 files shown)


IvanWng97:
Thanks @zerone0x — the direction here is exactly right (Option 1 from #473: expose `celestial::MIN_DISC_VIS` as `pub(super)` and reference it from the sky test instead of a duplicated `0.08`). 👍

One thing the "linker unavailable in my env" caveat hid: the reference didn't compile. From inside `sky::tests`, `super` is the `sky` module — which has no `celestial` submodule (`celestial` is a *sibling* of `sky` under `background`), so `super::celestial::MIN_DISC_VIS` was `error[E0433]: cannot find celestial in super`.

I pushed a one-line fix commit on top of yours (kept your commit intact) that references it by the crate-absolute path, robust to future module moves:

```rust
let min_disc_vis = crate::pixel_painter::background::celestial::MIN_DISC_VIS;
```

Verified locally: `cargo test -p pixtuoid-scene thick_cloud_hides_the_disc_uniformly` passes, and the full `just preflight` gate is green. Ready to merge once CI confirms.

PR #501, opened 2026-07-06
fix(site): stop docs-table code columns collapsing to 1ch

IvanWng97:
## Problem

On the rendered docs pages (e.g. `/config` "User settings"), a markdown table column whose cells are bare code tokens collapses to ~1 character wide — `theme` wraps letter-by-letter, each fragment drawing its own inline-code pill.

## Root cause

`global.css` gives inline code `overflow-wrap: anywhere` so long paths can't overflow the mobile viewport (`body{overflow-x:hidden}` would clip them unreachably). Unlike `break-word`, `anywhere` feeds its soft-wrap opportunities into **intrinsic sizing**: the code span's min-content contribution drops to 1ch, and table auto-layout crushes the code-only Key column while the long Description prose takes all the width.

## Fix

Exempt code inside table cells (`.prose th code, .prose td code { overflow-wrap: normal; white-space: nowrap }`). Inside tables the guard is redundant — `.prose table` already has `display:block; overflow-x:auto`, so wide content scrolls within the table. `nowrap` additionally prevents the hyphen-break (`max-` / `desks`) a squeezed column would otherwise take.

## Verification

- Live render at 1440px + 375px across all five doc routes: zero code spans line-wrap (`getClientRects().length > 1` sweep), no page-level overflow, keys render as single pills.
- Worst-case token (28ch `$XDG_STATE_HOME/pixtuoid/log`, logging table) fits with ~12px of internal scroll at 375px.
- `just site-check` green; prettier clean.
- Pre-existing, unrelated: `/parallel-delivery` reports ~1px body overflow at 375px from a Shiki ASCII-diagram `pre` (scrolls internally; no tables on that page).

https://claude.ai/code/session_019JojEsdMrzu5VVt4vhyBvC

### DIFF
--- a/site/src/styles/global.css
+++ b/site/src/styles/global.css
@@ -822,9 +822,13 @@ code,
    redundant (the table's own overflow-x:auto handles wide content) and harmful:
    `anywhere` shrinks min-content to 1ch, so table auto-layout collapses a
    code-only column and wraps its token letter-by-letter. nowrap also stops the
-   hyphen-break (`max-` / `desks`) a squeezed column would otherwise take. */
-.prose th code,
-.prose td code {
+   hyphen-break (`max-` / `desks`) a squeezed column would otherwise take.
+   Tradeoff: a genuinely long, break-point-free token in a cell now scrolls the
+   table rather than wrapping — such tokens belong in a fenced block, not a
+   table cell. (`table` in the selector lifts specificity to (0,1,3) so the
+   override beats the (0,1,2) guard above by weight, not source order.) */
+.prose table th code,
+.prose table td code {
   overflow-wrap: normal;
   white-space: nowrap;
 }
--- a/site/src/styles/global.css
+++ b/site/src/styles/global.css
@@ -818,6 +818,16 @@ code,
   font-weight: 400;
   font-synthesis: none;
 }
+/* inside tables, the inline-code `overflow-wrap:anywhere` guard above is both
+   redundant (the table's own overflow-x:auto handles wide content) and harmful:
+   `anywhere` shrinks min-content to 1ch, so table auto-layout collapses a
+   code-only column and wraps its token letter-by-letter. nowrap also stops the
+   hyphen-break (`max-` / `desks`) a squeezed column would otherwise take. */
+.prose th code,
+.prose td code {
+  overflow-wrap: normal;
+  white-space: nowrap;
+}
 /* architecture diagram (```mermaid → build-time inline SVG): frame it in a fixed
    light "terminal screen" so the day-palette SVG stays legible on night/dracula,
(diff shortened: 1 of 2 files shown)


IvanWng97:
## Two-lens review — disposition sweep (head c70494b9)

Three local lenses (correctness, design/blast-radius, rendered-runtime — the CSS-renders escalation) + the online review. Verdicts: APPROVE-WITH-NITS / APPROVE-WITH-NITS / APPROVE / Findings: 1 (MEDIUM).

| # | Finding | Source | State |
|---|---------|--------|-------|
| 1 | No regression test pins the fix (pan guard at `smoke.spec.ts:735` is blind to a column collapse) | online MEDIUM + lens 1 | **FIXED** — c70494b9 adds the column-collapse guard (`/config` @390px, every `.prose table th/td code` renders one line box); proven **red** against base 3c72f4fd (1 failed), green with the fix |
| 2 | Override wins by source order at equal specificity — a rule reorder silently regresses | lens 1 | **FIXED** — c70494b9 lifts the selector to `.prose table th/td code` (0,1,3) > (0,1,2), order-independent |
| 3 | Comment doesn't state the forward boundary (long break-point-free tokens now scroll instead of wrapping) | lens 2 N1 | **FIXED** — c70494b9 appends the tradeoff clause ("such tokens belong in a fenced block, not a table cell") |
| 4 | Alternative: swap the base rule to `overflow-wrap: break-word` and drop the exemption | lens 2 N2 | **REFUTED** — not strictly dominant: `break-word` (no nowrap) breaks space-containing commands (`just bump X.Y.Z`) mid-command, and re-verifying every prose paragraph across five docs is a far wider blast radius than the shipped table-only surface |
| 5 | Pre-existing mobile overflow: `/parallel-delivery` Shiki block 151–206px @375/320 + `nav__toggle` 5px @768 (byte-identical on baseline — not this PR) | lens 3 | **ISSUE-FILED** — #503 |

Independently verified by the lenses: root-cause mechanism spec-accurate (CSS Text L3: `anywhere` counts soft-wrap opportunities in min-content, `break-word` doesn't); no unpatched sibling (`Install.astro:241`'s `anywhere` targets a block `pre` — different class); `.prose` blast radius = the five doc routes only (`Docs.astro:63`); `.tools` matrix already uses the same nowrap+internal-scroll pattern (precedent, not duplication); day 12.96:1 / night ~7.7:1 code-pill contrast; no-JS and no-color-mix engines unaffected; extreme-content probe (60ch token) scrolls internally with zero page overflow.

Gates at head: full Playwright suite 21 passed; `just site-check` exit 0; gen-check unaffected (CSS-only, no office-look change).

https://claude.ai/code/session_019JojEsdMrzu5VVt4vhyBvC

PR #461, opened 2026-07-03
Whole-codebase architecture sweep (0.12.0): 22 behavior-preserving refactors — dedup, make-illegal-unrepresentable, param-grouping

IvanWng97:
Whole-tree architecture pass taken while 0.12.0 is still unreleased: dedup,
make-illegal-states-unrepresentable, co-related-parameter grouping, and module
decomposition — all **behavior-preserving** (gen-check byte-identical,
2044/2044 tests green). 22 refactors + 1 review round = 23 commits.

## What changed

**pixtuoid-core (5)**
- `IdentityCtx` wraps the `(source, session_id, cwd)` triple threaded through 4
  reducer sites (kills a two-`&str` positional-swap footgun, drops a
  `too_many_args` allow); `apply_session_start` extracts the ~245-line
  SessionStart arm.
- `derive_prefixed_label` reads the label prefix from the registry authority
  instead of re-hardcoding it in each JSONL deriver.
- `first_present_str` centralizes the tool-arg vocab scan across all 6 hook
  decoders (the review round completed the 6th — antigravity).
- `ProbeSnapshot` collapses `ids` + `pid_of` → one `pid_of` map (id-without-pid
  now unrepresentable).

**pixtuoid-scene (10)**
- `WanderTarget { dest, kind: Named{wp_idx,kind,seat} | Aimless }` replaces 4
  flat wander fields (dead "mixed" branch gone); `resolve_wander_target` is the
  ONE stateless/stateful resolver; `measured_leg_len` routes the 5 leg-length
  sites.
- `FloorVariant` enum replaces the 4 mutually-constrained room-presence bools.
- `stand_point` delegates to `approach_point` (render-anchor ≡ walk-goal by
  construction); `layout::coarse` hosts the shared coarse-grid primitives both
  A* and the reach BFS ride (was 2 hand-kept copies + a compile-assert).
- `epoch_ms` forwarder, `RouteCtx` rebuild-drop, `OUTFITS` single-pool,
  `resolve_characters` seated-closure.

**pixtuoid + pixtuoid-web (7)**
- `ConnState::NoCli { connected }` (illegal enum+bool combo unrepresentable);
  Claude hook merge delegates to the shared `flat_json_merge_*`;
  `install/verify.rs` write helpers split into `install/merge.rs`.
- onboarding dim now applies on the floor-transition path (`apply_dim` shared
  helper) — the one intentional behavior fix.
- web `Office` hire state grouped into `VisitorHires` / `ScheduledEvent`.
[...]

PR #350, opened 2026-06-17
refactor: comprehensive smell-audit cleanup — Grid collapse + 17 dedups + duplication review lens (0.11.0)

IvanWng97:
Follow-up to #349 that grew (by request: *fix everything, no deferrals*) into a comprehensive 0.11.0 cleanup. A burden-flipped whole-repo **smell audit** (default = smell unless justified) found 63 findings — all personally verified by reading the code — plus an exhaustive multi-lens **duplication hunt**. Every actionable finding is fixed here; every KEEP is cited.

## What landed

**Grid family (semver-breaking → 0.11.0)**
- `Frame` and `RgbBuffer` are now newtypes over `Grid<T>` (the row-major buffer `#333` introduced). `Deref` keeps `.width`/`.height`/`as_slice` transparent; `RgbBuffer` keeps its **own inherent unchecked `get`/`put`** for the blit hot path; `pub pixels`/`width`/`height` fields removed. Byte-identical render (golden + pixel-regression).

**17 deduplications** (one impl each, behavior-preserving)
- `lerp_rgb`→`mix_lab`, two `expand_tilde`, sanitizer×3, `SENTINEL_KEY`×5, FNV consts, `make_floor_info`, `render_plugin`, `parse_or_empty`, argv-flag parser, pet/mascot walk, `elapsed_ms`, `dist2`, `generic_tool_display`, `hook_path_str`×7, watchdog spawn, `settle_from_pair`, the wander-arm scaffold, `paint_overlays`, `SourceColors::by_prefix`, `floor_capacity`, spot-anchors, `wash_glass`, coffee-carrier merge, `SourceArgs` flatten, two `run_tui` extractions, a test `agent_slot` builder.

**10 smell fixes** incl. the Windows `--pack-dir ~\` expansion fix, `println!`→`tracing` under the alt-screen, `[02]` single `derive_with_routing` per agent/frame (byte-identical), and doc(hidden)/cfg(test) hygiene.

**Process** — a duplication/DRY lens added to the PR-bot rules, the two-lens design brief, and the whole-codebase SYSTEM lens (a diff-scoped review can't see a pre-existing copy, so the lens *requires* grepping outside the diff).

**28 KEEPs**, each cited (a refuted-recheck pass confirmed none were dodged behind "deliberate"): invariant #3 per-source modules, uniform fn-ptr signatures, per-CLI install mirroring, documented sharp edges, YAGNI. Full adjudication trail in `docs/REVIEW-LEDGER.md` (R0617-50-*).

[...]

PR #524, opened 2026-07-10
refactor: slim dead/duplicated fallback safety-nets across the workspace

IvanWng97:
## What

A whole-codebase fallback/兜底 (safety-net) audit (146 sites across the 5 Rust crates) found the defensive logic **~90% load-bearing** (liveness ladders, never-panic decoder/shim contracts, clock-regression guards, config-never-wipe, platform path precedence, render-edge clamps). This PR lands the genuinely-slimmable subset — **all behavior-preserving except two dead-in-prod removals** — and leaves every documented/test-pinned guard intact.

## Changes

**Dead defaults (the desk one also fixes a latent bug):**
- `pixel_painter`: desk footprint `unwrap_or(Size{DESK_W,DESK_H})` was dead **and wrong** (`decor.rs` is `DESK_W+4`, so a hypothetical `None` mis-placed the sprite); Lounge/Pantry table `(7,4)` `map_or` dups — now read `furniture_def(x).footprint` via `let-else` (the `mask.rs` idiom), no re-hardcoded literal.

**DRY-centralizations (guards preserved, one source of truth):**
- `correlation.rs`: 8× `duration_since(ts).is_ok_and(|d| d<TTL)` → `is_fresh()` (exact-boundary tombstone tests still pin strict `<`)
- `platform.rs`: 4× trim-nonempty closure → `fn nonempty`
- `pixel_painter`: 4× out-of-range frame guard (two spellings) → `frame_at` (guard stays — custom `--pack-dir` packs)
- `copilot.rs`: subagent child-key ladder ×2 → `copilot_child_key`
- `pose/pure.rs`: `corridor.unwrap_or(cubicle_aisle)` ×2 → one binding

**Micro / dead-code:**
- `anim.rs`: reuse the file's own `elapsed_ms` helper
- `floor.rs`: drop provable no-op `.max(1)`
- `codewhale.rs`: drop dead self-labeled-YAGNI Object arm + its pinning test
- `lighting.rs`: octant `match` → total const array (drops the dead `_ =>` arm; byte-identical render)
- `web/lib.rs`: drop redundant `.max(0.0)` (saturating cast covers it since Rust 1.45)
- `copilot.rs`: drop dead `.or_else(file_stem)` (returned a colliding constant `events` if ever hit; never reached by the directory-walking watcher)

**Kept (fallback IS necessary):** `install/verify.rs` `shell_shim_ref`'s last-token fallback is **test-pinned intentional behavior** (`shell_shim_ref_lone_trailing_quote_is_not_a_quoted_span` documents the WHY) — it recovers a path for `doctor`'s exists/executable stat on hand-edited configs.

**Process:** adds an **"unnecessary fallback / dead default"** factor to `.github/prompts/pr-review.prompt.md` (Family B) — the inverse of negative-space rule 2, with an explicit load-bearing-defense carve-out. That taxonomy is the shared source, so it upgrades both the diff gate and the whole-codebase audit.

## Verification

Local gates, all exit 0: `just clippy` · `just test` (2180 passed) · `just gen-check` (0 pixel drift) · `just hack` · `just lint` · `just wasm-build`. No public API changed (all new fns private/`pub(super)`) → semver-safe.

### DIFF
--- a/.github/prompts/pr-review.prompt.md
+++ b/.github/prompts/pr-review.prompt.md
@@ -94,7 +94,7 @@ given change/tree, say so, don't skip it.
   `Furniture::Desk` footprint `unwrap_or(Size{DESK_W,DESK_H})` both dead (the
   const table is always `Some`) AND WRONG (`decor.rs` is `DESK_W+4`, so a
   hypothetical None mis-placed the sprite), plus the `(7,4)` table footprint
-  re-hardcoded inline in three painters — read the authority (`if let Some(..) =
+  re-hardcoded inline in two more painters — read the authority (`if let Some(..) =
   furniture_def(x).footprint`), never re-copy its fallback. The test is "can the
   trigger fire, and does the arm duplicate/contradict an authority", NOT "does it
   look defensive": load-bearing defense STAYS (documented sharp edges, liveness
--- a/.github/prompts/pr-review.prompt.md
+++ b/.github/prompts/pr-review.prompt.md
@@ -84,7 +84,22 @@ given change/tree, say so, don't skip it.
   dedicated fix-round review; `just arch` had zero CI reach until #273).
 - **(B) Design-debt** — duplication/DRY (N implementations of one concept —
   weight by DIVERGENCE risk, not line count); god-object / oversized module with
-  a clean split; dead code / legacy remnant; leaky or missing abstraction;
+  a clean split; dead code / legacy remnant; **unnecessary fallback / dead
+  default** — a defensive arm (`unwrap_or`/`map_or`/catch-all `_ =>`/`.or_else`/
+  retry) whose trigger CANNOT occur, or that DUPLICATES — worse, CONTRADICTS — a
+  source of truth, is debt not safety: fallback only when a REAL failure mode
+  needs it. The INVERSE of negative-space rule 2 (which forbids demanding
+  defense-in-depth where a primary defense exists) — here, don't ADD or LEAVE a
+  guard for a state that can't arise. The 0.13 slimming audit found a
+  `Furniture::Desk` footprint `unwrap_or(Size{DESK_W,DESK_H})` both dead (the
+  const table is always `Some`) AND WRONG (`decor.rs` is `DESK_W+4`, so a
+  hypothetical None mis-placed the sprite), plus the `(7,4)` table footprint
+  re-hardcoded inline in three painters — read the authority (`if let Some(..) =
+  furniture_def(x).footprint`), never re-copy its fallback. The test is "can the
+  trigger fire, and does the arm duplicate/contradict an authority", NOT "does it
+  look defensive": load-bearing defense STAYS (documented sharp edges, liveness
+  ladders, the shim exit-0 contract, clock-regression `duration_since` folds,
+  config-never-wipe); leaky or missing abstraction;
   correlated-state bundling — N fields that ALWAYS change together (a phase + its
   clock + its profile; a liveness axis + its run-set; a decoder + the extractor it
   pairs with) belong in ONE struct/newtype so an illegal combination is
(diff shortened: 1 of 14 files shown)


IvanWng97:

**Verdict: APPROVE-WITH-NITS → resolved.** Ran the mandatory diff-scope gate as **4 differentiated lenses** on `0600a02e..d1abb795` in an isolated detached worktree. Two escalation triggers from `pr-review.prompt.md` fired, so the base two lenses were augmented:

| Lens | Verdict | Why this lens |
|---|---|---|
| Correctness / grounding | APPROVE (0) | traced all 14 refactors + callers |
| Design / blast-radius | APPROVE-WITH-NITS (1) | raised the NIT below |
| Behavior-preservation **per-call-site** | APPROVE (0) | trigger: diff claims "behavior-preserving" refactor (#461 class — 1-of-N hidden mover) |
| Wrong-abstraction, **adversarial-toward-revert** | APPROVE (0) | trigger: dedup/consolidation ships (#350 class — validate each helper shares ONE reason-to-change) |

All four returned real reviews (14–18 tool calls each, per-site tracing — no stubs).

### Findings → terminal state

- **NIT — prose miscount** (`.github/prompts/pr-review.prompt.md`): the new "unnecessary fallback" factor said the `(7,4)` footprint was re-hardcoded in *three* painters; it was **two** (`furniture.rs` LoungeSideTable + PantryTable) — the desk site used a *different* literal (`Size{DESK_W,DESK_H}`), already called out separately in the sentence. Premise verified by grep (2 hits at furniture.rs:87/129). → **FIXED in `16731d37`**.
- No MEDIUM+ findings. No lens recommended a revert-to-inline.

### Refuted-as-deliberate (coverage proof — each cites its authority)

- **is_fresh** (8 sites): reproduces `duration_since(ts).is_ok_and(|d| d < TTL)` verbatim — strict `<` preserved, per-site TTL consts stay inline, exact-boundary tests still route through the predicates (teeth intact); correctly excludes the `reducer.rs` age-**accumulation** site (different concern).
- **frame_at** (4 sites): `<[Frame]>::first()` is a pure `Option<&Frame>` read, so the pet/mascot `.or` (eager) ≡ `.or_else` (lazy) unification is behavior-identical; `pub(super)` matches the sibling `epoch_ms` convention (no `unreachable_pub`).
- **desk/table let-else**: `furniture_def` is a compiler-exhaustive const match → always `Some` for Desk/Lounge/Pantry, pack-independently (`--pack-dir` loads frames, not geometry); the removed Desk `unwrap_or(Size{DESK_W,DESK_H})` was dead **and** wrong (`decor.rs` is `DESK_W+4`). `let-else`+`return/continue` is the mandated form (`unwrap`/`expect` house-banned).
- **octant** const array: 8 entries transcribe index-for-index onto the deleted arms; `rem_euclid(8) ∈ 0..=7` is total so the removed `_ => (0,0)` was unreachable — `gen-check` 0-drift corroborates.
- **nonempty** (4 env resolvers): character-identical; does **not** reuse `install::io::nonempty_env` because `platform.rs` compiles to wasm under `default-features off` while `install/` is native-gated (importing = layering inversion) — acknowledged in the module comment.
- **copilot_child_key**: sharing across `subagent.started`/`completed|failed` **enforces** the SessionStart↔SessionEnd correlation invariant — the shared derivation is the feature, not a coupling risk.
- **codewhale Object arm** + **copilot `file_stem`**: both genuinely dead in prod (env-mode folds args to a JSON string; the watcher always yields a parented `<dir>/<id>/events.jsonl`); removed tests were the sole exercisers of the dead paths.
- **anim `elapsed_ms`**, **floor `.max(1)`**, **web `.max(0.0)`**, **corridor_zone hoist**: each behavior-preserving for all reachable inputs (divergence only ≥2^64 ms / provable no-op / saturating cast since Rust 1.45 / same immutable value at both sites).
- **Rank-14 KEEP honored**: `install/verify.rs` `shell_shim_ref` last-token fallback + its pinning test untouched (test-pinned intentional `doctor` affordance).

Not merging — leaving for the required checks + the online review bot's latest-comment verdict.

PR #535, opened 2026-07-10
test(reducer): shared apply-DSL for the reducer suite (−360 LOC, coverage-neutral)

IvanWng97:
## What

The reducer integration suite hand-rolled every event as an 8–10 line `r.apply(scene, AgentEvent::X { .. }, at, tp)` block. This adds five terse appliers to `tests/reducer/main.rs` — `act_start` / `act_end` / `waiting` / `proof_of_life` / `sess_end` — and routes ~155 of those blocks through them. A new reducer test is now **one line, not a block**.

**Pure test scaffolding** — no reducer/prod logic touched.

## Scope kept deliberately safe

Only the *uniform* shapes convert. Left inline (the DSL takes `Option<&str>`, so anything it can't express byte-for-byte stays as-is): every `SessionStart` (custom source/session_id/cwd/parent_id), `Identity`, `Rename`, `ToolDetail::`-enum details, and decoder feed-loops. Converting fewer calls correctly beats converting one wrongly.

## Verified teeth-preserving (not trusted)

- `#[test]` count **unchanged** in all 6 files (26/32/37/15/12/8)
- reducer suite **green: 131 passed, 0 failed**
- diff RED-FLAG scan: **zero** `assert` / `.tick(` / `matches!` / comment lines added or removed — the only changed lines are apply-blocks → DSL calls + the `use crate::{…}` imports
- args **spot-checked byte-exact**, incl. the transport-sensitive hook-wins dedup test (`Hook` vs `Jsonl`, `t0 + Nms` timestamps preserved)
- `clippy -D warnings` clean; all five helpers used (no dead test code)

Identical inputs (byte-exact DSL args) + identical assertions ⟹ identical mutant-kill set, so this is coverage-neutral by construction.

## LOC

Net **−360** (test files −451, DSL +91). Modest after rustfmt wraps the 7-arg calls — the plan's −1,423 assumed a richer DSL (incl. SessionStart) and no line-wrapping. The ergonomic win (a block → a call) is the real payoff.

### DIFF
--- a/crates/pixtuoid-core/tests/CLAUDE.md
+++ b/crates/pixtuoid-core/tests/CLAUDE.md
@@ -20,7 +20,7 @@ tests/
 │   │   └── fixtures/hook-payloads.jsonl   codewhale's OWN data (single-owner; NOT scanned)
 │   ├── snapshots/            insta snaps  (sources__conformance__<source>__<scenario>)
 │   └── fixtures/<source>/    ══ conformance scenarios ONLY — dir name MUST be a registered source ══
-├── reducer/main.rs           state-machine behavior (1 binary; shared builders `start`/`delegating_pair` live in main.rs)
+├── reducer/main.rs           state-machine behavior (1 binary; shared scaffolding lives in main.rs — builders `start`/`delegating_pair` + the apply-DSL `act_start`/`act_end`/`waiting`/`proof_of_life`/`sess_end`)
 │   ├── lifecycle.rs          SessionStart/End arms: registration/capacity, resurrect-in-place, hook synthesis of unknown ids, duplicate-start backfill, `Identity`
 │   ├── activity.rs           per-slot FSM: Active/Idle debounce, Waiting set/resolve gates, active_ms + tool_call_count
 │   ├── tasks.rs              active_tasks suppression, hook-wins dedup, drains, b1 cascade grace + waiting-clobber pins
--- a/crates/pixtuoid-core/tests/reducer/activity.rs
+++ b/crates/pixtuoid-core/tests/reducer/activity.rs
@@ -547,7 +547,6 @@ fn codex_permission_then_jsonl_output_resumes_to_active() {
     // Regression: a cx· agent stuck Waiting on a permission prompt must return
     // to Active once the transcript's function_call_output (an ActivityStart)
     // arrives. Hook and JSONL coalesce on the session UUID.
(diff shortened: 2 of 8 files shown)


IvanWng97:
## Two-lens disposition — findings resolved ✅

**Lenses (local, on the diff):** correctness/grounding + design/wrong-abstraction (the #350 dedup/consolidation trigger — a shared test-DSL is exactly the "unify duplication" shape, so the wrong-abstraction lens checked whether the appliers over-collapse distinct event shapes).

- **[FIXED — `6685121e`]** *NIT*: `codex_permission_then_jsonl_output_resumes_to_active` still hand-rolled `AgentEvent::ActivityStart { detail: Some(ToolDetail::from("exec_command")) }` inline, even though `ToolDetail::from(&str)` **is** DSL-expressible (`act_start` maps `Some(&str)` through the same `Into`). Left inline it was an inconsistent boundary; converted, and dropped the now-dead fn-local `use ToolDetail`. (`ToolDetail::`-**enum** details like line 299's `ToolDetail::Task` correctly stay inline — that's the DSL's stated carve-out.)
- **[FIXED — `e11df800`]** *MEDIUM (online bot)*: `tests/CLAUDE.md:23`'s tree comment named only `start`/`delegating_pair`, but this PR adds five shared appliers to `reducer/main.rs`. Docs-currency rule → enumerated all seven at the sibling `watcher/main.rs` line's granularity.
- **[ACCEPTED — residual]** *NIT*: the appliers take `tool_use_id`/`detail` as two adjacent `Option<&str>` positional args, so a caller *could* transpose them silently (a swap only trips tests that assert on the dedup-vs-display axis, not all of them). Accepted because it's test-only scaffolding, **each converted call was verified to preserve the original named-field mapping** (no swap introduced by this change), and transpose-proofing would need per-arg newtypes — ceremony that fights the DSL's whole point of *reducing* boilerplate. A typed `TestHarness` is out of scope for a test-DRY pass.

**Coverage-neutrality proof:** `#[test]` count unchanged per file (26/32/37/15/12/8); byte-exact inputs + unchanged assertions ⟹ identical mutant-kill set (`git diff | grep -E 'assert|\.tick\(|matches!'` on ±lines is empty — only apply-blocks + imports changed). Net −360 LOC (rustfmt re-wraps the 7-arg calls, which is why it's not the −1.4k a naive block→one-liner estimate would suggest).

New head `e11df800` — CI re-running after the two fix commits.

PR #539, opened 2026-07-10
refactor(scene): embed the default pack from one filename list (−37 LOC, byte-identical)

IvanWng97:
## What

`embedded_pack.rs` listed each of the **47** default sprites **twice** — a `let x = include_str!("…/<name>")` binding *and* a matching `("<name>", x)` tuple entry — so adding a sprite meant editing two lists that could silently drift. Replaced with a local `macro_rules!` that expands each filename (listed **once**) to `("<name>", include_str!(concat!("../sprites/default/", "<name>")))`.

**Ride-along:** `motion::walking_position`'s two **integer** `.max(0).min(u16::MAX as i32)` → `.clamp(0, u16::MAX as i32)` (ordered constant bounds → behavior-identical). The *float* `.max().min()` sites elsewhere deliberately stay — their NaN/panic semantics differ from `clamp`.

## Why this and not a crate

From the ecosystem audit: `include_dir` would add a proc-macro to the wasm-lean `pixtuoid-scene` crate and is ~2yr-stale; `rust-embed` breaks the wasm32 default-features=false constraint (runtime `std::fs`). The zero-dep macro is the proportionate fix — kills the double-list footgun, adds no dependency.

## Correctness (byte-identical embedded content)

`concat!("../sprites/default/", "<name>")` folds to the same path literal at compile time, so the embedded bytes are unchanged. Proof:
- Pack goldens that **read the real embedded bytes** pass: `embedded_pack_all_palette_keys_are_distinct_rgbs`, `embedded_pack_recolor_keys_are_distinct_rgbs`, `character_sprite_w_matches_the_embedded_pack`.
- `just gen-check` — **pixel-identical** across every committed baseline (zero render drift).
- `build.rs` still emits the per-file `rerun-if-changed`.

## Verification
- `just preflight` (lint → clippy → hack → test): **2199 passed**, 1 skipped.
- `just gen-check`: OK — every committed artifact pixel-identical.
- Net **−37 LOC**.

### DIFF
--- a/crates/pixtuoid-scene/src/embedded_pack.rs
+++ b/crates/pixtuoid-scene/src/embedded_pack.rs
@@ -133,105 +133,68 @@ pub(crate) fn test_default_pack() -> Pack {
 }
 
 fn load_embedded_pack() -> Result<Pack> {
-    let pack_toml = include_str!("../sprites/default/pack.toml");
-    let seated = include_str!("../sprites/default/seated.sprite");
-    let typing_0 = include_str!("../sprites/default/typing_0.sprite");
-    let typing_1 = include_str!("../sprites/default/typing_1.sprite");
-    let standing = include_str!("../sprites/default/standing.sprite");
-    let walking_0 = include_str!("../sprites/default/walking_0.sprite");
-    let walking_1 = include_str!("../sprites/default/walking_1.sprite");
-    let walking_back_0 = include_str!("../sprites/default/walking_back_0.sprite");
-    let walking_back_1 = include_str!("../sprites/default/walking_back_1.sprite");
-    let walking_coffee_0 = include_str!("../sprites/default/walking_coffee_0.sprite");
-    let walking_coffee_1 = include_str!("../sprites/default/walking_coffee_1.sprite");
-    let desk = include_str!("../sprites/default/desk.sprite");
-    let plant = include_str!("../sprites/default/plant.sprite");
-    let plant_tall = include_str!("../sprites/default/plant_tall.sprite");
-    let plant_fl = include_str!("../sprites/default/plant_flower.sprite");
-    let plant_suc = include_str!("../sprites/default/plant_succulent.sprite");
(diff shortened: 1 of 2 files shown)


IvanWng97:
## Two-lens disposition — APPROVE ✅

Both escalation triggers fired (#461 behavior-preserving refactor → per-call-site lens; #350 dedup/consolidation → wrong-abstraction lens), so both lenses were differentiated accordingly. Each verified against the actual files, not the PR's claims.

**Lens 1 — correctness / byte-identity:** the 47 macro filenames `diff`-identical to the old tuple keys and cross-checked against the 47 `.sprite` files on disk (zero missing, zero dead); `include_str!(concat!(...))` resolves same-file exactly as before (`concat!` const-folds to the identical path literal); the clamp is arithmetically identical (pure i32, constant bounds `0 < u16::MAX`, no NaN path) with no other `.max().min()` wrongly converted; `build.rs`'s directory-walk `rerun-if-changed` is untouched. **0 findings.**

**Lens 2 — design / wrong-abstraction:** RIGHT abstraction — `macro_rules!` is *necessary* (`include_str!` needs a literal, so a `const NAMES + .map()` can't embed), fn-local scope is correct, and the `../sprites/default/` prefix is baked in so the list can't pull from elsewhere. Key catch: **`init_pack.rs` is NOT a double-list** (single inline `(name, include_str!)` tuple for the 2-file skeleton), so leaving it alone is correct YAGNI — and the 47-entry list was the *only* real double-list in the tree. Greppability preserved (filename grep still lands one hit). Clamp ride-along acceptably scoped. **Ship it.**

- **[ACCEPTED — residual, pre-existing]** *NIT*: no dedicated unit test drives `walking_position`'s negative-overshoot→0 clamp boundary. Not introduced by this PR (the clamp's behavior is unchanged); out of scope for a byte-identical refactor.
- Online `review` bot: **Findings: 0**; `security-review`: pass.

**Gates:** `just preflight` 2199 passed · `just gen-check` pixel-identical across every committed baseline · clippy clean · net **−37 LOC**. Ready to merge.

PR #562, opened 2026-07-13
fix(layout): pod grid packs the bottom row, all partials join the lattice (#552, #553)

### LINKED ISSUE #552: layout: pod grid under-fills the bottom row (phantom trailing aisle in main_pod_used_h)
## Problem

The pod grid under-fills vertically: a bottom row of desks that would
physically fit is not emitted, so tall terminals show fewer rows than the room
can hold. Surfaced during #551 (user screenshot: "can we have another row
below?").

## Root cause

In `compute_pod_desks` (`crates/pixtuoid-scene/src/layout/compute.rs`):

```
main_pod_used_h = INTER_POD_AISLE_Y / 2 + couch_to_desk_extra + pod_rows * pod_stride_y
```

`pod_stride_y = pod_h + INTER_POD_AISLE_Y`, so `pod_rows * pod_stride_y` already
includes one trailing `INTER_POD_AISLE_Y` **below** the last pod. That phantom
aisle is subtracted from `residual_h`, leaving it ~`INTER_POD_AISLE_Y` (≈18–20px)
too small — often just under the threshold for the partial bottom row to fire.

## Fix sketch

Subtract the trailing aisle: `main_pod_used_h -= INTER_POD_AISLE_Y` (or stop
counting it in the stride sum). Verified locally this takes 2 → 3 rows at the
reported sizes.

## Why deferred (not in #551)

The fix **ripples into `pet_rest_pos_is_walkable`**: the newly-emitted bottom
desk row lands exactly on the pet's rest cell, so the pet-rest anchor must move
in the same change. That's out of scope for the desk-only PR. Pair this with the
item-2 generative placement-invariant harness (which would have caught the
under-fill as a "container not fully packed" invariant) so the fix is guarded.

Related: #551 (desk PR that deferred this).

### LINKED ISSUE #553: layout: partial desk columns use an off-grid stride → inter-desk spacing jumps on resize
## Problem

When the cubicle band is too narrow to fit a full second pod but wide enough for
a stray desk column, `compute_pod_desks` appends single desk *columns* at the
far-right edge. These partial columns sit at an irregular ~37px offset from the
main pod grid, and the offset jumps as the terminal width changes — so widening
the window (e.g. when a whiteboard gets inserted) makes the inter-desk distance
visibly shift. Surfaced during #551 (user `desk_bug` screen recording:
"why when we extend window width and insert a whiteboard, the distance between
desks changes").

## Root cause

The partial-column path (`'partial_x` loop) places columns at
`right_x + main_pod_used_w + aisle_x/2 + i * partial_col_stride`, where
`partial_col_stride = DESK_W + aisle_x/2`. That stride differs from the full
pod's `pod_stride_x`, so partial columns don't align to the pod rhythm, and
`residual_w` (hence how many partial columns appear and where) changes
discontinuously with width.

## Fix sketch

Options to weigh in the pod-grid PR:
- Distribute the residual width as extra aisle padding across the existing pod
  columns (even spacing, no stray columns), OR
- Snap partial columns onto the pod stride so they read as a continuation, OR
- Drop partial columns entirely below a min-second-pod width (accept the empty
  margin — simplest, matches the "meeting room too narrow → bare floor"
  degradation pattern).

## Why deferred (not in #551)

Pre-existing wart, independent of desk density; needs a placement-stability
invariant (monotonic capacity + bounded per-desk shift across a width sweep) to
verify whichever option we pick. Pair with the item-2 placement-invariant
harness.

Related: #551 (desk PR that deferred this).

IvanWng97:
Roadmap item A: the two deferred #551 layout-correctness bugs.

## What
- **#552**: `main_pod_used_h` counted a phantom trailing aisle, starving `residual_h` — a bottom desk row that physically fit never rendered. Now the whole residual math is GONE: the partial row is simply the first row of the next pod on the lattice.
- **#553** (owner-ratified snap-to-stride): partial right-edge columns continue the pod lattice (column i = the (i % POD_SIDE)-th column of the (pod_cols + i/POD_SIDE)-th pod) instead of the old off-grid `DESK_W + aisle/2` stride — inter-desk spacing can't jump on resize; band-relative offsets are width-monotone.
- **Row rhythm** (owner catch on the first cut): the fixed bottom row initially sat 14px below the last full row vs the 23px inter-pod rhythm — both axes now share `grid.pod_origin` as the ONE lattice source.
- **Ripples**: pet desk visit spot moved to the mid-aisle cell below the desk GROUND (the packed grid closed the old corner spot's snap escape; predicted in #552's filing); corridor-edge scatter plants yield to any desk their ground would overlap (placement-sweep catch; `mask::ground_rect` + `placement::rects_overlap` shared authorities); the no-route-lerp test now calls the real `desk_visit_spot` instead of a second formula copy.

## Tests (RED-first)
`pod_grid_fills_every_desk_row_that_fits` (no fittable row missing, size sweep) · `desk_columns_stay_on_one_lattice_and_never_shift_with_width` (X lattice + width monotonicity) · `desk_rows_stay_on_one_lattice` (the Y twin). Media + wasm regenerated quiesced.

Closes #552. Closes #553.

https://claude.ai/code/session_0122Ni8ZuDMdqS393ZtwFeE3


IvanWng97:
## Two-lens disposition (terminal at 469c75da)

Both lenses APPROVE-WITH-NITS; lens-1 additionally ran a 656k-desk-placement verification sweep (fill-maximality 0 violations, plant/desk ground overlaps 0, take_while == exact fitting count, anchor parity filter==stamp byte-identical).

| Finding | Disposition |
|---|---|
| L2-F1/L1-2: prod `expect()` diverging from the whiteboard-yield's is_some_and ruling | **FIXED** — `overlaps_a_desk_ground` is the one desk-scan authority; both yields read it |
| L2-F2/L1-4: third hand-rolled desk rect scan; dead `0..4` bound | **FIXED** — helper + `0..POD_SIDE` (further columns arithmetically unreachable) |
| L2-F3: visit-spot `+2` = second copy of OBSTACLE_PAD_PX | **FIXED** — derives from the const |
| L2-F5/L1-3: near-vacuous monotonicity disjunct | **FIXED** — unconditional subset assert |
| L1-1: spot comment overclaimed "stays walkable" (0.08% bottom-row spots land on a corridor appliance; snap covers, strictly better than the old 100%-snap-reliant spot) | **FIXED** — comment states the residue honestly |
| L1 NOTE: pod_origin one-past-grid u16 overflow at ~32,700-col terminals | **REFUTED-as-negligible** — same magnitude class as all pre-existing bounded math |

Blast-radius verified by lens 2: capacity/overflow/wasm-hire/hit-test all derive from home_desks; goldens' zero churn is the documented capped-vs-sweep division; pet+mascot spot sharing is the drawable.rs:212 documented design. Escalation lenses (film-critic frame census on regenerated media + render-and-watch on pets.mp4) running; gates re-ran green at this head (2304/2304).

PR #580, opened 2026-07-14
refactor: two G-audit design nits — appliance painters + scan-state struct (#576)

### LINKED ISSUE #576: G audit: two smaller design follow-ups (drawable appliance painters, focus X11 re-derivation, jsonl run_scan_pass params)
**From the roadmap-G whole-codebase audit** — LOW design nits, bundled:

1. **drawable.rs appliance painters inline** (LOW, standards): `paint_drawable` inlines vending/printer/coat-rack painting while `furniture.rs` explicitly exists "to keep the orchestrator focused on the render pipeline rather than individual furniture geometry". Extract `paint_vending`/`paint_printer`/`paint_coat_rack` into furniture.rs matching `paint_kitchen_island`'s shape. (Marginal — deletion test says it relocates rather than concentrates, so low priority.)

2. **focus/linux.rs X11 re-derivation** (LOW, deep-module): a single focus click re-derives the pid→window mapping several times (`focusable(pid)` opens a fresh X connection per call). Cache the mapping for the click's duration. Perf micro-opt.

3. **jsonl/mod.rs run_scan_pass 8 params** (LOW, data-clump): `#[allow(too_many_arguments)]` threading 8 params, 3 of them `&mut` out-params — a Data Clump wanting a `ScanContext` struct. Judge against the compute_waypoints 8-disjoint-facts precedent before acting.

All behavior-preserving; goldens/tests are the oracle. Deferred as they're refactors, not bugs.

IvanWng97:
Closes #576 — the three bundled G-audit design nits. Two implemented; one assessed and declined (below).

## (a) Extract inline appliance painters into furniture.rs

`paint_drawable` painted the vending/printer/coat-rack cells **inline** while the module's convention (and furniture.rs's stated purpose) is that per-furniture geometry lives in furniture.rs (`paint_kitchen_island`, `paint_meeting_chair`, `paint_fish_tank`, …). Moved the three arm bodies **verbatim** into `pub(super) fn paint_vending_machine / paint_printer / paint_coat_rack`; the arms are now thin one-liners mirroring the existing `MeetingChair` arm. **No visibility widening** — the shared `VENDING_PICKUP_SLOT` (already `pub(crate)`) and `epoch_ms` (already `pub(super)`, furniture.rs already calls `super::epoch_ms`) are reused via their existing paths.

## (c) Bundle run_scan_pass's out-params into a ScanState struct

`run_scan_pass` carried 8 params under `#[allow(clippy::too_many_arguments)]`; three (`vouch`, `pid_bindings`, `root_health`) are `&mut` out-params. Verified they're a genuine **Data Clump** (not the disjoint `compute_waypoints` case): they're `let mut` locals created together in the watch-loop setup and threaded through **every** `run_scan_pass` call — cohesive persistent-scan-state. Bundled into `struct ScanState { vouch, pid_bindings, root_health }`; `run_scan_pass` now takes `&mut ScanState` (8 → 6 params, **`#[allow]` removed**). Behavior-identical — same values, same mutation order.

## (b) X11 pid→window caching — ASSESSED, DECLINED

The finding: `ancestor_walk` calls `focusable(pid)` per level, and on X11 each opens a fresh connection + re-fetches `_NET_CLIENT_LIST` + scans `_NET_WM_PID`. The redundancy is real (~50–100 X round-trips vs ~35 for a map built once, at walk depth ~2–3 × ~10–30 windows). **Declined** anyway:
- **Imperceptible gain**: local-socket X round-trips are ~tens of µs; ~100 of them is single-digit ms on a **rare, deliberate, human-initiated** focus click. YAGNI.
- **A memoized pid→window map introduces a NEW silent-failure mode** the current design is immune to: if the first `focusable` scan transiently fails (connect error → empty map cached), every later level returns false → the walk finds no terminal → a **silent focus no-op**. Per-level re-connect naturally retries; a cache poisons on the first failure.
- `focus/linux.rs` is **codecov-ignored, untestable platform glue** — no oracle would catch that regression.

Net: trading untested stateful glue with a new poisoned-cache failure vector for an imperceptible µs win is a bad deal. (A per-instance `OnceCell` under `&self` would keep the trait shape, so this is a value/risk call, not a structural blocker — easy to revisit if the walk ever deepens or focus becomes hot.)

## Gates

`just preflight` green (**2310/2310**, clippy `-D warnings` incl. unreachable_pub); `just gen-check` green — **no media regeneration** (confirms (a) is a byte-identical render move); zero `.snap`/media churn; test-count parity.

https://claude.ai/code/session_0122Ni8ZuDMdqS393ZtwFeE3

### DIFF
--- a/crates/pixtuoid-core/src/source/jsonl/mod.rs
+++ b/crates/pixtuoid-core/src/source/jsonl/mod.rs
@@ -111,6 +111,37 @@ struct WatchCtx<'a> {
     live: &'a Arc<Mutex<HashSet<String>>>,
 }
 
+/// The persistent scan-state bundle threaded through every `run_scan_pass`
+/// (initial seed / 250ms rescan / 60s poll). Created once in `run` and borrowed
+/// `&mut` by each pass; the vouch and pid-binding maps also serve the
+/// instant-exit arm directly (see `run`'s `exit_rx` branch).
+struct ScanState {
+    /// Negative-vouch confirmations: a probe-missing id must stay missing for
+    /// `NEGATIVE_VOUCH_MIN_SPAN` before its exit is emitted (guards a probe
+    /// blip from ending a live session).
+    vouch: NegativeVouch,
+    /// pid → the session ids a healthy probe snapshot bound to it
+    /// (`ProbeSnapshot::pid_of` folded by `refresh_probe_snapshot`) — the
+    /// join the instant-exit arm uses to translate an OS exit into
+    /// SessionEnds. Entries leave on exit (whole pid), negative-vouch
+    /// confirm (single id, see `unbind_session`), or a rebind migration
+    /// (the snapshot observes the id under a different pid).
+    pid_bindings: HashMap<i32, HashSet<String>>,
+    /// Latches the root-scan's persistent-error breadcrumb (warn once per
+    /// failure streak, info on recovery — the `FailureLatch` convention).
+    root_health: FailureLatch,
+}
+
+impl ScanState {
+    fn new(negative_vouch_min_span: Duration) -> Self {
+        Self {
+            vouch: NegativeVouch::new(negative_vouch_min_span),
+            pid_bindings: HashMap::new(),
+            root_health: FailureLatch::default(),
+        }
+    }
+}
+
 pub struct JsonlWatcher {
     root: PathBuf,
     initial_window: Duration,
@@ -230,13 +261,10 @@ impl JsonlWatcher {
     /// healthy. The initial seed + the 250ms rescan + the 60s poll all run this
     /// SAME sequence; only the seed skips the un-claim drain (`drain = false` —
     /// nothing has been pushed at startup). `decoders` is `Copy`.
-    #[allow(clippy::too_many_arguments)]
     async fn run_scan_pass(
         &self,
         ctx: &WatchCtx<'_>,
-        vouch: &mut NegativeVouch,
-        pid_bindings: &mut HashMap<i32, HashSet<String>>,
-        root_health: &mut FailureLatch,
+        scan_state: &mut ScanState,
         exit_watch: Option<&ExitWatch>,
         unclaims: Option<&ChildEndUnclaims>,
         decoders: SourceDecoders,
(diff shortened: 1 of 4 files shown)


IvanWng97:
## Two-lens review — round 1 disposition (fix commit 8c0c97db)

**Lens 1 (correctness): APPROVE, 0 findings** — byte-identity verified line-by-line ((a) only the expected `super::` qualifications + `*busy`→`busy` differ; gen-check green = pixels byte-identical); (c) ScanState threading behavior-identical, no mutation reordered, 8→6 params (`#[allow]` validly removed); no semver surface moved.

**Lens 2 (design): APPROVE** — independently confirmed (a)'s placement matches the fish_tank/meeting_chair convention, and adjudicated (c) a **genuine Data Clump** (vouch+pid_bindings interact via the double-emit guard; root_health is a legit co-traveler; the sibling `notify_health` correctly stays OUT — the tell that the bundling is principled). **Challenged the (b) decline and confirmed it sound** with its own trace.

| Finding | Lens | Sev | State |
|---|---|---|---|
| furniture.rs header + scene CLAUDE.md inventory omit the 3 relocated appliances | L2 | MINOR | **FIXED** (8c0c97db) |
| PR body's (b) rationale partly inaccurate (the "trait-shape ripples" claim is wrong — a `&self` `OnceCell` keeps the trait) | L2 | MINOR | **FIXED** — PR body retightened to lead with the real reasons: imperceptible gain + the poisoned-empty-cache silent-failure mode the per-level-retry is immune to |
| `VENDING_PICKUP_SLOT` has no production consumer in drawable.rs now | L2 | OPTIONAL | **deliberate** — kept as the `DrawableKind::VendingMachine` geometry authority beside the variant (the painter reads it cross-module, same as the pixel test) |
| Co-Authored-By name | L2 | NIT | corrected to `Claude Opus 4.8` in the round commit |

**(b) X11 caching: terminal disposition = ASSESSED, DECLINED** — sound per both my analysis and lens 2's independent trace; the PR body durably records it. `Closes #576` is correct (lens 2 confirms the decline is an accepted terminal state).

Gates: preflight 2310/2310, gen-check green (no media regen), zero `.snap` churn. Awaiting CI + bot on 8c0c97db.

Issue #603, opened 2026-07-14
block_in_place wraps in tui are inert (run_tui runs on the block_on thread), and the #283 sharp-edge rationale is wrong


IvanWng97:
## Problem
Surfaced by the 2026-07-14 whole-codebase audit (adversarial verify of a refuted MEDIUM).

`run_tui` is awaited **inline** inside `run_async` via `rt.block_on(...)` on a `new_multi_thread` runtime (`runtime/driver.rs:41,44`), while `reducer_task` and the sources run on `tokio::spawn`'d worker threads (`driver.rs:109,128`). Therefore the render loop runs on the **block_on thread, not a tokio worker**.

`tokio::task::block_in_place` only migrates a *worker's* queued tasks to another worker so they aren't starved while the worker blocks. On the block_on thread there are no co-scheduled worker tasks to migrate, so **`block_in_place` is inert** — it runs the closure inline with zero scheduling effect. This means:

1. The **5 Sources-panel wraps + 3 onboarding wraps** (the #283→#330 "fix") are no-ops for their stated purpose. reducer_task/sources keep running on their own worker threads whether or not the panel I/O is wrapped.
2. The `tui/CLAUDE.md` rationale — *"yields this tokio worker so the input/render tasks aren't starved … Valid because run_tui runs on the multi-thread runtime"* — rests on a **wrong worker-thread model**. There are no separate input/render *tasks*; they are the loop itself.
3. The **real** UI hazard (sync `flock`/`fsync` in `config::save` and per-hop `/bin/ps`/`swaymsg` subprocess spawns in `focus_slot`, run inline in the render loop) **does** freeze render+input for the contention/probe duration — at all sites equally, wrapped or not. The genuine fix is off-thread offload (`spawn_blocking` + result plumbing), a design change.

## Why deferred (not fixed in the audit arc)
- The proper fix (spawn_blocking + plumbing across the panel/theme/focus I/O sites) is a design change, not a clean win.
- Correcting/removing the wraps contradicts a reviewed concurrency sharp edge (#283/#330); before rewriting it, this deserves an **empirical demonstration** (a test showing removing the wrap changes no observable behavior) so we don't replace one wrong rationale with another.

## Fix sketch
1. Add a small test/experiment confirming `block_in_place` is inert on the block_on thread here.
2. Correct the `tui/CLAUDE.md` block_in_place rationale (and the onboarding twin) to the accurate model.
[...]

PR #605, opened 2026-07-14
harden socket/env path resolution: XDG absolute-guard, omp trim, config-lock O_NOFOLLOW (audit cluster A)

IvanWng97:
Whole-codebase audit **cluster A** — three N-1-of-N env/path sibling gaps (a guard present on the hardened siblings, missing on one member).

### FIND-18 · MEDIUM · live-repro
Both socket-path resolvers (shim `paths.rs` + daemon `claude_code/native.rs`) guard `PIXTUOID_SOCKET` with `!trim().is_empty()` but read `XDG_RUNTIME_DIR` **bare**. `XDG_RUNTIME_DIR=""` → `/pixtuoid.sock` → bind fails fatally:

```
ERROR source died: opening hook socket lock at /pixtuoid.sock.lock: Read-only file system
```

→ hook plane dead the whole run instead of falling to the safe `/tmp/pixtuoid-{uid}/` subdir. Per the XDG spec the value MUST be **absolute**, so the fix filters for non-empty **and absolute** — which also closes the worse *relative*-path case (`foo` → each process resolves against its OWN cwd → shim and daemon silently rendezvous at different paths, no error at all). Applied identically at both resolvers (parity-pinned); the parity test gains an empty/whitespace/relative branch asserting both fall to `/tmp`.

### FIND-01 · LOW
`omp_agent_dir` read `PI_CODING_AGENT_DIR` via `var_os` + byte-empty filter (a `"  "` value resolved the dir to a relative `"  "`) — the lone sibling not using the trim-based `platform::nonempty` its four peers use (the exact pattern `copilot.rs` documents as a *fixed* bug). Routed through `nonempty`; test gains the whitespace case.

### FIND-29 · LOW
`lock_config` opened `<target>.lock` without `O_NOFOLLOW` while its twin hook socket-lock (`hook/unix.rs`) has it. Added for parity (lock content is never read → behavior-identical except symlink refusal). New symlink-refusal test.

Gates: `clippy -D warnings` clean; parity/omp/lock tests green. Part of the 2026-07-14 whole-codebase audit (deferred design items → #603/#604).

### DIFF
--- a/crates/pixtuoid-core/src/source/claude_code/native.rs
+++ b/crates/pixtuoid-core/src/source/claude_code/native.rs
@@ -50,8 +50,12 @@ impl ClaudeCodeSource {
         }
         #[cfg(unix)]
         {
+            // XDG spec: absolute-only. Empty → `/pixtuoid.sock` (fatal bind); relative →
+            // shim/daemon cwd mis-rendezvous → treated as unset. Parity with paths.rs.
             if let Ok(dir) = std::env::var("XDG_RUNTIME_DIR") {
-                return PathBuf::from(format!("{dir}/pixtuoid.sock"));
+                if !dir.trim().is_empty() && std::path::Path::new(&dir).is_absolute() {
+                    return PathBuf::from(format!("{dir}/pixtuoid.sock"));
+                }
             }
             // No XDG_RUNTIME_DIR (macOS, bare Linux): a per-user SUBDIR the bind
             // creates 0700-owned-by-us (`hook::unix::ensure_owned_socket_dir`),
--- a/crates/pixtuoid-core/src/source/claude_code/native.rs
+++ b/crates/pixtuoid-core/src/source/claude_code/native.rs
@@ -164,9 +164,8 @@ mod tests {
             PathBuf::from("/run/user/1000/pixtuoid.sock")
         );
 
-        // Set-but-empty / relative XDG_RUNTIME_DIR is invalid per the XDG spec
-        // (absolute-only) — treated as unset, falls to the /tmp subdir, never
-        // `/pixtuoid.sock` (empty) or a cwd-relative path (parity with paths.rs).
+        // Invalid (empty/whitespace/relative) XDG_RUNTIME_DIR is unset per the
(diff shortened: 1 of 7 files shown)


IvanWng97:
**Review dispositions (2-lens on Sonnet, after the Fable-5 lenses ran out of credits):**

Both lenses **APPROVE-WITH-NITS**; both mutation-tested the new tests (they fail on revert). Folded in / dispositioned:
- **Finding B (per-crate test parity)** — ✅ FIXED (commit): extended `native.rs` + hook `main.rs` socket tests with the empty/whitespace/relative XDG branch (restoring the 3-way coverage the parity test had) + fixed the stale "three branches" comment.
- **Finding A (XDG_CONFIG_HOME/STATE_HOME absolute-guard)** — the two lenses **disagreed on scope** (real footgun vs. socket-cross-process-specific boundary). ISSUE-FILED → #610 for a focused, correctly-scoped decision rather than expanding this socket PR into a 4-site cross-crate change.
- **Redundant `!trim().is_empty()` conjunct** (lens1, LOW) — kept deliberately: it names the "empty → fatal bind" case for the reader and is harmless (is_absolute alone would suffice, but the explicit guard documents intent).
- **Semver** — behavior-only change on a pub surface (relative/empty XDG_RUNTIME_DIR was already broken: fatal bind or cwd mismatch), so no one relied on it; 0.x, no re-bump.

IvanWng97:
**CI unblock (one extra file in the diff):** the `snapshot hygiene` job failed twice on this branch on an **unrelated** test — `fd_probe::pids_by_name_finds_a_spawned_child` — which spawns `sleep 30` then immediately probes the proc-table for it (a spawn→`/proc/<pid>/comm`-visible race). It passed on #607/#608's identical suite runs, and `fd_probe` is untouched by this PR's socket/env logic. Rather than whack-a-mole re-run, folded a bounded ≤500ms settle-poll into the test (the enumeration itself still hard-fails — only the membership check retries). It's the only non-socket file in the diff; the reviewed logic is unchanged.

PR #606, opened 2026-07-14
single-source the character footprint + corridor gates; delete 3 dead Theme roles (audit cluster C)

IvanWng97:
Whole-codebase audit **cluster C** — magic-number/single-source + dead-code (all LOW, behavior-preserving for the bundled pack).

### FIND-04 · single-source + wide-pack correctness
`pixel_painter/sim.rs` re-hardcoded the 8×12 character footprint in the A* occupancy reservation. **First pass** derived it from `CHARACTER_SPRITE_W`/`WALKING_Y_OFF` — but that's identical to the old hardcode, so it single-sourced without fixing the wide-pack case (the online bot's HIGH). **Now** `char_w` (the pack-resolved width `resolve_characters` already computes for the sprite anchor) is resolved ONCE in `sim_step` and threaded into `resolve_characters`, so the occupancy box and every anchor share the one width. `char_w` is pack-constant → the occupancy overlay's cache-signature stays stable. Bundled pack (`char_w == 8`) is byte-identical; a 10-wide `--pack-dir` robot now reserves its real footprint.

### FIND-12 · single-source + misleading-identifier
The two corridor-appliance spawn gates tested the **same** quantity through two identifiers (`cubicle_aisle.width` vs `right_w`) and carried anonymous fit thresholds (10/30/9/40). Both gates now read `cubicle_aisle.{height,width}`; thresholds are named consts.

### FIND-15 · dead-code
Three `Theme` color roles — `SurfaceColors.baseboard`, `OfficeColors.room_wall_body`, `FurnitureColors.chair_seat` — are read by **nothing** (git pickaxe: `room_wall_body`'s last reader went with the #72 frosted-glass conversion, `chair_seat`'s with #556; `baseboard` never had one). Deleted the fields + all 18 per-theme literals.

**semver:** removing 3 pub `Theme` fields is breaking on the `pixtuoid-scene` surface — covered by the already-open **0.15.0** minor over published 0.14.0.

Gates: workspace `clippy -D warnings` clean; **558** scene tests green (insta goldens byte-identical). Part of the 2026-07-14 whole-codebase audit.

### DIFF
--- a/crates/pixtuoid-scene/src/layout/compute.rs
+++ b/crates/pixtuoid-scene/src/layout/compute.rs
@@ -1425,8 +1425,13 @@ pub(super) fn compute_waypoints(
 
     // Corridor appliances — stored as centre points (same convention
     // as Pantry/Couch). Painter derives top-left via sub(w/2, h/2).
-    // Sizes: vending 4×6, printer 5×4.
-    if cubicle_aisle.height >= 10 && cubicle_aisle.width > 30 {
+    // Sizes: vending 4×6, printer 5×4. Both gates read `cubicle_aisle.{h,w}`
+    // (== right_{h,w}) so the two siblings can't drift to different identifiers.
+    const VENDING_MIN_AISLE_H: u16 = 10;
+    const VENDING_MIN_AISLE_W: u16 = 30;
+    const PRINTER_MIN_AISLE_H: u16 = 9;
+    const PRINTER_MIN_AISLE_W: u16 = 40;
+    if cubicle_aisle.height >= VENDING_MIN_AISLE_H && cubicle_aisle.width > VENDING_MIN_AISLE_W {
         waypoints.push(Waypoint {
             pos: Point {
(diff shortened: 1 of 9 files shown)


IvanWng97:
**Disposition — online bot MEDIUM (wide-pack occupancy test gap):** ISSUE-FILED → #609.

The char_w code fix is done and the bot verified it correct (occupancy box + every anchor now share the one pack-resolved `char_w`, resolved once in `sim_step`; bundled pack byte-identical). The remaining gap is a *test* for the wide-pack path, which needs a **real ≥10-wide `Pack` fixture** — the anchor tests only fake `w=10` by passing it directly, there is no 10-wide `Pack` to drive `sim_step` end-to-end. Building that reusable fixture + an `AtWaypoint` occupancy test is genuine test-infrastructure, tracked in #609 (the under-reservation it guards is cosmetic and `--pack-dir`-only).

Issue #612, opened 2026-07-14
test-teeth: reducer behavior-pinning tests for the active_tasks orphan (FIND-03) + parentless-revival resurrect guard (FIND-35)


IvanWng97:
> **Corrected 2026-07-14** after #613's lens-2 empirically disproved the original plan (it ran the reducer). The FIND-03 orphan is NOT load-bearing and FIND-35's sketch had a B-shadows-C flaw — both rewritten below. See #613 for the corrected sharp edge/comments.

Deferred test-teeth from the whole-codebase audit. Delicate reducer/decoder setups that warrant a focused pass.

## FIND-36 — all-source decoder content-cap test (the main item)
The cap chokepoints (`ellipsize`/`generic_tool_display`/`describe_tool_target`) in `decoder.rs` are routed by all 11 sources, but only CC has a test asserting the cap FIRES. The documented residual class is a per-decoder chokepoint BYPASS (a label minted from raw content without `ellipsize` — #272 missed 3 hook-only decoders, copilot twice). Add a table-driven `every_agent_decoder_caps_its_tool_display`: drive each registered source's real decoder with an over-cap field, assert the emitted display length ≤ `MAX_DECODED_FIELD_CHARS + 1 + ": ".len() + MAX_TOOL_TARGET_CHARS + 1`, plus a completeness assert against `REGISTERED_SOURCES` minus daemons. Each source has a bespoke input shape — that's the work.

## FIND-03 — pin the ACTUAL orphan behavior (1 test, re-scoped)
The `active_tasks` insert in `track_active_tasks` is deliberately un-gated. It is **harmless + self-correcting**, NOT load-bearing (the original plan's "test 2 proves the guard would regress" was **wrong** — gated vs ungated is byte-identical because `tick`'s `active_tasks.retain` reaps the orphan every ~1s, ~4 reaps before a desk can free at the 4.5s EXIT_GRACE_WINDOW). So:
- **Test 1 (keep, re-scoped):** under desk exhaustion a hook Task dispatch for an unknown id leaves an orphan; the next `tick` reaps it (no ghost slot, session never registers). Build the Task `ActivityStart` inline (`detail: Some(ToolDetail::Task ...)` — the DSL's `act_start` can't express an enum detail).
- **Test 2: DROP it.** There is no regression to pin — the orphan doesn't preserve suppression in the realistic path.

## FIND-35 — parentless-revival resurrect guard (1 test, corrected)
[...]

PR #613, opened 2026-07-14
core correctness + refutation records: daemon recycled-pid HashSet mirror, active_tasks sharp edge, pet-hitbox test (audit cluster B2)

IvanWng97:
Whole-codebase audit **cluster B2**. Every finding here was run through a **design + adversarial-verify workflow** before implementation — which REFUTED the FIND-03 code change (a regression) and flagged the FIND-20 `return`-vs-`break` bug in the first draft.

### FIND-20 · LOW · latent N-daemon race — FIXED
`PresenceExitWatch` keyed its `pid → source` binding by pid alone with last-writer-wins, so a transient A→B pid recycle flipped a still-live daemon's binding (misrouting A's death, losing B's). Inert today (one daemon: `openclaw`) but the reducer/sweep drivers are registry-driven for N. Mirror `HookPidWatch` exactly: `pid → HashSet<source>`, extracted `note_source`/`take_sources` (unit-testable without the platform watcher), drain does take-on-death emitting one `PidExited` per bound source and **`return`s** (not a bare `break`) when the receiver is gone. Single-daemon path is byte-identical (set dedups == overwrite). Two unit tests incl. the recycle regression pin.

### FIND-03 · LOW · REFUTED-to-sharp-edge
The design+verify pass **refuted** adding a `contains_key` guard: the `active_tasks` insert in `track_active_tasks` is *deliberately* not slot-gated (unlike the hook-wins dedup record). The slot-less orphan eats nothing and `tick` reaps it, but it must persist across a desk-exhaustion window to keep subagent-leak suppression alive for a delegating parent that materializes via its child's misattributed hook — gating it reintroduces the parent-shows-child's-tool leak. Recorded as a `pixtuoid-core/CLAUDE.md` **sharp edge** + a 2-line WHY at the insert. Behavior-pinning tests → #612.

### FIND-35 · LOW · comment accuracy — FIXED
The resurrect-in-place 3-conjunct comment overstated which test pins the middle conjunct; corrected to name the exiting + incoming-parent pins and mark the slot-parent conjunct belt-and-braces (no single test isolates it → #612).

### FIND-17 · LOW · test-teeth — FIXED
Pet click-hitboxes were hardcoded with nothing pinning them to the embedded pet sprite dims (unlike the character/desk pins). Added `pet_hitboxes_track_the_embedded_pack`.

**Deferred (test-teeth, → #612):** FIND-36 (all-source decoder cap test) + the FIND-03/FIND-35 reducer behavior-pinning tests — delicate setups that deserve focused work over tail-of-session.

Gates: `clippy -D warnings` clean; daemon + pet tests green. Part of the 2026-07-14 whole-codebase audit.


IvanWng97:
**Disposition — lens 2 REQUEST-CHANGES (both MUST-FIX addressed, commit be5b95b9):**

Lens 2 empirically disproved the FIND-03/FIND-35 documentation (it *ran* the reducer — exactly the teeth the gate is for). I verified the mechanism myself (`tick`'s `active_tasks.retain` reaps every ~1s at SWEEP_TICK_INTERVAL_SECS=1; a desk only frees at the 4.5s EXIT_GRACE_WINDOW → the orphan is gone ~4 reaps earlier) and corrected:
- **FIND-03** — the sharp edge + inline comment now say the orphan is HARMLESS + reaped by tick (not "load-bearing / the only carrier"); gated-vs-ungated is byte-identical, so leave it ungated as cosmetic self-correcting debt.
- **FIND-35** — now states NEITHER parent conjunct is isolated by an existing test (the subagent test blocks on the slot's own `parent_id.is_some()` regardless of the incoming one).
- **#612** — deferred-test sketches corrected (test 2 proves no regression; the parentless sketch has the same B-shadows-C flaw).

FIND-20 + FIND-17 were verified clean by both lenses — unchanged. Re-review requested on the corrected head.

PR #614, opened 2026-07-14
focus-click follows the live sprite (MEDIUM #22) + desk hover box from the table (#23) (audit cluster E)

IvanWng97:
Whole-codebase audit **cluster E**. Both fixes were run through a **design + adversarial-verify workflow** before implementation (both PROCEED — the #22 motion/route_ctx/cached-clock safety was confirmed, the #23 desk anchor traced exactly).

### FIND-22 · MEDIUM · confirmed live
Click-to-focus (the 0.14.0 headline) silently missed any agent **not at its home desk**. The CLICK path used home-desk-only `hit_test_from_tui`; HOVER used the router-aware `hit_test_agent` (anchors on `character_anchor`, follows the walking sprite). So a wandering / pantry / entry-walk agent (idle agents are away ~25–35% of the time + every ~4s entry walk) was hoverable but **not clickable**, and clicking its vacated desk focused the *absent* agent.

**Fix:** `TuiRenderer::hit_test_agent_at` — the click twin of the hover hit-test — reading a cached `last_render_now` (new field, set in `render()`, same discipline as `last_scale`) so the click resolves against the **drawn** frame's clock. `character_anchor` is idempotent per `now`, so re-deriving at click time is a state no-op (the event loop is single-threaded with `render()`). `focus_clicked_agent` switches to it; `hit_test_from_tui` is **retained** (the render harness's seated-agent locator + its unit tests) with its stale "click-to-pin" doc corrected. New harness test proves the gap end-to-end: an exiting agent mid-walk is found by `hit_test_agent_at` at its live cell where `hit_test_from_tui` misses it.

### FIND-23 · LOW · single-source
`hit_test_furniture`'s desk arm hardcoded `DESK_W+2 × DESK_H`, contradicting its own "derive from the furniture table" header — 2px narrower than the painted desk (east overhang un-hoverable) + drift-prone. Now derived from `furniture_def(Desk).visual` (`DESK_W+4 × DESK_H+2`, top-left anchored); drops the now-unused `DESK_W`/`DESK_H` import. `furniture_hit_test_finds_desk` gains an east-overhang assertion (RED on the old box, table-derived so it can't re-drift).

Gates: `clippy -D warnings` clean; full `pixtuoid` suite green. Part of the 2026-07-14 whole-codebase audit.

### DIFF
--- a/crates/pixtuoid/src/tui/hit_test.rs
+++ b/crates/pixtuoid/src/tui/hit_test.rs
@@ -61,7 +61,13 @@ pub(crate) fn hit_test_agent(
 /// multi-floor `desk_index` was exactly the global/local confusion the
 /// `GlobalDeskIndex` newtype exists to prevent: while viewing floor ≥ 1 it
 /// could pin an invisible agent from another floor.)
-pub fn hit_test_from_tui(scene: &SceneState, layout: &Layout, mx: u16, my: u16) -> Option<AgentId> {
+#[cfg(test)]
+pub(crate) fn hit_test_from_tui(
+    scene: &SceneState,
+    layout: &Layout,
+    mx: u16,
+    my: u16,
+) -> Option<AgentId> {
     const SPRITE_W: u16 = pixtuoid_scene::layout::CHARACTER_SPRITE_W;
     const SPRITE_H_CELLS: u16 = pixtuoid_scene::layout::CHARACTER_SPRITE_H_CELLS;
     for agent in scene.agents.values() {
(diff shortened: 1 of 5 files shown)


IvanWng97:
**Dispositions — lens 2 APPROVE-WITH-NITS (both taken, commit pushed):**
- **FIND-A** (redundant cache) — FIXED: threaded `now` through `focus_clicked_agent` → `hit_test_agent_at`; deleted the `last_render_now` field + Option + the unreachable-skew caveat. You're right the `last_scale` analogy didn't hold (it caches a *computed* value; this cached a *received* arg).
- **FIND-B** (least-privilege) — FIXED: demoted `hit_test_from_tui` to `#[cfg(test)] pub(crate)` (its callers are all cfg(test) now — the harness is `#[cfg(test)]`) + dropped it from the renderer.rs `pub` re-export.
- Lens 1's Arc-clone-comment nit + FIND-23 verified clean. Re-review welcome on the new head.

PR #670, opened 2026-07-18
fix(site): retry the shared wasm init so a transient fetch drop self-heals

IvanWng97:
## Problem

`pixtuoid.dev` occasionally loads with the office **stuck on the still poster** — the live wasm office never comes up — and a reload fixes it. Diagnosed as a **transient wasm-fetch failure that couldn't self-heal**:

- The hero backdrop and the Showcase VIBING channel share **one** `window.__pixWasm` init promise (needed — the glue's `__wbg_init` memoizes only the *resolved* instance, not the in-flight promise, so independent inits would double-instantiate and stomp the module-global).
- That promise was memoized **even when it rejected**, and the boot `.catch` **never retried** — so a single dropped ~874KB fetch (flaky link, or a brief glue/wasm cache-desync during a Pages deploy: `max-age=600`, unhashed filenames) stranded the office on the poster until a manual reload.

The live deploy is healthy on every systemic axis (CSP `wasm-unsafe-eval` present, `application/wasm` MIME, gzip, live sha256 == committed manifest), so this is a resilience gap, not a broken build.

## Fix

Retry the import+instantiate **2× with 400ms backoff, inside the one shared `__pixWasm` promise** (a per-consumer retry would reintroduce the double-instantiate race). Only once the retries exhaust does it reject → poster, the deliberate no-JS/no-wasm/reduced-motion fallback, unchanged. Duplicated across the two `is:inline` consumers (`OfficeBackdrop.boot` / `Showcase.bootCanvas`) — separate script scopes, no shared const, pinned by a cross-ref comment.

## Test

New Playwright case: abort the **first** `pixtuoid_web_bg.wasm` fetch, let retries through, assert the office still goes `● LIVE` (`abortedFirst && wasmHits > 1`). **Red-checked** — it fails without the fix (office never attaches `.backdrop.is-live`).

## Gates

- `just site-check` — tsc + 38 unit + eslint + build ✓
- `just site-e2e` — 77 Playwright tests ✓ (incl. goes-live, failure-poster, VIBING second-consumer, new retry)
- `just preflight` — full Rust gate green (pre-push hook, 2527 tests) ✓

Two-lens review in progress; disposition will follow in-thread before this is merge-ready.

### DIFF
--- a/site/CLAUDE.md
+++ b/site/CLAUDE.md
@@ -269,7 +269,11 @@ display-line authority (`starText`), unit-tested on its null-stars arm since
   call via `window.__pixWasm` — the generated glue's `__wbg_init` guards only the
   already-resolved instance, not an in-flight promise, so independent `mod.default()`
   calls racing before either resolves instantiate two separate wasm instances that
-  stomp the single module-global `wasm`. Schema: `kind:"live"`
+  stomp the single module-global `wasm`. A bounded retry (2×, 400ms backoff) wraps
+  that shared init for transient wasm-fetch resilience, but stays INSIDE the one
+  `__pixWasm` promise — a per-consumer retry would reintroduce the very
+  double-instantiate race above (the retry consts are pinned equal across the two
+  boot scripts by `config/wasm-init-consts.test.mjs`). Schema: `kind:"live"`
   + `variantGroups` (per-group `retint`) + `poster` + `timeSlider` in `showcase.json`,
   resolved by `showcaseGroups` in `consts.ts`, validated by the `astro.config.mjs`
   showcase guard's live branch. Fallback (no-JS / no-wasm / reduced-motion): the
--- a/site/config/wasm-init-consts.test.mjs
+++ b/site/config/wasm-init-consts.test.mjs
@@ -1,0 +1,32 @@
+// The bounded wasm-init retry lives in TWO is:inline scripts that share no JS
+// module scope — OfficeBackdrop's boot() and Showcase's bootCanvas() — so its
+// consts are duplicated by necessity (an is:inline script can't `import` a
+// shared const at runtime). Only ONE copy ever executes per page (the
+// `window.__pixWasm ||` short-circuit picks whichever consumer boots first), so
+// a drift is runtime-inert — but a one-sided retune would silently make the
+// retry budget depend on which consumer booted first. Pin the two copies equal
+// so the "keep in sync" comment is a fact, not a hope (the repo magic-number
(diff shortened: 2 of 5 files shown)


IvanWng97:
## Two-lens merge review — disposition sweep

Diff-scope gate run with two differentiated lenses (correctness/grounding + design/blast-radius) on the worktree. **Both APPROVE-WITH-NITS, zero MEDIUM+, zero blockers.** Every finding driven to a terminal state:

| # | Lens | Finding | Disposition |
|---|------|---------|-------------|
| L1-1..5 | correctness | single-instance invariant preserved · retry bounded/terminating (3 attempts, awaited 800ms) · splash can't hang (independent `MAX_ENGINE_WAITS` ≈4s backstop) · sibling sweep: no 3rd `__pixWasm` consumer · test has real teeth | **CONFIRMED-OK** (no change) |
| L1-6 | correctness | comment overstated recovery — the retry re-fetches the **binary**; a first-load glue-`import()` failure is import-map-cached and not covered | **FIXED** — both boot comments scoped to "wasm-binary fetch" |
| F2 | design | stale sibling comment: pre-existing "~740KB" vs new "~874KB" for the same artifact | **FIXED** — unified to ~874KB (`OfficeBackdrop.astro:1109`) |
| F1 / L1-7 | both | retry consts duplicated across the two `is:inline` scripts, pinned by comment not test | **FIXED** — `config/wasm-init-consts.test.mjs` pins `WASM_INIT_RETRIES`/`BACKOFF_MS` equal across both files (repo magic-number convention prefers a test-pin) |
| F3 | design | doc-currency: `site/CLAUDE.md` `__pixWasm` contract entry didn't mention the retry | **FIXED** — VIBING entry now records the "retry stays inside the one shared promise" invariant |
| F4 | design | cross-trigger recovery: `__pixWasm` keeps the rejected promise on final exhaustion, so a later-booting consumer can't retry after a recovered outage | **ISSUE-FILED** → #671 (narrow precondition; nulling on final rejection verified race-safe; wants its own e2e) |

Both lenses cleared the two things that most needed checking: the **success path adds zero latency** (retry only fires on `.catch`), and the **failure-path ~800ms is absorbed by the splash's existing 4s cap** (no new hang).

Review round folded in **13581d0c**. Gates re-run green: `just site-check` exit 0 (40 unit incl. the new bridge test); e2e goes-live + failure-poster + retry all pass (CSP hashes re-derived for the edited inline scripts). `just preflight` green on push.

Not merge-ready until the online review bot posts `Findings: 0` at HEAD and CI is green — a human merges.

PR #680, opened 2026-07-19
fix(ci): deploy renders /architecture — pin withastro/action to npm + guard

IvanWng97:
## Bug: `/architecture` deploys as an empty page

`https://www.pixtuoid.dev/architecture/` ships an **empty `<article>`** — no headings, prose, or the mermaid diagram — while every other doc route renders fully. (Verified by downloading the deployed Pages artifact: `dist/architecture/index.html`'s article is 4 bytes.) A local build of the exact deployed commit renders it **fully** (48 KB, all headings, the SVG), so source + deps are fine — the **deploy build** is where it breaks.

## Root cause

`ARCHITECTURE.md` is the only doc with a ```mermaid block, rendered to inline SVG at build by `rehype-mermaid` (needs Playwright Chromium). The deploy build log shows `withastro/action@v6`'s package-manager detection **fell back to pnpm** (the repo has no `pnpm-lock.yaml` to match). pnpm's fresh install resolved a *different* Playwright than the Chromium the prior step installed for npm → the classic `Executable doesn't exist at …/ms-playwright/chromium-XXXX` → `rehype-mermaid` failed and collapsed the whole `<Content />` to an empty article **without erroring the build**, so the deploy stayed green. CI (`site.yml`) never caught it because it builds directly on the host (its build renders fine).

## Fix (keep the official Astro action)

Per `withastro/action`'s own `action.yml`, it has a `package-manager` input. Pass **`package-manager: npm`** so it reuses `package-lock.json` → the build's Playwright matches the pre-installed Chromium → the diagram renders. This is the community-standard pattern for rehype-mermaid on Pages (install Chromium → `withastro/action`), which the repo already followed — the only bug was the pnpm drift.

Plus a **fail-loud guard**: assert `dist/architecture/index.html` rendered its body after the build (deploy is `needs: build`), so a broken diagram can never silently ship an empty page again.

## Verification
- ✅ `actionlint` clean; `package-manager` input confirmed from the action's `action.yml`; no stray `pnpm-lock.yaml`.
- ✅ Local build of HEAD renders `/architecture` fully; the guard grep passes on a good build, fails on empty.
- ⚠️ The failure is deploy-env-specific and can't be fully reproduced locally — but this uses the action's sanctioned input, and the guard guarantees the worst case is a **red deploy**, never another silently-empty page.

### DIFF
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -43,22 +43,38 @@ jobs:
         with:
           path: ~/.cache/ms-playwright
           key: playwright-${{ runner.os }}-${{ hashFiles('site/package-lock.json') }}
-      # withastro/action only runs `astro build`; gate the deploy on lint + types
-      # too, so a build-clean but lint/typecheck-failing commit can't ship. Also
-      # install Chromium here (same runner → cached) so withastro/action's build
-      # can render the /architecture ```mermaid block to an inline SVG.
+      # withastro/action runs `astro build` + uploads the Pages artifact; gate the
+      # deploy on lint + types too (a build-clean but typecheck-failing commit
+      # mustn't ship), and install Chromium here (same runner → global cache) so
+      # the action's build can render /architecture's ```mermaid block to an SVG.
       - name: Validate (lint + types) + Chromium for the diagram
         working-directory: site
         run: npm ci && npm run lint && npm run check && npx playwright install --with-deps chromium
       - uses: withastro/action@v6
-        # the ticker's build-time merged-PR fetch: unauthenticated quota on
-        # shared runner IPs is routinely exhausted -> silent canned-reel
-        # fallback; the ambient token (contents: read) gives real headroom
+        # the ticker's build-time merged-PR fetch: unauthenticated quota on shared
+        # runner IPs is routinely exhausted -> silent canned-reel fallback; the
+        # ambient token (contents: read) gives real headroom
         env:
           GITHUB_TOKEN: ${{ github.token }}
         with:
           path: ./site
           node-version: 26
+          # PIN npm. Without it, v6's detection fell back to pnpm, whose fresh
+          # install resolved a DIFFERENT Playwright than the Chromium installed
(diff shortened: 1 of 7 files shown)


IvanWng97:
## Review disposition — HEAD `6c78cd92`

Two differentiated lenses + the online bot. Every finding driven to a terminal state:

| Source | Finding | Disposition |
|---|---|---|
| Lens 1 (correctness) | old inline `grep 'Data flow'` defeated by the on-page TOC (green on a collapsed article) | **FIXED** — generic body-exclusive `check:docs`; proven: old grep passes an emptied `/architecture`, check:docs fails it |
| **Bot HIGH** | `check:docs` reachable only via `npm run verify`; `site.yml` (PR-gating CI) runs steps individually → guard didn't gate PRs | **FIXED** — added `npm run check:docs` to `site.yml` after build; now runs in verify / CI / deploy build-cmd |
| **Bot MED** | stale `verify` chain in `site/README.md` | **FIXED** |
| Lens 2 F3 | article regex required `prose` as sole class | **FIXED** — `\bprose\b` |
| Lens 2 F4 | `site/CLAUDE.md` mermaid note incomplete | **FIXED** — added the version-mismatch silent-empty sharp edge + the fix |
| Lens 2 F6 | body-size comment | **FIXED** |
| Lens 2 F5 | tighten svg → `aria-roledescription="flowchart"` | **REFUTED** — couples to diagram type; generic `<svg>` + size floor is more robust |
| Lens 1/2 F1/F2 | `package-manager: npm` + guard design | **CONFIRMED-OK** |

Both bots' caveat (no-network sandbox → couldn't verify `withastro/action`'s `package-manager`/`build-cmd` inputs) is covered: I fetched the action's `action.yml` — both inputs exist and are honored.

Gates: `check:docs` passes (5 doc pages + `/architecture` svg); prettier/eslint/actionlint clean; `just site-check` green; `preflight` green on push. Awaiting the fresh bot pass at this HEAD (`Findings: 0`) before merge — a human merges.

PR #679, opened 2026-07-19
refactor: architecture-deepening — 5 byte-identical deep-module wins from the design review

IvanWng97:
## What

Five deep-module refactors surfaced by an architecture-deepening review
(`/improve-codebase-architecture` → 10-lane survey → adversarial deletion-test
verify → grill each design one decision at a time). Each concentrates a
duplicated or leaked concern behind a small interface. **All five are
behavior-preserving** — byte-identical render / geometry / timing / decoding.

| # | Commit | Concentrates |
|---|--------|--------------|
| #1 | `badge_hue` | the CLI-identity `·`-split + `by_prefix` decomposition, re-derived in all 3 painters → one resolver beside `label_tone_rgb`; painters keep only their Span / blit / JSON encode |
| #9 | `generic_keyed_detail` | every non-CC source's `first_present_str → generic_tool_display` last mile → one chokepoint, so a copy-pasted 9th source **structurally can't bypass** `MAX_TOOL_TARGET_CHARS` |
| #6 | `FurnitureDef::ground_rect` | the `visual`/`ground_x`/`ground_y` field-threading duplicated at ~15 mask / compute / sweep sites → one accessor (+ `ground_rect_of` for the waypoint's runtime footprint) |
| #4 | `snapshot_leg_profile` | the `route → measure → freeze` pipeline hand-assembled at 5 one-shot-leg sites (entry/exit/snap-back + wander out/back) → one composer |
| #5 | `sofa_east_drain_edge` | `place_wall_decor`'s bounds-reconstruction of the meeting sofa → a room accessor reading the **real placed** sofa (extends #668's `coat_rack_pos` pattern; closes the `rooms/mod.rs` "polish arc's next step" debt) |

Commits are in disjoint files and can be reviewed / reverted independently.

## Why they're safe

Each was verified byte-identical against the guard that already pins its area:
- **#1** — tui/floating label tests + the web `overlay_json` badge-hue test (JSON output unchanged; wasm regenerated).
- **#9** — the `every_agent_decoder_caps_its_tool_display` cap guard.
- **#6 / #5** — the generative `placement_sweep` connectivity/sealed-pocket harness (sizes × seeds).
- **#4** — the 110 pose/motion regressions.

New direct unit tests added for #9 / #6 / #4 / #5 (the interface each refactor
created is now the test surface).

## Test

`just preflight` green — **2535 tests run, 0 failed, 2 skipped**. `just gen-wasm`
[...]

PR #682, opened 2026-07-19
fix(ci): deploy re-renders /architecture — disable the poisoned Astro build cache

IvanWng97:
## Problem

The #680 merge's deploy ([run 29671701103](https://github.com/IvanWng97/pixtuoid/actions/runs/29671701103)) **failed at the `check:docs` guard** — `/architecture` still rendered empty. The guard working as designed (it reddened instead of shipping empty), but the underlying render was still broken.

## Root cause (from the deploy log — NOT a Chromium problem)

`withastro/action@v6`'s default **`cache: true`** caches the Astro build cache (`node_modules/.astro`) under a bare `astro-cache-<os>-` restore-key. So each deploy **restores the *previous* deploy's cache**:

```
Cache restored from key: astro-cache-Linux-d346569f…   # d346569f = the prior #679 deploy
…
/architecture/index.html (+7ms)                        # +7ms = a cache HIT, not a real render
```

The #679 (pnpm-era) deploy had cached an **empty** `/architecture` (the Chromium miss #680 diagnosed). Every deploy since served that empty `<Content />` back as a **+7ms cache hit** — never launching Chromium, never re-rendering the ` ```mermaid ` block, no error. So #680's `package-manager: npm` render fix was correct but **never actually ran**.

`site.yml` (the PR gate) renders fine precisely because it has **no cross-run Astro cache** — always a fresh build.

## Fix

`cache: false` on the action → a fresh, real mermaid render every deploy; the poison can't persist. `check:docs` (from #680) still proves the render before upload, so a genuine render failure would redden rather than ship empty.

- `.github/workflows/pages.yml` — add `cache: false`
- `site/CLAUDE.md` — document the cache-poisoning failure mode alongside the existing Chromium-mismatch note.

## Verification

The real proof is the deploy itself (this failure class only reproduces in the action's cross-run cache env; local + `site.yml` always render fresh). Once merged, the `main` deploy does a cold build → real render → `check:docs` passes → `/architecture` goes live.

### DIFF
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -59,6 +59,16 @@ jobs:
         with:
           path: ./site
           node-version: 26
+          # DISABLE the action's Astro build cache (node_modules/.astro). It is
+          # keyed with a bare `astro-cache-<os>-` restore-key prefix, so a deploy
+          # RESTORES the most recent prior deploy's cache — and a deploy that once
+          # rendered /architecture EMPTY (the pnpm-era Chromium miss below) cached
+          # that empty <Content /> and every later deploy served it back as a
+          # +7ms cache HIT, never re-rendering the ```mermaid block. `cache: false`
+          # forces a fresh, real mermaid render each deploy — poison can't persist.
+          # (site.yml renders fine precisely because it has no cross-run Astro cache.)
+          # A static 7-page build; the lost speedup is milliseconds, correctness isn't.
+          cache: false
           # PIN npm. Without it, v6's detection fell back to pnpm, whose fresh
           # install resolved a DIFFERENT Playwright than the Chromium installed
           # above → rehype-mermaid's inline-SVG render of /architecture failed
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -67,7 +67,9 @@ jobs:
           # +7ms cache HIT, never re-rendering the ```mermaid block. `cache: false`
           # forces a fresh, real mermaid render each deploy — poison can't persist.
           # (site.yml renders fine precisely because it has no cross-run Astro cache.)
-          # A static 7-page build; the lost speedup is milliseconds, correctness isn't.
+          # Cost is a cold content-layer rebuild (incl. the one Chromium mermaid
+          # render) — seconds on a 7-page site, negligible for an infrequent deploy;
+          # correctness isn't negotiable.
           cache: false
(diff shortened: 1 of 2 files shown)


IvanWng97:
## Two-lens merge review — both lenses APPROVE

Ran two differentiated review lenses (correctness/grounding + design/blast-radius), each armed with the failed-deploy log to adversarially challenge the root-cause claim.

**Lens 1 (correctness):** APPROVE. Verified against `withastro/action@v6`'s `action.yml` that `cache: false` gates **both** the restore *and* save steps (not a silent no-op), the restore-key is the bare `astro-cache-<os>-` prefix (so cross-deploy restore is real), and — critically — the fresh render will now succeed: `playwright` is lockfile-pinned to `1.61.1`, Validate installs the matching Chromium, and the action's `npm install` on the already-`ci`'d tree is a version no-op. **Empirical proof on this very PR:** `site.yml`'s `paths:` includes `pages.yml`, so this PR triggers it, and `site.yml` runs the *identical* cacheless render path — a green `site` check here IS proof the fresh `/architecture` render works. Worst case of `cache: false` is a red deploy (check:docs backstop), never a silently-shipped empty page.

**Lens 2 (design/blast-radius):** APPROVE. `cache: false` is the only lever the action exposes (no cache-key input), so it's the minimal correct fix under the "keep withastro/action" constraint. Blast radius negligible (cold rebuild of a 7-page static site = seconds, well inside the 20-min timeout) and it *removes* risk — the deploy path now matches the proven-good `site.yml` cold-build path. The remaining Playwright-Chromium cache is exact-match keyed on the lockfile (no bare-prefix fallback), so it can't repoison.

### Disposition
- **N1 (Lens 2, low) — FIXED** (`0bd3efc8`): the comment said the lost speedup is "milliseconds" (the +7ms mermaid HIT), but `cache: false` disables the whole content-layer cache → cost is a few *seconds* of cold rebuild. Reworded.
- **`#682` self-reference (Lens 1, low) — VERIFIED**: this PR is #682, so the `site/CLAUDE.md` reference is correct.
- **N2/N3 (Lens 2, very low) — no action**: the poisoned cache entry self-evicts (7-day LRU) and reverting `cache: false` is a deliberate act guarded by the documented sharp edge; the comment/CLAUDE.md duplication matches the file's heavy-annotation convention.

Awaiting green CI + the online review bot's verdict at HEAD before merge.

Issue #223, opened 2026-06-10
Liveness ladder follow-ups: Codex open-FD binding + instant exit detection (kqueue/pidfd)


IvanWng97:
## Problem
The sessions-registry probe covers CC only. Two follow-up rungs from the prior-art research (c9watch/Irrlicht/clawd-on-desk):
1. **Codex**: no first-party registry, but Codex HOLDS its rollout file open for the session lifetime (unlike CC) — open-FD binding gives an exact pid↔transcript join: /proc/<pid>/fd readlink on Linux, lsof-style proc_pidfdinfo on macOS. Today Codex mid-attach still relies on the 1h mtime window (idle Codex sessions stay hidden until next write).
2. **Instant exit detection**: abrupt exits (Ctrl-C) currently fall to the slow stale sweep. Once a slot is bound to a pid (registry or FD), kqueue EVFILT_PROC NOTE_EXIT (macOS) / pidfd (Linux) gives immediate, race-free exit events.

## Why deferred
Both are platform-FFI features (per-source capability in SourceDescriptor), independent of the mid-attach PR's correctness; each needs its own adversarial review of unsafe code (#201 precedent).

Origin: prior-art research synthesis on the mid-attach visibility PR.

IvanWng97:
Rung 1 (Codex open-FD binding) shipped in #227 (merged 588f78d): live_codex_rollout_ids via source/fd_probe.rs (macOS libproc FFI on the existing libc dep / Linux /proc), wired through with_liveness_probe + codex_id_from_path, plus ProofOfLife ongoing-liveness so the FD vouch also exempts live sessions from staleness sweeps (an idle-but-open Codex TUI no longer ghosts; the 5-min short-idle reap now effectively fires after the process exits). Remaining scope of this issue = rung 2 only: instant exit detection via kqueue EVFILT_PROC NOTE_EXIT (macOS) / pidfd (Linux) to replace the sweep-latency exit path with race-free immediate events.

PR #684, opened 2026-07-19
refactor(core): give the #223 probe ladder one owner (ProbeLadder, a pure failure-detector)

IvanWng97:
## What

Deferred arch-deepening candidate **#2** from the `/improve-codebase-architecture`
review: give the #223 probe ladder's scattered derived-liveness state **one
owner**, designed as a **functional core / imperative shell**.

Before: `NegativeVouch` owned the vouch hysteresis, but the `pid → ids` bindings
were a bare `ScanState` map with **three writers** — the rebind loop, the vouch's
confirm-unbind (via a `&mut` param), and the `exit_rx` arm poking
`scan_state.pid_bindings.remove` — and the `forget`↔`unbind` inverse was a prose
contract split across a free fn + a method.

After — one `ProbeLadder`:
- **`fold(snap, now) -> ProbeOutcome { exits, newly_watched }`** is **pure**: it
  advances the miss windows + pid bindings and *returns the effects* (exit ids to
  emit, new pids to watch). It never touches `ctx`/`tx`/the exit watch. `now` is a
  parameter → a unit-testable snapshot→exits state machine with synthetic time.
- The **shell** (`refresh_probe_snapshot`) applies the outcome **emit-before-watch**
  — the pre-owner order — so the `AgentEvent` stream is byte-identical.
- **`pid_died(pid) -> Vec<Id>`** replaces the `exit_rx` map poke; `forget`/`unbind`
  are now private methods one owner keeps in lockstep.

## Why it's safe (byte-identical, empirically proven)

Behavior is **not** merely argued — it's pinned by the 10 black-box shell tests in
`tests/watcher/liveness.rs` (negative vouch, instant exit, rebind migration, probe
failure/panic), all green. `exit_watch.watch()` sends nothing synchronously and
each just-confirmed id is absent from `snap.pid_of`, so the emit-then-watch order
produces the identical channel sequence.

The detection **mechanism** is untouched and already best-practice (pidfd / kqueue
`NOTE_EXIT` + start-time recycle check + 60s hysteresis) — this reorg only gives
its scattered **state** an owner.

## Test

- **5 new pure-ladder unit tests** exercise the new interface (the deepening's
  payoff — the interface IS the test surface): sustained-miss confirm,
  reappearance-cancels-window, new-pid-once, rebind-migration-keeps-sibling,
  pid_died-disarms-vouch. Zero mocks, synthetic time.
- The 2 former white-box tests that drove the internal map are **replaced** by
  those (the map is now private).
- Full `pixtuoid-core` suite green; clippy clean; docs (`crates/pixtuoid-core/CLAUDE.md`)
  updated to cite `ProbeLadder`/`pid_died`/the FCIS split; repo-wide stale-grep 0.

## For the reviewer

- **Highest-blast-radius subsystem** (liveness). The one thing to verify is the
  emit-before-watch ordering claim — the 10 black-box tests are the teeth.
- Design was grilled against two web-searched best practices: **Functional Core /
  Imperative Shell** (return effects as data, incl. the `watch()` commands →
  single-method AND byte-identical) and process-death detection (pidfd/kqueue +
  recycle check — confirmed already-correct, so only structure changed).
- Separate PR from the pixel-painter cleanups (`#10`/`#11`) on purpose — disjoint
  crate, disjoint blast radius.

https://claude.ai/code/session_0121zsMsaPE6YZgv6UrAXUVE


IvanWng97:
## Two-lens merge-gate review — disposition (head `32865ea1`)

**Scope:** diff. **5 differentiated lenses**, each MEDIUM+ finding adversarially verified. **Claim under test:** the FCIS `ProbeLadder` refactor is behavior-preserving in the highest-blast-radius subsystem (liveness).

| Lens | Result |
|---|---|
| correctness + grounding | **clean** — all 4 scenarios (sustained vouch / instant exit / rebind / probe-fail) traced identical; emit-before-watch keeps the AgentEvent sequence byte-identical (each confirmed id is absent from `snap.pid_of`; `watch()` sends nothing sync) |
| #461 byte-identity hunt | **clean** — every moved piece (fold vs observe, rebind vs the loop, pid_died vs exit_rx, shell emit+watch) diffed old-vs-new, observably identical |
| liveness / concurrency | **clean** — single-task ladder mutation; ExitWatch thread speaks only via `exit_tx`; no ghost/missed-exit window opened by the reorder |
| design / blast-radius | 1 **NIT** → **REFUTED-WITH-TRACE** ↓ |
| test-teeth | 1 **LOW** → **FIXED** ↓ |

### Findings → terminal state
- **LOW — the "unbind drops an emptied pid" invariant lost its only test** (the deleted white-box `unbind_session_drops_only_emptied_pid_entries` pinned it; the new pure tests covered keep-a-sibling but not drop-emptied — load-bearing, since a leaked `{pid:{}}` reads as "not newly seen" and would skip an exit-watch re-registration). → **FIXED** in `32865ea1` with an interface-level test (`fold_drops_a_pid_emptied_by_a_confirmed_exit_so_a_reused_pid_re_registers`).
- **NIT — `ProbeOutcome`'s `pub` fields on a `pub(super)` struct** → **REFUTED-WITH-TRACE**: `unreachable_pub` (the mechanical definition of the repo's pub-discipline, `Cargo.toml` + `CLAUDE.md`) does not fire on a reachability-capped field, clippy `-D warnings` is clean, and it matches **49** sibling bare-pub-capped structs — direct twins one crate over (`pixel_painter::ambient::SunbeamColumn`, `creatures::GatewayMascotDef`). Sanctioned pattern, no action.

**Empirical:** 15/15 (5 new pure-ladder + 10 black-box shell) + full `pixtuoid-core` suite green; clippy clean; repo-wide stale-grep 0.

**Merge head `32865ea1`.** All findings terminal. The online bot should re-review this head (expect `Findings: 0`). **A human merges.**

PR #685, opened 2026-07-19
refactor(scene): single-source desk-shadow cy + ceiling-pool geometry

IvanWng97:
## What

Two **byte-identical** background-pass cleanups from the
`/improve-codebase-architecture` review (deferred candidates **#10** + **#11**),
both verified against current code.

| # | Concentrates |
|---|--------------|
| **#10** | the desk floor-shadow `cy` hardcoded `desk.y + 7` — a **second copy** of `desk_furniture_def().visual.h` (= `DESK_H+2`) that the desk z-key (`enqueue_desk_cubicles`) already derives. A `DESK_H` retune would move the z-sorted sprite but leave the shadow behind (the magic-number / latent-drift class the `CLAUDE.md` convention forbids). → `desk_shadow_ellipse` reads `visual.h` from the one authority. |
| **#11** | the three inline ceiling-pool ellipse blocks (per-desk + pantry + corridor) → a `ceiling_pool_regions(layout)` iterator, naming the pool half-extents (`10/5`, `12/6`, `14/5`) + the desk cy-lift as consts instead of anonymous inline literals. |

The half-extents stay inline (per-piece taste, no authoritative source); the
floor-lamp halo stays a **separate** painter and the paint order (pools → lamp)
is preserved, so byte-identity holds.

## Why they're safe

Byte-identical by construction: `visual.h == 7` today, and the iterator yields the
same ellipses in the same order. Pinned by the existing
`paint_frame_is_pure_and_byte_identical` guard + **two new teeth** (the shadow cy
tracks the z-key row; the pools yield desks→pantry→corridor at the expected
extents). No pixels change → `gen-check` goldens are unaffected.

## Not done (deliberately)

The full floor-shadow **table** consolidation (`#10`'s broader form) was **dropped**
as *just-moves-shallow* — the couch/printer/generic shadows are hand-tuned per-kind
constants + iteration exceptions with no per-def derivation. Only the desk `cy`,
which has a real authority (the z-sort row), reads from the def. Shadow deliberately
**stays render-side** (not moved onto `FurnitureDef`): its strength is a daylight
function only the painter uses.

## Test

Full `pixtuoid-scene` suite green (642 tests) + the two new teeth. Separate PR from
the `ProbeLadder` liveness refactor (`#2`) on purpose — disjoint crate, disjoint
blast radius.

https://claude.ai/code/session_0121zsMsaPE6YZgv6UrAXUVE

### DIFF
--- a/crates/pixtuoid-scene/src/pixel_painter/mod.rs
+++ b/crates/pixtuoid-scene/src/pixel_painter/mod.rs
@@ -364,6 +364,58 @@ pub fn render_to_rgb_buffer(ctx: &mut PixelCtx<'_>) -> PixelPassResult {
     }
 }
 
+/// The soft floor shadow under one home desk. `cy` sits on the desk's z-sort
+/// row — `desk.y + visual.h` — the SAME `visual.h` `enqueue_desk_cubicles` keys
+/// the desk sprite on, read from the one authority (`desk_furniture_def`) rather
+/// than re-hardcoded, so a `DESK_H` retune moves the shadow WITH the sprite's
+/// south base instead of leaving it behind. The half-extents stay inline: a
+/// shadow's width/depth is a per-piece taste value with no source to derive from.
+fn desk_shadow_ellipse(desk: Point) -> Ellipse {
+    Ellipse {
+        cx: desk.x + DESK_W / 2,
+        cy: desk.y + crate::layout::desk_furniture_def().visual.h,
+        half_w: DESK_W / 2 + 1,
+        half_h: 3,
+    }
+}
+
+/// The ceiling-fluorescent light pools, in paint order: one narrow tube per
+/// desk, then a wider fixture over the pantry and the corridor. THE authority
+/// for where the pools sit + how wide each glows, so the paint pass is a short
+/// loop and the per-region ellipse extents aren't anonymous inline literals. The
+/// floor-lamp halo is deliberately NOT here — it is a different painter
+/// (`paint_floor_lamp_halo`) with its own strength + south-offset anchor; the
+/// paint order (pools THEN lamp halo) is load-bearing for byte-identity.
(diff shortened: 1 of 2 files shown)


IvanWng97:
## Two-lens merge-gate review — disposition (head `13162fc5`)

**Scope:** diff. **4 lenses** (scaled to the cosmetic blast radius), each MEDIUM+ finding adversarially verified. **Claim under test:** both cleanups are byte-identical (render output unchanged).

| Lens | Result |
|---|---|
| byte-identity (#461) | **clean** — `visual.h == DESK_H+2 == 7` today so the shadow cy is unchanged; `ceiling_pool_regions` yields the same ellipses in the same order (desks→pantry→corridor), `saturating_sub(2)` preserved |
| render / gen-check / order | **clean** — pool paint order preserved (pools THEN lamp halo); floor-lamp halo correctly excluded; goldens won't move (no regen needed) |
| design + magic-# convention | 1 **NIT** → **FIXED** ↓ |
| test-teeth | 1 **NIT** → **REFUTED-WITH-TRACE** ↓ |

### Findings → terminal state
- **NIT — `desk_shadow_ellipse` doc overstated** ("half-extents have no source"), but `half_w: DESK_W/2+1` DOES track `DESK_W` (the same authority `cx` uses); only `half_h: 3` is source-less taste. → **FIXED** in `13162fc5` (comment-only reword).
- **NIT — "Pinned by `paint_frame_is_pure_and_byte_identical`" overstates coverage** (that test proves idempotence, not equivalence-to-old) → **REFUTED-WITH-TRACE**: the phrase lives only in the commit/PR prose, not the code, and the real byte-identity backstop IS the CI `gen-check` reference-png diff, which the PR body already names. No code defect.

**Empirical:** 172 `pixel_painter` tests + the 2 new teeth green; `paint_frame_is_pure_and_byte_identical` green; scene suite green.

**Merge head `13162fc5`.** All findings terminal. Bot re-review expected `Findings: 0`. **A human merges.**

PR #688, opened 2026-07-19
refactor: resolve 3 arch-review deepening candidates + 2 fold-in cleanups

IvanWng97:
## What

Five small **deep-module** refactors surfaced by an `/improve-codebase-architecture` pass (survey + adversarial-verify workflow over all 12 subsystems). Each was verified against the deletion test **and** the repo's documented sharp edges before landing.

**Headline: the tree is already deep.** Of 12 subsystems surveyed, only **3 candidates survived** (all "Worth exploring", none "Strong"); 9/12 verified deep, every rejection citing code or a sharp edge. These changes are polish, not structural fixes — all are **behaviour-neutral**.

## Changes

| Commit | Change | Deletion test |
|--------|--------|---------------|
| `d0ab2f72` | **core** — delete `TestRenderer::render()`/`count()` retired-trait residue (duplicated `record()`, driven only by its own self-test) | vanishes |
| `7ff15afa` | **runtime** — one owner for the grow-only `SourceDeath` cursor: `unseen_deaths` → `runtime/mod.rs`, both consumers (headless driver, floating window) ride it | concentrates |
| `f11e7392` | **scene** ★ — `pose::pure::in_thinking_window`: one owner for the `was_active && Δt < THINKING_WINDOW_SECS` gate the pure overlay and the routed render computed byte-identically | concentrates |
| `695b0fbd` | **audio** — single-source the 4 synth RNG seeds (`BUILD/TYPING/DROP/PICK_SEED`); native + wasm must seed identically or they build different frozen buffers | — |
| `695b1a97` | **install** — pin the shell-quote write/read pair with a `posix_unquote(shell_single_quote(p)) == p` round-trip test | — |

### Notes for review

- **C2 scope, verified.** The survey also proposed a `WanderKind::at_waypoint_pose` method — **rejected**: `motion/` holds zero `Pose` references, so it would invert the `pose → motion` layering; the two `WanderKind → Pose` match arms are compiler-exhaustive anyway and can't silently drift. Only the boolean-predicate half (which the compiler *can't* catch drifting) survived.
- **F1 does not regenerate `site/public/wasm/`.** The seed *values* are unchanged (output-identical), CI's wasm-check is build-only with no byte-compare, and regenerating would inject rustc-version drift noise — same output-driven-regen rule as the committed demo media. The web crate compiles clean on `wasm32-unknown-unknown`.
- **One fold-in was deliberately NOT done:** routing the ~24 inline `if px<w && py<h { buf.put() }` clip guards through the existing `Grid::set`. Reading the real code: only ~5 sites are the bare shape; the rest wrap a computation block (converting hoists per-pixel work to run even when clipped — a hot-loop behaviour change), and the guard is the uniform house idiom across dozens of paint sites. Converting a subset creates two idioms for one op — worse for locality than the consistent status quo. Left as-is.

## Verification

- `just preflight` green locally: **2541 tests passed, 2 skipped, 0 warnings** (lint → clippy `-D warnings` → cargo-hack feature powerset → full suite).
- New/updated tests: `pose::pure::tests::in_thinking_window_gates_on_recency_and_prior_activity`, `runtime::tests::unseen_deaths_yields_each_death_exactly_once` (relocated), `install::hook_cmd::unix::tests::quoting_round_trips_through_the_verify_reader`.
- Docs-currency: `render/mod.rs` + `crates/pixtuoid-core/CLAUDE.md` sharp edge updated for the `TestRenderer` residue removal.

Still needs the two-lens review merge gate.


IvanWng97:
## Two-lens review (diff scope, merge gate) — ✅ all lenses APPROVE, 0 findings

Ran the repo's mandatory pre-merge review on `main...HEAD`. Two escalation triggers fired (dedup/consolidation; "behaviour-preserving refactor"), so **4 differentiated lenses** rather than the 2-lens floor. Each did substantive work (14–21 tool calls hitting the real targets — not a rubber-stamp), and every MEDIUM+ finding would have been adversarially verified before counting (there were none).

| Lens | Focus | Verdict |
|------|-------|---------|
| Correctness / grounding | byte-identity per claim · sibling-set (all seed sites migrated) · CI-only gates | APPROVE · 0 |
| Design / blast-radius | seed home & doc-hidden semver surface · layering · DRY residue | APPROVE · 0 |
| Wrong-abstraction (adversarial-toward-revert) | do the shared owners each have one reason to change? | APPROVE · 0 |
| Behaviour-identity (per-call-site) | each edited site IDENTICAL vs MOVED; the "one hidden mover" hunt | APPROVE · 0 |

**Key confirmations:**
- **Byte-identity held at every site** — C2 `in_thinking_window` equals both former inline expressions (incl. `Duration::ZERO` fallback + integer `as_secs()`) and the routed short-circuit still runs before `advance_wander`; C3 `unseen_deaths` == the old `skip(deaths_seen)` for the grow-only watch Vec (they diverge only at `seen > len`, unreachable); F1 all 4 seed **values** unchanged across the 8 native + 4 web sites; C1 `render()`/`count()` had no consumer but the deleted self-test.
- **Consolidations are sound** — each shared owner has one reason to change (thinking-window definition / grow-only cursor invariant / native↔wasm buffer lockstep); the deliberate pure-vs-routed divergences (dwell constants) stay inline.
- **CI-only gates reasoned CLEAN** — semver: the `TestRenderer` feature-gated removal + the doc-hidden `pub const` adds are covered by the already-open `0.15 → 0.16` minor; gen-check: behaviour-neutral (no rendered frame changes, F1 is audio); wasm-check is build-only (web compiles clean on `wasm32`); `windows-test` unaffected (F2 is `cfg(unix)`).

**Still required before merge (not done by me):** green CI at HEAD + the online review bot's `Findings: 0` on this commit + a human merge. I don't merge.

IvanWng97:
**Risk-radar motion/pose item — disposed (discharged).** C2 (`pose::pure::in_thinking_window`) trips the standing "render and watch" escalation, but it's a **provably byte-identical predicate extraction** (same 3-line `was_active && Δt < THINKING_WINDOW_SECS`, same two call sites, same control-flow position before `advance_wander`) — confirmed independently by the online bot, the local behaviour-identity lens, and a new `in_thinking_window_gates_on_recency_and_prior_activity` pinning test. Empirically: **`smoke` is green (`just gen-check` pixel-diff)** + `snapshot hygiene` green, so the rendered office is identical to baselines — there is no motion behaviour change to watch. The SeatedThinking gate is a discrete pose selection (not a walk animation), which the fixed-time gen-check renders cover. No live-replay blocker.

PR #709, opened 2026-07-20
feat(web,site): Web Worker synthesis — the ♩ click becomes upload-only

### LINKED ISSUE #705: site: pre-warm the wasm audio synthesis at page idle — make the play click near-instant
The click-to-sound wait (~2s) on the site hero is the synthesize-at-launch cost: today the ♩ click starts the warmup (bank → rain → all five beds), pumped off setTimeout so the click handler returns, but stage 2 still builds five beds in one chunk and the user waits for sound.

Two-part fix, both enabled by machinery #704 landed:
1. Pre-synthesize at page idle: synthesis needs NO user gesture (only AudioContext creation does). Run the warmup in the background after page load; the ♩ click then only creates the context + uploads buffers — near-instant sound.
2. Split warmup stage 2 per-bed via synth::gen_bed (the chunk unit the 10-min swap rebuild already uses), so no single main-thread block exceeds one bed even during pre-warm.

Care: keep the deterministic track selection consistent between pre-warm time and click time (pre-warm the CURRENT track_epoch block; if the block flips before the click, the normal chunked swap handles it).

IvanWng97:
Closes #705. Synthesis moves OFF the main thread: at page idle a module worker (`site/public/audio-worker.js`) loads its OWN wasm instance, runs the new `SynthTake` export (the same `WebAudioDriver` warmup, pumped to done — blocking is fine there), and transfers the buffers back; the page adopts them piece-by-piece (`Office::audio_adopt_begin/_oneshot/_loop/_finish`, one small memcpy per `setTimeout(0)` tick). After adoption the ♩ click is **upload-only — near-instant**.

Design keys:
- **Adopt-into-driver, zero divergence.** The engine tick never touches sample data (bank/rain/beds are only the upload byte source), so an adopted driver is state-identical to a locally-warmed one — upload, tick, and the #707 chunked swap downstream never know the difference.
- **First ready wins.** A ♩ click during the prewarm defers to the settle (never two drivers racing one office); `audio_adopt_finish` refuses to stomp a click-warmed driver; every push validates order/bounds and finish validates completeness, so a torn handoff (dead worker mid-stream) falls back to the click-time chunked pump.
- **The worker is an accelerator, never a dependency.** Dead worker / no module-worker support / reduced-motion (skipped as a low-power signal) all land on the intact fallback path — pinned by a dead-worker e2e. The 10-min swap deliberately stays on the main thread (inaudible under the silence hold; routing it through the worker would be scope without a symptom).
- Staleness self-heals: an adopted take whose 10-min epoch flipped before the click catches up through the normal swap machinery.

Tests: Rust — byte-for-byte adoption round-trip, adopted-driver tick/swap parity, torn-handoff refusals, first-ready-wins, worker/main determinism pin. e2e — adopted path pins click→ready < 2.5s (the replaced synthesis path measures seconds), dead-worker fallback still reaches ready. Gates: preflight green (2623 tests), `just site-check` green, full e2e 89/89 twice consecutively, wasm regenerated (own chore commit).


IvanWng97:
**Two-lens review disposition** (merge-gate, lenses differentiated):

- **Lens 1 (correctness/grounding): APPROVE-WITH-NITS, 1 LOW** — a ♩ click landing before the idle callback would let `prewarmSpawn` burn a redundant worker synthesis whose result the finish guard refuses (never incorrect, purely wasteful). **Fixed** at `6a94f05c`: spawn guard also skips on `audioReady`/`audioBooting`. Cleared after tracing: detached-view hazards, transferred-buffer reuse, double-driver races, track reconstruction exactness, adoption validation, test teeth, docs grounding.
- **Lens 2 (design/blast-radius/drift): APPROVE-WITH-NITS, 1 MEDIUM→LOW** — the worker's loop bound minted a fresh JS copy of `LoopStem::ALL.len()` with no pin. **Fixed** at `6a94f05c`: new `SynthTake::loop_count()` getter is the one authority (worker copy-out + adopt pump + lib tests all derive from it); `ONESHOT_WIRES` got its pairing comment (authority = `pool_from_wire`'s domain, round-trip-pinned in Rust). Also took the optional torn-stage cleanup (failure branches consume the stage via `audio_adopt_finish`, freeing the staged MBs). Cleared: CSP (same-origin worker under `default-src 'self'`; empirically backstopped by the adopted-path e2e), `asset()` URL resolution, pub/pub(crate) discipline, knip/prettier/eslint on the new file, gen-wasm consistency, no `__pixWasm`/VIBING interaction, docs currency, no semver surface.

Both lenses ran gates (pixtuoid-web 39/39, clippy, site-check) rather than eyeballing. Local full-e2e flake note is in the fix commit body — CI (near-serial) is the arbiter and passed at the prior push.

PR #712, opened 2026-07-20
fix(site): hero strip first-row top border shaved on iOS — pad the composited layer

IvanWng97:
Owner phone report: the hero badge grid's FIRST row loses ~1px off its top edge on iOS Safari (flat shave, rows 2+ fine — user-confirmed shape B, ~1px).

Mechanism: `.hero__badges` is a pinned compositing layer (`translateZ(0)`, the existing Safari hover-jitter fix) with `padding: 0`, so row 1's 1px chip border sits on the layer's exact top raster edge; iOS rasterizing the layer at a fractional offset (measured 421.05px locally) drops the topmost device-pixel row. Interior rows are immune; headless Chromium doesn't reproduce (different rasterizer). Ruled out by pixel-probe + rendered crop at iPhone metrics: geometric clipping (none), overlaying elements (none at ±3px), translucent chip fill (opaque).

Fix: 2px of the strip's top margin moves INSIDE the layer as `padding-top` (margin compensated via `calc` — net geometry identical), so the raster boundary crops padding instead of the border.

Verify on-device after deploy: the first row's top border should render at full weight.

Gates: `just site-check` green, `just site-e2e` 89/89.

### DIFF
--- a/site/src/components/Hero.astro
+++ b/site/src/components/Hero.astro
@@ -124,8 +124,16 @@ const badges = sources.filter((s) => s.status === 'supported');
     display: flex;
     flex-wrap: wrap;
     gap: 0.5rem;
-    margin: 1.15rem 0 0;
-    padding: 0;
+    /* 2px of the old margin moved INSIDE the layer as padding-top (net
+       geometry identical): with padding 0 the first row's 1px chip border
+       sat on the composited layer's exact top edge, and iOS Safari
+       rasterizing the pinned layer at a fractional offset shaved that
+       device-pixel row — the row-1-only "top of the badges is clipped"
+       phone report. Rows 2+ are interior to the layer and were never hit;
+       headless Chromium doesn't reproduce it (different rasterizer). The
+       padding keeps the border off the raster boundary. */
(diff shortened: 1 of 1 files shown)


IvanWng97:
**Two-lens review disposition**: Lens 1 (correctness/grounding) APPROVE — verified net-identical geometry in every layout context (one rule, no breakpoints, BFC = no margin collapse), no hover/hit-test change, compiled-dist cascade checked, and found the calc compensation is LOAD-BEARING (padding-only would exceed the ≤7.9vh scroll-budget pin's ~1px headroom). Lens 2 (design/blast-radius) APPROVE-WITH-NITS — independently confirmed this is the correct minimal lever (outline/chip-padding/de-compositing all fail or regress the hover-jitter fix), robust to the entrance-animation variant, scroll pin unaffected at any font scale. Their shared nit (comment over-attributed the layer to translateZ — the rise animation's fill overrides it at steady state) + the load-bearing-calc note are folded in at `57ccece8`. On-device confirmation happens post-deploy (the mechanism is CI-unverifiable by nature; the shape — B, ~1px flat shave, row 1 only — is owner-confirmed).

PR #724, opened 2026-07-20
refactor(site): share the wasm-Office driver across the two hero consumers (#721)

### LINKED ISSUE #721: Share the wasm-Office driver (sharedWasm + officeFrameView) across the two hero consumers
## Sweep-3 ⑯ — deferred (site refactor, separate verification lane)

The architecture-deepening round-3 scan surfaced a duplicated wasm-`Office` canvas driver across the two live-office Astro consumers:

- **`site/src/components/OfficeBackdrop.astro`** (`boot()` ~1197, `paint()` ~627, `initSharedWasm` IIFE ~1220)
- **`site/src/components/Showcase.astro`** (`vPaint()` ~308, `initSharedWasm` IIFE ~348)

### What's duplicated
1. **`initSharedWasm` IIFE — byte-identical** in both: the memoized (`window.__pixWasm`) bounded-retry (`WASM_INIT_RETRIES=2`, `WASM_INIT_BACKOFF_MS=400`) wasm-glue import. The retry CONSTS are already pinned by `site/config/wasm-init-consts.test.mjs`, but the whole STRUCTURE is copy-pasted.
2. **The frame-memory contract** — `new Uint8ClampedArray(wasm.memory.buffer, office.frame_ptr(), office.frame_len())` (re-read EVERY frame, since a resize/`memory.grow` moves them) + the `view.length !== img.data.length` length guard. Duplicated, NOT structurally pinned (only the const-drift is).

### Proposed extraction
A committed **`site/public/office-driver.js`** ES module (precedent: `site/public/audio-worker.js` is a runtime-loaded public module), injected via `define:vars` (`officeDriverUrl: asset('office-driver.js')`) and dynamically `import()`ed at boot, exporting:
- `sharedWasm(wasmJsUrl)` — the memoized retry-init (replaces both IIFEs).
- `officeFrameView(office, wasm)` — the `frame_ptr`/`frame_len` re-read view + length semantics (the frame contract in one place).

`Showcase.vPaint` can ride the full blit; `OfficeBackdrop.paint` diverges (its reveal-roll branch sits between the view and the final `putImageData`), so it uses `officeFrameView` + its own reveal/blit — the CONTRACT is single-sourced without forcing the divergent blit.

### Why deferred (not done in the sweep-3 Rust PR)
- Restructures BOTH consumers' boot flows (sync → an `await import(officeDriverUrl)` hop on the critical path) — a real cross-file refactor.
- The only real verification for a blit-contract change is **`just site-e2e`** (Playwright runtime): a mistake is a user-facing broken hero animation that `tsc`/`eslint` (`just site-check`) cannot catch.
- Cleanly separable as a site-consumer change (the site has its own CI + delivery lane).

### Reconcile on landing
Update `site/CLAUDE.md`'s VIBING entry: it currently pins only the retry-consts equality (`wasm-init-consts.test.mjs`); note that the frame contract + init now live in `office-driver.js`.

IvanWng97:
Closes #721. The `initSharedWasm` IIFE was **byte-identical** (verified line-for-line) across `OfficeBackdrop.astro` and `Showcase.astro`, and the frame-memory view (`new Uint8ClampedArray(wasm.memory.buffer, office.frame_ptr(), office.frame_len())` + length guard) was duplicated but — unlike the retry consts — **not structurally pinned**.

**Extraction** (per the issue's sketch): a runtime-loaded `public/office-driver.js` ES module (the `audio-worker.js` #705 precedent — is:inline scripts can't static-import a bundled `src/` module) exporting:
- `sharedWasm(wasmJsUrl)` — the ONE memoized `__pixWasm` bounded-retry init.
- `officeFrameView(office, wasm)` — the re-read-per-frame view (the frame contract in one place).

Both consumers dynamic-`import()` it at boot and drive through it. **OfficeBackdrop keeps its own reveal-roll blit** (it sits between the view read and `putImageData`), so only the *contract* is single-sourced, not the divergent blit.

**Verification lane** (the issue names this correctly — a blit mistake is a user-facing broken hero `tsc`/`eslint` can't catch): `just site-e2e` **89/89**, exercising both heroes' full boot→paint through the shared driver. `just site-check` green.

Notes: both consumers already boot the office **off the critical path** (OfficeBackdrop `load`→`requestIdleCallback`, Showcase IntersectionObserver), so the extra same-origin import hop isn't above-fold-critical; a failed driver fetch degrades to the poster through each caller's existing `.catch`. `wasm-init-consts.test.mjs` repurposed from *pin-two-copies-equal* → *guard-the-single-source* (consts in office-driver.js, neither consumer re-inlines). knip ignores the runtime-loaded module; `site/CLAUDE.md` VIBING entry reconciled.

### DIFF
--- a/site/CLAUDE.md
+++ b/site/CLAUDE.md
@@ -366,6 +366,10 @@ would make browsers *ignore* `'unsafe-inline'`. Consequences to not "fix":
 - **`script-src` carries no `'unsafe-inline'`** — every inline script is
   whitelisted by content hash, recomputed on each build. Adding/editing an
   `is:inline` script needs NO manual CSP step.
+- **A hand-written `public/*.js` module loaded by URL** (`audio-worker.js` via
+  `new Worker`, `office-driver.js` via `import()`) rides `script-src 'self'` as
+  an EXTERNAL resource — it is not hashed and needs no CSP step; only its
+  runtime-loading `is:inline` caller is (auto-)rehashed when its content changes.
 - **`style-src` keeps `'unsafe-inline'` and must stay hash-free**: Shiki
   spans, the build-time mermaid SVG, and the few `style={}` attributes are
   inline style ATTRIBUTES, which hashes cannot express (one present hash
--- a/site/CLAUDE.md
+++ b/site/CLAUDE.md
@@ -305,10 +305,17 @@ display-line authority (`starText`), unit-tested on its null-stars arm since
   stomp the single module-global `wasm`. A bounded retry (2×, 400ms backoff) wraps
   that shared init for transient wasm-fetch resilience, but stays INSIDE the one
   `__pixWasm` promise — a per-consumer retry would reintroduce the very
-  double-instantiate race above (the retry consts are pinned equal across the two
-  boot scripts by `config/wasm-init-consts.test.mjs`). On FINAL exhaustion the
-  promise is NULLED so a later-booting consumer re-attempts — safe because it only
-  clears a SETTLED (rejected) promise, never an in-flight one (#671). Schema: `kind:"live"`
+  double-instantiate race above. Since #721 that shared init AND the frame-memory
+  contract (`new Uint8ClampedArray(wasm.memory.buffer, office.frame_ptr(),
+  office.frame_len())`, re-read per frame) live ONCE in the runtime-loaded
+  **`public/office-driver.js`** module (`sharedWasm` / `officeFrameView`), which both
+  is:inline consumers dynamic-`import()` at boot (the audio-worker.js precedent — a
+  bundled `src/` import can't reach an is:inline script); the divergent blit stays
+  per-consumer (OfficeBackdrop's reveal-roll sits between the view read and the
+  `putImageData`). `config/wasm-init-consts.test.mjs` now pins that single source (the
+  consts live in office-driver.js and neither consumer re-inlines them). On FINAL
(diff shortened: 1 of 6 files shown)


IvanWng97:
**Two-lens review disposition**: Lens 1 (correctness/behavior-preservation) **APPROVE, Findings 0** — traced sharedWasm's check-and-set (no await gap → double-instantiate race stays closed), the new driver-fetch failure landing on the SAME poster `.catch` as a wasm-fetch failure, officeFrameView timing (driver assigned in the first `.then`, office in the nested one that gates paint — always non-null), and confirmed every inner `.then` body byte-preserved + no third consumer missed (the Float32 audio view is a distinct contract, correctly untouched). Lens 2 (design/CSP/docs) **APPROVE-WITH-NITS** — CSP clean (`import()` of a same-origin public .js rides `script-src 'self'`, external/unhashed, like the audio-worker.js precedent), precedent transfer + knip/prettier/eslint handling correct, no missed consumer. Both nits taken at `ed084623`: the 'not re-inlined' test now matches a DECLARATION (`const NAME =`) instead of a bare substring (a comment naming the const would have spuriously red-ed it in this comment-dense repo), and the CSP section documents that a hand-written `public/*.js` module rides `script-src 'self'`. Gates: site-check green, site-e2e 89/89 (both heroes' boot→paint through the shared driver), the repurposed unit test 4/4.

PR #726, opened 2026-07-21
New instructions for Homebrew

chrisfinazzo:
## Summary

New instructions for Homebrew - Pixtuoid is now available in Homebrew Core - along with a few minor copyedits in the README. Feel free to squash these commits together, they should have been done all at once.

## Related issue

N/A

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] New theme / sprite pack
- [ ] New `Source` adapter (another agent CLI)
- [x] Docs only
- [ ] Refactor / chore

## How I tested it

N/A

## Visual verification (required for sprite / rendering changes)

- [x] Not a visual change, or — screenshot/GIF attached below and it reads at half-block scale.

## Checklist

- [ ] I read the relevant `CLAUDE.md` (root + the nested one for the crate I touched).
- [ ] New behavior has a test (this repo is TDD-first).
- [ ] No `unwrap()` in non-test code; errors propagate via `anyhow`/`thiserror`.
- [ ] No new `println!`/`eprintln!` on a production path (use `tracing`).
- [x] Docs updated in the same commit if I changed module structure, architecture, or public API (`CLAUDE.md` / `README.md`).
- [ ] Checked against the [recurring pitfalls](https://github.com/IvanWng97/pixtuoid/blob/main/docs/CONTRIBUTING.md#recurring-pitfalls-this-codebases-review-history-distilled): char-safe slicing · no parallel copies without a bridge test · sanitize at the decode boundary · negative branches pinned.
- [ ] `just preflight` passes locally.

## AI assistance

- [ ] This PR was written/heavily-assisted by an AI agent.

PR #729, opened 2026-07-21
chore: finish the homebrew-core migration (#726 follow-up)

IvanWng97:
Builds on #726 (@chrisfinazzo) — their two commits are included here with authorship intact. **Supersedes #726.** Merge with **rebase** so their commits land under their own name.

> **Scope:** this PR is about the *already-released* 0.15.0 being in homebrew-core. It changes no version and triggers no release. The 0.16 notes below are recorded for whenever that release happens — they are not a prerequisite for merging this.

## Why #726 alone doesn't land

#726 hand-edited the README's install block, but that block is **generated** from `site/src/install.json`. `just gen-readme-check` was red on it, and the next `just gen-readme` would have silently reverted the fix. Moving the change into the JSON fixes the README, the site's Install tab and the CI gate together — and reproduces the contributor's text byte-for-byte.

It also couldn't see the surfaces the README doesn't reach: the lead-in sentence *outside* the generated markers (still claimed macOS-only), the site e2e's clipboard assertion pinned to the old tap string, and Raycast's README + its binary-not-found runtime string.

## Retiring the tap

`release.yml`'s homebrew job published a second formula at the same version core already ships, from prebuilt binaries, installing neither a man page nor shell completions. Two formulae for one project, only one of them maintained. The job is removed and `IvanWng97/homebrew-pixtuoid` is deprecated in favour of core (deprecated, not deleted — existing installs keep resolving and get a migration warning).

`HOMEBREW_TAP_TOKEN` is now unused and can be revoked.

## Treating homebrew-core as a real consumer

Core's formula is coupled to us in ways the tap never was. Its `test do` asserts the exact `connect --json` envelope, the literal `OK: pack "skeleton"` line and the `claude-code` source id; its `install` opens with `rm ".cargo/config.toml"` and depends on the `man` / `completions` subcommands. This PR marks each of those at its source and records the asymmetry in `CLAUDE.md`: Raycast and the site fail in *our* CI, homebrew-core fails in *theirs*, while our suite stays green because it asserts those same strings as its own goldens.

`packaging-build` mirrors how core actually builds us, which no existing job does — `cargo install --locked`, default features, on both bottled platforms, then a replay of core's test block. It catches a stale `Cargo.lock` (every other job runs plain cargo, which silently *updates* it; `--locked` makes that a hard error) and any newly-required system library.

## Noted for a future 0.16 release — not for this PR

`main` turns on the `audio` feature by default, which pulls `rodio → cpal → alsa → alsa-sys` on Linux (macOS is unaffected — CoreAudio). Core's formula currently declares only `depends_on "rust" => :build`, and `release.yml` builds Linux artifacts `--no-default-features`, so nothing we own builds the configuration core builds.

Whenever 0.16 is eventually tagged, the core formula will need `alsa-lib` (Linux) and `pkgconf` (build) added **in the same PR as the version bump**. Separately, core has no `livecheck` block, so it falls back to the Git strategy's `DEFAULT_REGEX = /\D*(.+)/` — under which a `v*-rc.N` tester tag parses as a version above `0.15.0`. Worth a strict regex before any pre-release tag is pushed publicly.

Neither is a blocker here. `packaging-build` is what turns both into a red CI job instead of a surprise.

## Verification

`just preflight` **0** (2639 tests) · `just lint` **0** · `just site-check` **0** · `gen-readme-check` **0** · Raycast `tsc`/`eslint` **0**.


IvanWng97:
## Two-lens review — round 1 disposition

Four lenses (correctness · design/blast-radius · editorial · release-pipeline). Escalation triggers that fired: **public-facing artifact** (README / site Install tab / Raycast copy) and **blast radius** (deleting a job from the irreversible publish pipeline).

**All four returned REQUEST-CHANGES and converged independently on the same blocker.** Every finding below is at a terminal state; fixes are in `68590964`.

### FIXED

| Finding | Lenses | What was wrong |
|---|---|---|
| `ci.yml` asserted a `depends_on "alsa-lib"` that has never existed, and the apt step made the job green through the one gap actually outstanding | 1·2·3·4 (92–95) | The formula's whole history is two commits; its only dep is `rust => :build`, correctly — it builds v0.15.0, which predates `audio`. Tracked as #731 |
| `--version` captured to a file, never asserted | 1·2·4 (80–95) | It's the formula's *first* assertion and a **content** check `set -e` can't make. Dead oracle, in an oracle-shaped job, on a branch stacked on `fix/false-green-oracles` |
| "~an hour before `release.yml` finishes" | 2 (93) | Invented. Last four releases: 10/9/9/8 min. Number removed — the argument only needs "before" |
| "irreversible crates.io publish" stated in 4 places, 2 updated | 2 (88) | `CLAUDE.md:179`, `justfile:537`, `CONTRIBUTING:73` were stale. Lens found one; sweeping found three |
| `npm/generate.test.mjs` still cited "a Ruby heredoc" | 4 (85) | That heredoc *was* the job this PR deletes. THREE→TWO copies |
| Matrix legs shared one static check name | 4 (88) | `release.yml` interpolates `${{ matrix.target }}`; now matches |
| Contract enumeration omitted `man`/`completions` | 2·3 (62–85) | Those break core's **build**, not its test — strictly worse than anything marked. CLAUDE.md + `cli.rs` markers added, step renamed |
| Raycast error string: `cargo install pixtuoid` | 3 (80) | Never installs `pixtuoid-hook`, which Connect then writes into `settings.json`. Now matches `install.json` |
| README lead-in duplicated `install.json` outside the markers | 2·3 (78–80) | Already disagreed on capitalization two lines down. Now `Pick one:` — the generated block owns the platforms |
| `blurb` casing / `WSL` precision | 2·3 (66–82) | Reverted to `any OS`; `WSL`→`WSL2` (Homebrew supports WSL2 only) |

### REFUTED-STALE

**Tap has no migration path for installed users** (lens 4, 80) — `deprecate!` with `replacement_formula: "pixtuoid"` landed in the tap at `310e51f`, *before* that lens ran; the `pushed_at` it observed was that very push. Lens 2 independently confirmed it. No action.

### ISSUE-FILED

**#731** — homebrew-core must add `depends_on "alsa-lib"` (+ `pkgconf`, + a strict `livecheck` regex) in the same PR as whichever version first ships `audio`. Nothing in this repo can fix it: the change belongs in another repo's formula and the sha256 needs a tag tarball that doesn't exist.

### Open — maintainer decisions, deliberately not actioned

1. **Lost `caveats`** (lens 4, 75). The old tap printed uninstall hygiene ("disconnect before uninstalling or hooks stay in `settings.json`"). homebrew-core discourages `caveats`, so `brew install pixtuoid` now prints nothing. Real UX regression; the fix likely gets bounced upstream.
2. **Per-PR cost** (lens 2, 70). `packaging-build`'s value concentrates at release time but bills two cold release builds on every PR (`cargo install` uses its own target dir, so `rust-cache` largely misses). Suggested split: `cargo metadata --locked` in a cheap existing job for lock drift, full job on `release/**` or nightly.
3. **Not a required context** (lens 4, 88). `packaging-build` is absent from branch protection, so it currently cannot block a merge — advisory only. Repo setting; not touched.

### Gates at this head

`just preflight` **0** (2639 passed) · `just lint` **0** · `just site-check` **0** · `npm generate.test.mjs` **0** · Raycast `tsc`/`eslint` **0** · `gen-readme-check` **0**. Exit codes captured directly, never through a pipe.

Issue #731, opened 2026-07-21
homebrew-core: add `depends_on "alsa-lib"` before the first release that ships the `audio` feature


IvanWng97:
## Problem

`pixtuoid` is in homebrew-core with `autobump: true`. Its formula builds **from source with default features** on macOS *and* Linux.

The live formula (verified 2026-07-21) declares exactly one dependency:

```ruby
depends_on "rust" => :build
```

That was correct for **v0.15.0**, which has no `[features]` stanza at all — `git show v0.15.0:Cargo.lock | grep -c alsa` → `0`.

It is **not** correct for `main`. The `audio` feature (#633) landed after v0.15.0 and is on by default:

```toml
[features]
default = ["audio"]
audio = ["dep:rodio"]
```

`rodio` is not target-gated, so on Linux:

```
pixtuoid → rodio → cpal → alsa → alsa-sys
```

`alsa-sys`'s build script **panics** when `pkg-config` can't find ALSA. So the first autobump to a version carrying `audio` will hard-fail homebrew-core's `arm64_linux` + `x86_64_linux` builds.

Our own release never builds this configuration — `release.yml:126,198` ship Linux artifacts `--no-default-features`. homebrew-core is the only from-source default-feature Linux build.

## Fix

In the homebrew-core bump PR, **in the same commit as the version bump**:

```ruby
depends_on "pkgconf" => :build
depends_on "rust" => :build

on_linux do
  # rodio/cpal (the default-on `audio` feature) links ALSA on Linux.
  depends_on "alsa-lib"
end
```

Idiom verified against `Formula/t/termusic.rb` (Rust + cpal + ALSA). Note it is `pkgconf`, not `pkg-config` — homebrew-core renamed it.

Submitting the bump manually also preempts BrewTestBot, which would otherwise open a version-only PR that reds both Linux bottles.

## Also worth doing in the same PR

The formula has no `livecheck` block, so it falls back to the Git strategy's `DEFAULT_REGEX = /\D*(.+)/` (verified in `Library/Homebrew/livecheck/strategy/git.rb`). That strips leading non-digits and captures the rest, so a `v0.16.0-rc.1` tester tag parses as version `0.16.0-rc.1` — greater than `0.15.0`. `release.yml:46-47` deliberately allows those tags **at unmerged commits**, so an autobump could pick one up.

```ruby
livecheck do
  url :stable
  strategy :git
  regex(/^v?(\d+(?:\.\d+)+)$/i)
end
```

[...]

PR #732, opened 2026-07-21
chore: patch three raycast advisories + make packaging-build cacheable

IvanWng97:
Two unrelated follow-ups from the #729 merge, kept as separate commits so the security fix is readable on its own.

## 1. Three high-severity advisories in the Raycast extension (`b2247915`)

`npm audit fix --package-lock-only` — three patch bumps, lockfile only, `package.json` untouched:

| package | bump | chain | scope |
|---|---|---|---|
| brace-expansion | 5.0.6 → 5.0.7 | `eslint` → `minimatch@10` | dev |
| brace-expansion | 2.1.1 → 2.1.2 | `@raycast/api` → `@oclif/core` → `ejs` → `jake` → `filelist` → `minimatch@5` | **runtime** |
| js-yaml | 4.2.0 → 4.3.0 | `json-schema-to-typescript` | dev |

All three are ReDoS / quadratic-CPU advisories. Only the middle one sits under a runtime dependency.

**Why Dependabot couldn't do this**: its update job for these failed and opened no PR, because every one is a *transitive* dep absent from `package.json` — there is no manifest edit to propose, only a lockfile bump. That's exactly what this is.

`npm audit` now reports **0 vulnerabilities**. Diff is 9 insertions / 9 deletions — no incidental lockfile churn.

## 2. `packaging-build` was cold-building on every run (`df52e112`)

`cargo install` defaults to a **temporary** target dir it deletes on exit, so `Swatinem/rust-cache` had nothing to persist and both matrix legs did a full cold release build every time — the per-PR cost the #729 review flagged.

Setting `CARGO_TARGET_DIR` to the workspace `target/` puts the artifacts somewhere rust-cache actually saves. Verified locally that `cargo install` honours it: release artifacts land in the named dir and the binary still installs to `--root/bin`.

**The review's suggestion — gate the job to release branches or nightly — was rejected**, for a reason that only shows up when the two follow-ups are considered together: such a job also has to leave branch protection, since a required context that never runs on PRs blocks every PR. And the job exists to fail *before* an irreversible tag rather than in Homebrew's CI after one. Cutting the cost keeps both properties; cutting the coverage keeps neither.

Related: `packaging-build (ubuntu-latest)` and `packaging-build (macos-14)` were added to `main`'s required status checks (18 → 20), so this PR is the first to actually be gated on them — and the first real test of the caching change.

## Verification

`npm ci` 0 · `tsc` 0 · `eslint` 0 errors · `actionlint` 0 · `just preflight` 0 (the pre-push hook). Exit codes captured directly.

Not in scope: the `caveats` regression from the tap retirement (homebrew-core discourages `caveats`, so there is nothing to change), and #731 (needs a tag tarball that doesn't exist).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -326,8 +326,13 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-14]
     runs-on: ${{ matrix.os }}
-    # Two release-mode `cargo install`s from a cold cache; `cargo install` uses
-    # its own target dir, so rust-cache only partly covers it.
+    # Two release-mode `cargo install`s. `cargo install` defaults to a TEMP
+    # target dir it deletes on exit, so rust-cache would have nothing to save
+    # and every run would be a cold release build — the env below points it at
+    # the workspace target/ instead, which rust-cache does persist. First run
+    # on a new lockfile is still cold; the ceiling is sized for that.
+    env:
+      CARGO_TARGET_DIR: ${{ github.workspace }}/target
     timeout-minutes: 45
     steps:
       - uses: actions/checkout@v7
--- a/integrations/raycast/package-lock.json
+++ b/integrations/raycast/package-lock.json
@@ -1676,9 +1676,9 @@
       }
     },
     "node_modules/brace-expansion": {
-      "version": "5.0.6",
-      "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-5.0.6.tgz",
-      "integrity": "sha512-kLpxurY4Z4r9sgMsyG0Z9uzsBlgiU/EFKhj/h91/8yHu0edo7XuixOIH3VcJ8kkxs6/jPzoI6U9Vj3WqbMQ94g==",
+      "version": "5.0.7",
+      "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-5.0.7.tgz",
+      "integrity": "sha512-7oFy703dxfY3/NLxC1fh2SUCQ0H9rmAY+5EpDVfXjUTTs+HEwR2nYaqLv+GWcTsumwxPfiz6CzCNkwXwBUwqCA==",
       "license": "MIT",
       "dependencies": {
         "balanced-match": "^4.0.2"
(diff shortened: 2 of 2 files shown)


IvanWng97:
Closing in favour of a clean re-do. The two-lens review found one commit here to be a **no-op** and the other to **understate its own blast radius** — both verified independently before acting.

## `df52e112` (CARGO_TARGET_DIR) — dropped, it does nothing

I claimed `cargo install` uses a temp target dir it deletes on exit, so rust-cache had nothing to persist. That is true for **crates.io / `--git`** sources. This job installs with **`--path`**, and cargo's own docs say:

> If the source is crates.io or `--git` then by default the crate will be built in a temporary target directory. […] **When using `--path`, by default it will use `target` directory in the workspace of the local crate.**

Counterfactual test, which is the one I should have run the first time:

```
$ env -u CARGO_TARGET_DIR cargo install --locked --path crates/pixtuoid-hook --root <scratch>
$ ls target/release/pixtuoid-hook
target/release/pixtuoid-hook          # ← already in the workspace target rust-cache caches
```

My original "verification" only confirmed that *setting* the variable redirects the build. It never tested whether the problem existed. Setting it points cargo at the directory it was already using.

Worse than neutral: `CARGO_TARGET_DIR` is hashed into rust-cache's key (`config.ts` hashes any `CARGO*` env var), so the change forked the cache key and orphaned ~530 MB — on a cache measured at **10.54 GB against GitHub's 10 GB LRU ceiling**. Cost, no benefit.

## `b2247915` (security bump) — message was wrong on scope

The lockfile in that very diff contradicts its own commit message:

```
brace-expansion 5.0.7  dev=False   ← labelled "dev"; actually RUNTIME
brace-expansion 2.1.2  dev=False   ← correct
js-yaml         4.3.0  dev=True    ← correct
```

`minimatch@10` is reached from `@raycast/api` → `@oclif/core`, a production dependency. **Two of three are runtime-reachable, not one.** Also "three patch bumps" is wrong — js-yaml 4.2.0 → 4.3.0 is a *minor*, and necessarily so (4.3.0 is the advisory's first patched version). Understating blast radius in a security commit is the worst place to be sloppy, and the record is immutable once merged.

## Replacement

**#735** — same lockfile change, corrected message, no CI commit.

Also from the review: `site/` carries the same brace-expansion advisory. Not folded in, because `npm audit fix` there also wants astro 7.0.9 → 7.1.3 and a **major** `@antfu/install-pkg` bump — a dependency sweep wearing a security patch's clothes, and it collides with the open astro PR. Tracked as **#734**.

Standing issue worth its own decision: the Actions cache is over quota and LRU-evicting. Not caused by this PR; surfaced by it.

Issue #633, opened 2026-07-16
feat: ambient office sound — busy-ness-mixed lofi stems + procedural one-shots (hybrid, default off)


IvanWng97:
## What

Ambient sound for the office — a lofi bed that *reacts to how busy the office is*, plus reactive one-shots tied to moments the scene already models.

## Locked decisions (owner-ratified, 2026-07-16)

- **Q1 — Asset strategy: HYBRID.** AI-generated musical **stems** (committed OGG, mono, low-bitrate, ≲1–2MB total) for the beds; **procedurally synthesized** one-shots (pure Rust — filtered noise + envelopes, zero assets/license surface) for keyboard clacks / dings / glugs. AI assets follow the `gen-media` pattern: a `scripts/gen-audio` manifest records each asset's tool/prompt/params for reproducible provenance.
- **Q2 — Cue map v1:** keyboard clacks (density ∝ active-agent count), weather layer (rain/storm synced to visual weather), appliance one-shots (printer whir / vending drop / cooler glug — the 0.15.0 animations are the hook points), door chime on walk-in, elevator ding on floor nav. **NO needs-you sound in v1** (deferred; overlaps the future OS-notification feature). Chitchat murmur: cut.
- **Q3 — Default OFF.** Onboarding offers it once; `m` mute key; `[audio] enabled/volume` config.
- **Q4 — Architecture:** playback = `rodio`/`cpal` in the **binary only** (new `audio/` module) — core/scene stay audio-dep-free (invariant #1's sibling). The scene emits a semantic **`AudioCue` model** (the overlay/board precedent) that painters consume; the web hero can ride WebAudio on the same model later. Audio thread fully decoupled from render (drop cues on backpressure, never block).
- **Q5 — STEMS (the headline).** 3–4 stems of one composition (pad / keys / drums / texture) mixed at runtime by office busy-ness: empty office = pad only, agents active = layers fade in, rain = rain stem joins. Deterministic, zero runtime AI — "generative feel" from mixing, not generation. Extension point: `--sound-dir` user sound packs (the sprite-pack philosophy).

## Phased plan

[...]

PR #738, opened 2026-07-21
fix(release): install ALSA headers in the tag-time check job

IvanWng97:
The `check` job in `release.yml` runs `just clippy` + `just test` on ubuntu-latest with DEFAULT features. #633 made `audio` default → `rodio → cpal → alsa-sys`, whose build script **panics** when pkg-config can't find ALSA:

```
Pkg-config failed - usually this is because alsa development headers are not installed.
For Debian/Ubuntu users:  # apt-get install libasound2-dev
```

`ci.yml` installs `libasound2-dev` in eight jobs; `release.yml` installed it in zero. `v0.15.0` predates the audio feature (`git show v0.15.0:Cargo.lock | grep -c alsa` → 0), so the **first tag after #633 is the first to hit it** — in the one gate that only runs when it's already too late to iterate. This is not reachable from a PR (`release.yml` runs only on `v*` tags), so nothing here has ever been exercised.

Fix: one apt step, byte-for-byte the line `ci.yml` already uses.

## History of this PR (two commits)

It originally also renamed the Windows tester tag channel `-win.N` → `-winrc.N`, so its suffix would carry `rc` and be caught by Homebrew's pre-release keyword filter — stopping an unreviewed tester tag from auto-bumping homebrew-core.

**That was reverted after a fact-check.** Homebrew's documented mechanism for excluding pre-release tags is a `livecheck` regex on the *formula*, which excludes every suffix rather than only keyword-matching ones. The keyword list is Homebrew's internal fallback, not a naming contract for upstreams, so a rename here rested on an implementation detail and left the next channel name (or a typo) able to reopen the gap. The correct fix already existed in **#731** (the livecheck regex). Commit `619043e4` reverts the rename; the pre-release/autobump concern is left entirely to #731.

So this PR now closes exactly one gap: the missing ALSA headers.

## Verification

`actionlint` 0 · `just preflight` 0 (pre-push hook).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -49,17 +49,8 @@ jobs:
       - uses: extractions/setup-just@v4
       - name: Validate tag commit is on main
         # A release tag pushed at an arbitrary (unreviewed) commit must not
-        # build + publish (#215). Pre-release tester tags (v*-winrc.N / -rc.N)
+        # build + publish (#215). Pre-release tester tags (v*-win.N / -rc.N)
         # are deliberately exempt — they exist to test commits BEFORE merge.
-        #
-        # The suffix MUST contain one of Homebrew's UNSTABLE_VERSION_KEYWORDS
-        # (alpha beta bpo dev experimental prerelease preview rc — see
-        # Library/Homebrew/livecheck/livecheck.rb). homebrew-core autobumps us
-        # off `git ls-remote --tags`, and anything the keyword filter misses
-        # parses as a NEWER version: the old `-win.N` name contained no keyword,
-        # so a tester tag on an UNREVIEWED commit could have been auto-bumped
-        # into homebrew-core and bottled on six platforms. `-rc` was safe only
-        # by accident of its spelling. Keep `rc` in any new channel name (#731).
         if: ${{ !contains(github.ref_name, '-') }}
         run: |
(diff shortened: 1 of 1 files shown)


IvanWng97:
## Disposition of the earlier MEDIUM (the comment-only `-winrc` fix)

The review flagged that renaming `-win.N` → `-winrc.N` was comment-only — the actual gate (`if: !contains(github.ref_name, '-')`) never checks that a dash-suffixed tag's *suffix* contains a Homebrew keyword, so a future typo or new channel name could reopen the gap.

**The finding was correct, and it's now moot: the entire `-winrc` change has been reverted** (`619043e4`).

A fact-check against Homebrew's docs showed the rename was the wrong layer entirely. The documented way to keep pre-release tags out of autobump is a `livecheck` regex on the **formula** (`/^v?(\d+(?:\.\d+)+)$/i`), which excludes *every* suffix — not just keyword-matching ones — and doesn't depend on our tag names at all. That's tracked in **#731** and is where the whole concern now lives. Renaming our channel was an invented workaround resting on Homebrew's internal keyword list; reverting it is strictly better than adding the machine-check the finding suggested, because the machine-check would still be defending the wrong boundary.

So: **finding → obsoleted by revert; root concern → #731.** This PR is now ALSA-only.

PR #746, opened 2026-07-22
fix(audio): join the device thread on quit so playback stops before exit

IvanWng97:
## Symptom

Quitting pixtuoid with sound on left the music playing in the background; the user had to `sudo killall coreaudiod` to recover. **Reproduced and confirmed fixed on macOS** by the reporter (unmute → `q` → audio now stops immediately).

## Root cause (read from the code, not guessed)

The audio device thread was **detached**. `AudioHandle::respawn_in_place` spawned `run_loop` and discarded the `JoinHandle`:

```rust
Ok(_) => *self.tx.lock()... = Some(tx),   // ← JoinHandle dropped → thread detached
```

Teardown relied entirely on Drop order: on quit the handle drops → the shared sender drops → `run_loop` returns on `RecvError::Disconnected` → its `RodioSink` drops → the OS output device closes. But **nothing waited for that**. `run_tui` broke its loop and returned; `main` returned right after; the process exit **raced** the detached thread's teardown and usually won. On macOS a half-closed CoreAudio output strands playback — hence the `killall`.

## Fix — match the reference lofi TUI's synchronous stop

[lowfi](https://github.com/talwat/lowfi/blob/main/src/player.rs) (the lofi TUI this audio work was modelled on — its issues are cited in the audio code) stops its sink **synchronously** on quit: `close()` → `sink.stop()` before exit. rodio's own docs confirm "Dropping the Sink stops all sounds." This PR is that principle, adapted to pixtuoid's off-thread device:

- Keep the `JoinHandle` (new shared `join` cell on `AudioHandle`, alongside `tx`).
- `AudioHandle::shutdown()` — close the channel, then **join** the thread so the `RodioSink` Drop completes before the process exits. Bounded (2s) so a pathological rodio/cpal device-close can't hang quit (no worse than today's always-detached behaviour).
- Called once from each painter's exit: `run_tui` after the loop (covers `q` / Ctrl-C / external terminate / error), `floating` after `run_app` returns (covers close + the window-creation error exits).
- A re-spawn (`+` retry) now joins the prior thread instead of orphaning it.

## Regression test — deterministically red-capable

`shutdown_joins_the_device_thread_...` is device-free (a fake device thread mirroring `run_loop`'s `Disconnected => return`). Its teardown takes a measurable 300ms, so an **un-joined** `shutdown()` provably returns mid-teardown and the assert fails. Verified by removing the join and watching it go red, then restoring. A zero-cost teardown would let the detached thread win the race by luck and pass even unfixed — that false-green is exactly what the delay removes.

## What is NOT verified here

The end-to-end macOS coreaudiod symptom has no automatable seam (needs a real device + interactive quit). The test pins the fix **mechanism** (join → teardown completes before return); the **symptom** was confirmed by the reporter on hardware.

## Verification

`clippy -p pixtuoid --all-targets` 0 warnings · `cargo test -p pixtuoid --lib` **828 passed / 0 failed** · regression test red-without-fix / green-with-fix · `just preflight` 0 (pre-push hook).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2943,6 +2943,7 @@ dependencies = [
  "dasp_sample",
  "num-rational",
  "thiserror 2.0.18",
+ "tracing",
 ]
 
 [[package]]
--- a/crates/pixtuoid/CLAUDE.md
+++ b/crates/pixtuoid/CLAUDE.md
@@ -266,7 +266,15 @@ src/
 │                       knob; `m` = the whole opt-in, persisted via save_audio_muted) + volume clamped [0,1];
 │                       the system LAZY-SPAWNS on the first unmute (muted = zero cost: no device/thread/
 │                       buffers) — run_tui swaps the fresh handle into the renderer; floating boot-spawns
-│                       iff !muted AND has the SAME m/+/- runtime keys. The mute/volume TRANSITION is
+│                       iff !muted AND has the SAME m/+/- runtime keys. TEARDOWN-ON-QUIT (the "music keeps
+│                       playing after quit / killall coreaudiod" bug): the device runs on a spawned thread
+│                       whose RodioSink Drop closes the OS output — detached, that Drop RACES process exit
+│                       and on macOS strands CoreAudio. So `AudioHandle::shutdown()` (drop the sole sender →
+│                       run_loop returns → JOIN, bounded by SHUTDOWN_JOIN_TIMEOUT which must exceed a release
+│                       synth build, else it falls back to detached) runs at EVERY exit: run_tui after its
+│                       loop + a driver.rs backstop for run_tui's pre-loop `?`; floating after run_app (with
+│                       the boot-spawn placed AFTER the fallible `?` steps). Idempotent; disabled/muted =
+│                       no-op. The mute/volume TRANSITION is
 │                       ONE authority — `audio::apply_audio_action(&mut AudioUi, action, paused, spawn)`
 │                       (audio/mod.rs, unit-tested); the PERSIST protocol around it (mute-save-now, volume
 │                       debounce→flash-expiry, exit-flush) is a SECOND authority BOTH painters OWN —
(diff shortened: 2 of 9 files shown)


IvanWng97:
## Three-lens review — disposition

Lenses: correctness/grounding · concurrency/lifecycle (fired on the thread-teardown seam) · test-teeth/design. **All three: APPROVE-WITH-NITS, no REQUEST-CHANGES.** Fixes folded in `e6ef3aeb`.

### FIXED

| Finding | Lens | What was wrong |
|---|---|---|
| **F1** `SHUTDOWN_JOIN_TIMEOUT` 2s < the synth window, and its comment claimed "50ms / only pathological hangs" | concurrency (MEDIUM) | `run_loop` is blind to the closed channel during a ~2s (release) `AssetBank`/`TrackBeds` build; a quit then forced the thread past the build (~4s worst case) → 2s join timed out → **fell back to detached (the bug)**. Bumped to 8s + honest comment (incl. the respawn-caller nuance) |
| **F2** boot-spawn preceded the fallible `?` steps; floating comment over-claimed "every exit" | correctness + concurrency | floating: moved boot-spawn AFTER pack/runtime/event-loop build. TUI: spawn is in `driver.rs`, `run_tui`'s pack-load `?` precedes its in-loop shutdown → added a driver backstop (`shutdown()` on a shared clone after `run_tui` returns; idempotent) |
| **§2** test could be stronger | design | Extended the real-`run_loop` test with a device Drop flag — now pins that `run_loop` DROPS its device on disconnect (RodioSink→OS close) end-to-end, guarding a future leak-refactor. Reuses the existing test's synth cost. The device-free shutdown-joins test stays |
| **F3 / §5** invariant + docs | concurrency + design | Call-ordering invariant note on `shutdown()`; teardown-on-quit invariant added to `crates/pixtuoid/CLAUDE.md` (docs-currency) |

### Verified clean (not flagged)
Channel-close crux (sole sender, no clones), no deadlock (no lock held across the join), `join_with_timeout` panic-free, idempotent shutdown, winit `run_app` returns on desktop. Lens 1 confirmed the regression test is **red-capable by mutation** (removed the join → exit 101). F4 (respawn-join dead branch — kept as future-proofing) and F5 (panic path — accepted, pre-existing) noted, no action.

### ISSUE-FILED
**#752** — unify the two exit calls into `AudioController::close()` and fix the pre-existing Ctrl-C-loses-volume-flush asymmetry (design lens §3). A persist-protocol concern, deliberately out of scope for this device-teardown bug.

### Gates at `e6ef3aeb`
`cargo clippy -p pixtuoid --all-targets` **0** · `cargo test -p pixtuoid --lib` **828 passed / 0 failed** · both key tests green · `just preflight` **0** (pre-push).

PR #794, opened 2026-07-27
fix(site): drive the hero reveal roll by painted frames, not the clock

IvanWng97:
## The bug

On a first visit the hero office reveal froze on a half-drawn frame for ~1.4s, then snapped straight to the settled office — the roll was never seen.

## Root cause

**Safari's own tab-snapshot IPC blocks the page's main thread.** Captured with macOS `sample` attached to Safari 27's real processes, on two independent stalls (1307ms, 1454ms):

```
web process   Messages::WebPage::TakeSnapshot          <- Safari asks for the tab thumbnail
              -> WebPage::takeSnapshot
                -> WebImage::createImageBufferBackendHandle
                  -> RemoteImageBufferProxy::flushDrawingContext()
                    -> IPC::Semaphore::waitFor(IPC::Timeout)
                      -> semaphore_timedwait_trap      <- blocked on the GPU process

GPU process   RemoteImageBufferSet::endPrepareForDisplay
              -> CA::CG::ContextDelegate::operation_
                -> _dispatch_sync_f_slow -> __DISPATCH_WAIT_FOR_QUEUE__ -> kevent_id
```

Neither side is computing: the GPU process's main thread is 2933/2935 samples idle in `mach_msg2_trap` and all 28 `RemoteRenderingBackend` threads are parked. It is a CoreAnimation queue-ownership wait — which is why the duration is near-constant rather than proportional to page weight, and why **no JS callback in the profile exceeded ~4ms**.

It lands on the reveal because it co-occurs with the first live office frame — the same moment the boot splash's Level-2 `__pixEngineReady` gate lifts on. (Co-occurrence, not a proven common trigger: a profile shows timing, not WebKit's snapshot heuristic.)

## Why this took several wrong turns first

The stall is **intermittent** (Safari does not snapshot every load) and does not reproduce in Playwright's WebKit or Chromium (no tab UI to snapshot). Single-shot A/B runs therefore produced clean-looking separations that were pure noise. Two hypotheses were raised and then **refuted by controls**:

- **the #705 audio prewarm worker** — refuted: an arm where `new Worker` was never constructed still stalled 1452ms.
- **the roll's per-pixel alpha / canvas compositing** — refuted: with both variants served from the SAME url (path and navigation order had been confounded with the variant), base stalled 6/7 and the "fix" 7/10.

The profiler settled it. Lesson recorded in the sharp edge: pin changes here with repeated interleaved trials, never one sample per arm.

## The fix

The stall is Safari's and cannot be prevented. What it must not do is **eat the reveal**: `rt = (nowMs - revealStartSim) / REVEAL_MS` was a wall-clock ramp, so it kept advancing while nothing painted.

The roll now accumulates from **painted frames**:

- `reveal.elapsed += Math.max(0, Math.min(step, REVEAL_MAX_STEP_MS))` — a hitch contributes one frame of progress, so the roll pauses through it and resumes rather than being skipped.
- the roll holds until `REVEAL_READY_FRAMES` consecutive on-budget frames land, so a stall sits on the flat bg-tone cover (reads as a beat) instead of a half-drawn frame.
- `REVEAL_READY_MAX_WAIT` bounds that hold — a device that never meets the budget must still get its office, so the gate is a courtesy, never a precondition.

The code is **cause-agnostic**: it depends on no WebKit internal, only on "a frame took too long". The internals appear in the sharp edge to explain why, not as a contract.

## Verification

**Real Safari, 24 alternating loads, fixed vs pre-fix, both served from the same url.** With a ~1.4s stall inside the roll:

| variant | stall in roll | roll wall-clock |
|---|---|---|
| pre-fix | 1455 ms | **1644 ms** — unchanged, so the animation was eaten |
| fixed | 1429 ms | **2971 ms** = 1590 + 1429 — paused, then played in full |

Unstalled rolls unchanged (fixed 1590-1621ms vs pre-fix 1619-1646ms). The readiness gate holds 126-165ms across all 12 fixed trials — 3-4 frames, never the fallback.

**Tests:** three pins in `smoke.spec.ts` (stall-survival, slow-device fallback, un-reduce replay), each red before its fix. Since the real stall cannot be reproduced headlessly, they inject a synthetic main-thread block instead. Negative-control caveat and final gate results are in the review section below.

## Docs

Sharp edge added to `site/CLAUDE.md`, including the intermittency warning and the "frame-driven, never clock-driven" invariant.

---

## Local review (two-lens gate)

Findings driven to terminal states:

| # | Finding | State |
|---|---|---|
| F1 | **Regression this PR introduced.** `revealElapsed` was never reset, so the reduced-motion un-reduce path (which expects a fresh roll) left `rt >= 1` and the office SNAPPED — the very failure this PR fixes, on the neighbouring path. | **FIXED** (`1c70eca9`) + new e2e pin |
| F2 | `REVEAL_MAX_STEP_MS` and `FRAME_MS` were independent literals; raising `FRAME_MS` to meet it would silently disable the readiness gate, green, untestable. | **FIXED** — derived, value byte-identical |
| F10 | `Date.now()` is not monotonic; a backward jump rewound the roll. | **FIXED** — clamped at 0 |
| F4 | The reduced-motion comment was detached from the `if` it explains. | **FIXED** |
| F3 | The sharp edge was inserted mid-bullet, orphaning `public/wasm/`'s "Never hand-edit … regenerate from the crate". | **FIXED** — own top-level bullet |
| F5 | Four doc overclaims (common-trigger vs co-occurrence; an absolute from one profile; "the three" with no antecedent; "broken frame"). | **FIXED** |
| F9 | `pixtuoid-scene`'s `MAX_DT_S` settled the same question for audio. | **REFUTED as shared code** (different clock, unit, side of the wasm boundary) — cited in the comment instead |
| F8 | This is the second bounded hold on a first visit; it stacks with `Base.astro`'s `MAX_ENGINE_WAITS`. | **FIXED** — documented |
| F7 | The 1 Hz hidden-tab audio tick is permanently off-budget. | **REFUTED** — self-healing on return, and the roll surviving the hide is the intended improvement |
| — | Bundling the four reveal fields into one object. | **FIXED** — F1 is the evidence they share a lifecycle |

### Review integrity, stated plainly

- The **rendered-runtime / motion lens has now run** (it initially died on a session limit) — see the PR comment below: the roll was watched frame-by-frame, clean and under two verified stall placements, 6/6 loads settled, zero console errors. **APPROVE.** Its first pass reported a false clean because an injected stall silently never fired; the run now self-verifies that the stall executed.
- Two lenses initially returned stubs; the design lens was re-run as a focused agent and is what caught F1. The correctness lens never returned — its checklist items were covered by the design lens and by my own verification, but it is not an independent second opinion.
- My earlier claim that the new pins were "negative-controlled" was **wrong**: mutating `dist/index.html` breaks the inline script's CSP hash, so the browser blocked the whole script and the tests failed for that reason. Both pins have since been re-controlled by mutating the **source** and rebuilding, and both now fail on the intended assertion.

## Gates

`just site-check` EXIT=0 · `just site-e2e` **94/94** EXIT=0

## Follow-up (not this PR)

`pixtuoid-scene audio::synth::tests::day_take_stems_match_the_ratified_fingerprints` takes 101s alone against nextest's 180s cap and times out under parallel load. Latent flake; will file separately.

### DIFF
--- a/site/CLAUDE.md
+++ b/site/CLAUDE.md
@@ -173,6 +173,28 @@ display-line authority (`starText`), unit-tested on its null-stars arm since
   wrong-time still — the day/night flip is gone by construction; the splash in
   turn HOLDS on `window.__pixEngineReady` (Level-2 gate) so it lifts straight into
   the roll. Any failure / no-JS / no-wasm / reduced-motion keeps the still poster).
+  **SHARP EDGE — the reveal roll is FRAME-driven, never clock-driven.** Right
+  after a first visit settles, Safari blocks the page's ENTIRE main thread for
+  ~1.3-1.5s inside its own tab-snapshot IPC: `WebPage::TakeSnapshot` →
+  `RemoteImageBufferProxy::flushDrawingContext()` → `IPC::Semaphore::waitFor` →
+  `semaphore_timedwait_trap`, while the GPU process sits in
+  `CA::CG::ContextDelegate::operation_`'s `dispatch_sync` (captured with `sample`
+  on both processes, Safari 27). NEITHER side is computing — it is a
+  CoreAnimation queue-ownership wait, which is why the duration is near-constant
+  and why no JS callback ever exceeds 4ms. It fires on the same "page settled"
+  signal the splash's Level-2 `__pixEngineReady` gate waits for, so it lands ON
+  the roll whenever it fires. A wall-clock ramp
+  (`(nowMs - revealStartSim) / REVEAL_MS`) kept advancing while nothing painted,
+  so the roll froze on a half-drawn frame and SNAPPED to the settled office — the
+  reveal was never seen. So `paint()` accumulates `revealElapsed` from
+  `Math.min(step, REVEAL_MAX_STEP_MS)` per PAINTED frame, and holds the start
+  until `REVEAL_READY_FRAMES` consecutive on-budget frames land (putting the
+  stall on the flat bg tone instead of a broken frame) — bounded by
+  `REVEAL_READY_MAX_WAIT` so a device that never meets the budget still gets its
+  office. Don't "simplify" any of the three back to a clock. The stall itself is
+  Safari's and cannot be prevented; it does NOT reproduce in Playwright's WebKit
+  or Chromium (no tab UI to snapshot), and it is INTERMITTENT — a single-shot A/B
+  against it will lie, so pin changes here with repeated interleaved trials.
   Never hand-edit
   (prettier/eslint/knip all ignore it); regenerate from the crate.
   The hero renders at `BUF_H=130`, `SEED=0` (a deliberately closer camera than
--- a/site/CLAUDE.md
+++ b/site/CLAUDE.md
@@ -173,28 +173,6 @@ display-line authority (`starText`), unit-tested on its null-stars arm since
   wrong-time still — the day/night flip is gone by construction; the splash in
   turn HOLDS on `window.__pixEngineReady` (Level-2 gate) so it lifts straight into
   the roll. Any failure / no-JS / no-wasm / reduced-motion keeps the still poster).
-  **SHARP EDGE — the reveal roll is FRAME-driven, never clock-driven.** Right
-  after a first visit settles, Safari blocks the page's ENTIRE main thread for
-  ~1.3-1.5s inside its own tab-snapshot IPC: `WebPage::TakeSnapshot` →
-  `RemoteImageBufferProxy::flushDrawingContext()` → `IPC::Semaphore::waitFor` →
-  `semaphore_timedwait_trap`, while the GPU process sits in
-  `CA::CG::ContextDelegate::operation_`'s `dispatch_sync` (captured with `sample`
-  on both processes, Safari 27). NEITHER side is computing — it is a
-  CoreAnimation queue-ownership wait, which is why the duration is near-constant
-  and why no JS callback ever exceeds 4ms. It fires on the same "page settled"
-  signal the splash's Level-2 `__pixEngineReady` gate waits for, so it lands ON
-  the roll whenever it fires. A wall-clock ramp
-  (`(nowMs - revealStartSim) / REVEAL_MS`) kept advancing while nothing painted,
-  so the roll froze on a half-drawn frame and SNAPPED to the settled office — the
-  reveal was never seen. So `paint()` accumulates `revealElapsed` from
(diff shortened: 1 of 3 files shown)


IvanWng97:
Follow-up filed: #795 (`day_take_stems_match_the_ratified_fingerprints` runs 101s against nextest's 180s cap; times out under parallel load and takes the whole suite with it). Unrelated to this diff — it blocked this PR's first push attempt.

IvanWng97:
## Rendered-runtime / motion lens — now run

The earlier attempt at this lens died on a session limit, so the animation had not been watched. Re-run and completed; this closes the gap flagged in the PR body.

**Method.** Built the production site, drove it with Playwright, screenshotted `.backdrop` from `is-live` until `#office-overlay.is-on`, and read the frames as images. Stalls were injected from the driver *after* the office went live, with the page recording that the stall actually ran — an injection I can't verify fired is worthless.

**That check earned its keep immediately:** my first pass reported a clean result for the "stall 200ms into the roll" arm, but `stallActuallyRan` came back `null` — the stall had never fired and the run was meaningless. Re-done with verified injection.

### Roll timing

| run | stall fired | settled at |
|---|---|---|
| no stall | n/a | 1360 ms |
| stall @200 ms | **1400 ms** | 2767 ms ≈ 1360 + 1400 |
| stall @1200 ms | **1400 ms** | 2824 ms ≈ 1360 + 1400 |

Wall-clock extends by exactly the stall in both placements — the roll pauses through it and resumes. Clock-driven behaviour would have left it at ~REVEAL_MS regardless.

### What the frames actually show

- **Clean roll**: chroma fringe at 8 ms converging as the floors scroll, alpha ramping monotonically, captions on at settle. No stutter, no jump-back, no double-start, no flash, no frame where the office appears then re-hides.
- **During the stall** the frozen frame is a coherent early-reveal frame, and after it the roll has visibly *continued* (floors advanced, fringe converged) rather than snapping to settled.
- **Correction to the docs, confirmed by the frames**: that frozen frame is the faint chroma-split ghost of the scrolling floors, **not** the "flat bg tone" the first version of the sharp edge claimed. The review round had already corrected that wording on the design lens's argument; the capture confirms the corrected text is the accurate one.

### Reliability + state sweep

- 6 fresh first-visit loads all reached captions (1233–1386 ms). No load failed to settle.
- **Zero** console errors or unhandled rejections across every capture.
- Reduced-motion, no-JS, 390px mobile pan, `#office-pause` freeze/resume and tab-hidden/refocus are pinned by the existing suite (12 matching tests), green at 94/94.

**Verdict: APPROVE.**

PR #818, opened 2026-07-29
refactor(tui): one definition of the terminal-mode unwind order

### LINKED ISSUE #804: tech-debt: single-source the terminal-mode unwind order (tui::unwind_terminal_modes vs crash::restore_terminal)
## Problem

The terminal-mode unwind ORDER — `DisableMouseCapture` **while raw mode is still
on**, then `LeaveAlternateScreen`, then `disable_raw_mode` LAST — is
safety-relevant (getting it wrong strands the user's shell echo-less on Windows),
and it is currently restated at four sites in `crates/pixtuoid`:

- `src/tui/mod.rs` — the two `setup_terminal` rollback arms (`execute!(out,
  DisableMouseCapture, LeaveAlternateScreen); disable_raw_mode()`).
- `src/tui/mod.rs` — `teardown_terminal`, which carries the canonical WHY
  comment.
- `src/crash.rs` — `restore_terminal(&mut impl Write)` + the panic hook's
  following `disable_raw_mode()`.

Two in-flight audit branches each add a generic, writer-injected helper for the
same contract, neither able to see the other:

- `fix/audit-tui` adds `tui::unwind_terminal_modes<W: Write>(out, disable_raw)`
  (private `fn`, error-policy tested against a `FailingWriter`).
- `fix/audit-bin-runtime` adds `crash::restore_terminal<W: Write>(w)` (the
  stdout-vs-stderr fix, tested for both the writer seam and the end-to-end stream
  choice).

Once both land, the tree has two near-identical helpers for one ordering
contract, which is precisely the drift hazard each was extracted to remove.

## Correction to the obvious fix

The review that raised this proposed widening the tui helper to `pub(crate)` and
calling it from `crash.rs`. **That does not compile.** `crash.rs` is a
**bin-crate** module (`main.rs` declares `mod crash;`); `tui` lives in the
`pixtuoid` **lib** crate. `pub(crate)` on a lib item is invisible to the bin.
The reach already in `crash.rs` shows the required visibility:

```rust
// crash.rs:149 — "crash.rs is a BIN-crate module, hence the `pixtuoid::` path"
pixtuoid::tui::widgets::REPO_URL
```

So a single owner means a bare `pub fn` on the lib side. That is affordable —
the binary's lib target is explicitly not a semver surface (workspace
`CLAUDE.md`) and `widgets::REPO_URL` sets the precedent — but it is a deliberate
API widening, not the mechanical `pub(crate)` bump the finding described.

## Fix sketch (integration-time, after both branches merge)

1. Keep `tui::unwind_terminal_modes` as THE seam; make it `pub` with a doc
   comment naming the ordering invariant.
2. Route the two `setup_terminal` rollback arms through it, deleting their two
   restatements of the WHY.
3. Delete `crash::restore_terminal`; the panic hook becomes
   `let _ = pixtuoid::tui::unwind_terminal_modes(&mut std::io::stdout(),
   crossterm::terminal::disable_raw_mode);`, dropping the hook's now-redundant
   trailing `disable_raw_mode()`.
4. Keep `crash.rs`'s re-exec stream test (`the_hook_restores_on_stdout_and_keeps_
   the_report_on_stderr`) pointed at the hook — the *stream choice* is a crash-hook
   decision and must stay pinned where it is made, independent of who owns the
   sequence.

## Why not now

Both branches are unmerged and neither can reference the other's symbol; doing it
on either one guarantees a conflict in the same `tui/mod.rs` hunks and would land
a half-collapse. It is a single-commit cleanup once both are on `main`.

IvanWng97:
Closes #804.

## The problem

The terminal-mode unwind ORDER — `DisableMouseCapture` **while raw mode is still
on**, then `LeaveAlternateScreen`, then `disable_raw_mode` **last** — is
safety-relevant: getting it wrong strands the user's shell echo-less on Windows,
because `DisableMouseCapture` restores the input mode snapshotted at Enable time.

It was restated at four sites: `teardown_terminal`, both `setup_terminal`
rollback arms, and `crash.rs`'s own `restore_terminal` plus the hook's trailing
`disable_raw_mode`. Two in-flight audit branches each extracted a helper for the
same contract, neither able to see the other — so landing both produced exactly
the duplication each extraction was meant to remove. #804 deferred the collapse
until both were on `main`; they now are.

## The change

`tui::unwind_terminal_modes` becomes THE seam and goes `pub` — `crash.rs` is
declared in `main.rs`, so the call is genuinely lib→bin cross-crate (the
`REPO_URL` precedent). The two rollback arms route through it;
`crash::restore_terminal` is deleted and the hook hands the seam `stdout` plus
`disable_raw_mode` as its injected closure, dropping the now-redundant trailing
call.

`grep DisableMouseCapture` over production code now returns **one** ordering
statement.

## Coverage moved, not collapsed

The issue's fix sketch said "delete `crash::restore_terminal`". Following that
literally would have **lost** a property: `tui`'s tests pin the error policy (raw
disabled even when the escape write fails; first error wins) but **not** that the
sequence reaches the writer it was handed — only `crash.rs` pinned that.

So that test moves to `tui::teardown_tests`, beside the sequence it asserts on.
`crash.rs` keeps `the_hook_restores_on_stdout_and_keeps_the_report_on_stderr` —
the STREAM choice is the hook's own decision and is not observable from inside
the seam. The stdout-vs-stderr rationale moved to the hook's call site with the
function it documented, rather than vanishing with it.

## Gates

Exit codes read directly, never through a pipe:

- `cargo fmt --all --check` → **0**
- `cargo clippy --workspace --all-targets -- -D warnings` → **0**
- `cargo test -p pixtuoid` → **0** (16 test binaries)
- `just doc-check` → **0**

`crash::tests::the_hook_restores_on_stdout_and_keeps_the_report_on_stderr` passes
end to end through the shared seam — the real proof the consolidation is
behaviour-preserving. Docs swept: zero remaining references to the deleted fn.


IvanWng97:
Merge record @ `2a9909c1`.

**Online review lens: unsatisfiable.** Both bot jobs failed identically on all three heads of this branch (`f6a59b58`, `85a8026f`, `2a9909c1`) with `--json-schema was provided but Claude did not return structured_output. Result subtype: success`. The analysis job errors, so the publisher job is `SKIPPED` and no `Findings: N` comment can exist at HEAD. This is a reproducible harness fault, not budget exhaustion and not a finding — filing separately, since it silently disables the automated second lens repo-wide.

**Local two-lens review run in its place** (both APPROVE-WITH-NITS), findings dispositioned:

| # | Finding | Disposition |
|---|---|---|
| MEDIUM | unwind ORDER pinned by nothing — swapping `disable_raw()` above the escape write left the suite green | **fixed** — `raw_mode_is_disabled_only_after_the_escape_bytes_are_written`, negative-controlled (swap fails exactly one test) |
| F1 | CLAUDE.md said the order "was restated at each" of four exit paths; `teardown_terminal` always delegated (`origin/main:tui/mod.rs:620`) — three restated it | **fixed** |
| F4 | `tui/mod.rs:598` called `crash.rs` "a separate crate"; it is a module of the bin crate | **fixed** |
| LOW | stdout-vs-stderr trade rationale lost with `crash::restore_terminal`'s doc — the on-record refutation of a finding that will be re-raised | **fixed** (restored into CLAUDE.md) |
| F2 | `"; the setup error is what propagates"` — deducible from the two lines below (comment test 1) | **cut** |
| F3 | `"Unix-only for that reason."` — restates the `#[cfg(unix)]` below, entailed by the preceding clause (test 2) | **cut** |
| F5 | name the invariant in `tui/CLAUDE.md` | declined — `setup_terminal`/`teardown_terminal` are equally `pub` and unnamed there; silence matches precedent |
| F6 | narrow the `pub` surface so the test-only disabler is not cross-crate API | declined — adds a second name for one behavior; crate is not a semver surface |
| F7 | duplicate `LEAVE_ALT_SCREEN` const | declined — cannot be shared across the test-mod/bin-crate boundary; drift is self-detecting (both assert against real crossterm output) |

**Premise independently confirmed, not assumed:** a second sweep found terminal-mode code in exactly two files, `floating/` has zero (winit/softbuffer — never enters terminal modes), `git grep restore_terminal` returns nothing tree-wide, and `run_tui`'s four fallible `?` sites do not bypass the unwind (loop wrapped at `tui/mod.rs:753`, `teardown_terminal` unconditional at `:1224`).

Also surfaced: `cargo test -p pixtuoid --lib crash` runs **zero** tests while exiting 0 — `crash.rs` is in the bin target, so `--lib` cannot see it. Correct form is `--bin pixtuoid crash`.

Issue #819, opened 2026-07-29
ci(review): a spent Claude quota takes the review gate down silently — publisher skips, PR reads UNSTABLE, no signal at HEAD


IvanWng97:
Superseded the original root-cause analysis in this issue: **it was wrong.** Corrected below.

## What actually happens

The `claude review` / `claude security review` jobs fail with:

```
##[error]--json-schema was provided but Claude did not return structured_output. Result subtype: success
```

That message points at the schema. The schema is fine. The result payload in the same log says:

```json
{ "type": "result", "subtype": "success", "is_error": true,
  "duration_ms": 373, "num_turns": 1, "total_cost_usd": 0 }
```

373 ms, one turn, **zero cost** — the request was rejected before any inference. Compare a healthy run of the identical workflow ~45 min earlier:

```json
{ "type": "result", "subtype": "success", "is_error": false,
  "duration_ms": 269602, "num_turns": 24, "total_cost_usd": 1.0537598 }
```

So the cause is **a spent Claude quota**, not a schema or contract fault. `subtype` stays `"success"` while `is_error` flips to `true`; the action keys its error message off the missing `structured_output` and never surfaces the real reason. That mis-framing is what sent this investigation down the wrong path in the first place.

Timeline: eight runs succeeded at 2026-07-28 18:39 (a burst of audit PRs), everything from 18:40 onward failed, and it was still failing at 03:27 the next day. No `.github/` commit landed between the last success and the first failure.

## The defect that IS ours, and is worth fixing

Quota exhaustion is a recurring operating condition, not a bug. The problem is that the repo cannot tell the difference between "the second lens found nothing" and "the second lens never ran":

- `analyze` fails → `publish` is gated on `needs.analyze.outputs.claude_outcome == 'success'` (`claude-readonly-review.yml:172`) → `SKIPPED`.
- No `Findings: N` comment is ever written at HEAD.
- `ci-gate` is the only required check, so the PR sits at `UNSTABLE` — visually identical to a flaky advisory check.

The repo's documented merge gate is "the online review bot's `Findings: 0` at HEAD". When the quota is spent that gate is unsatisfiable, and nothing says so. PRs #805–#818 were merged during this window under a local two-lens substitute; that was the right call, but it depended on a human noticing the red job and reading its log — twice, in this case, getting the reason wrong.

## Proposed fix

[...]

PR #807, opened 2026-07-28
fix(runtime): floating pins a full CPU core, +9 defects

IvanWng97:
Also: the panic hook restored the terminal on stderr while stdout owned the alt screen (a redirected stderr stranded it), x11rb held at 0.13 to stop a Linux double-compile, compositor env markers read non-empty, and the wasm artifact regenerated for the weather fallback.

### Commits

- `8923a52d` style(audio): apply the redundancy-within comment test to the new WHY
- `c2a8d885` docs(floating): give the boot-seed residual the issue it claims (#803)
- `72c8bc71` docs: name gen-wasm as its own step, since `just gen` excludes it
- `a3a4a777` test(crash): apply the Windows rationale to BOTH restore tests, not one
- `619c3f34` build(deps): restore the 9 windows-sys edges the x11rb hold moved
- `59ed31c2` chore(wasm): regenerate the committed site artifact for the weather fallback
- `6006c9ea` fix(web): fall back to the clock when a forced weather name is unknown
- `cd541a8a` refactor(version): demote the version module to pub(crate)
- `47c3f3b9` docs(floating): stop claiming the boot seed and the first redraw agree
- `aa5c09a2` fix(crash): restore the terminal on stdout, the stream that entered the alt screen
- `2243ff00` fix(snapshot): report walkability for the floor seed actually rendered
- `93c2fcfd` build(deps): hold x11rb at 0.13 so it stops double-compiling on Linux
- `63d481d1` fix(focus): null the compositor child's stdio, and read its env marker non-empty
- `52b177a4` test(audio): block the run_loop barrier on the sink, not a 60s wall clock
- `4d2567b5` perf(floating): gate the redraw REQUEST on the tick deadline, not just the wait

---

### Provenance

Part of a whole-codebase audit (13 branches, 136 commits). Pipeline: 11 subsystem
finders + 8 whole-tree specialist sweeps + loop-until-dry security/concurrency
hunts → 312 raw findings → adversarial skeptics (default REFUTE, 3 differentiated
lenses on HIGH/security/concurrency) → **108 distinct confirmed defects** (35%
refutation rate, stable across two rounds) → fixes → a two-lens merge review per
branch plus trigger-driven escalation lenses → a fix round on that review's
findings.

Every defect has exactly one terminal state: **181 FIXED · 8 REFUTED-with-citation
· 12 ISSUE-NEEDED** (filed: #799 #800 #802 #803 #804).

### Composition verified

[...]

Issue #644, opened 2026-07-17
feat: mood multi-track lofi — music follows office time & weather (Lofi Girl reference)


IvanWng97:
Owner-ratified direction (2026-07-17): the office's lofi bed grows from ONE composition to a small MOOD LIBRARY, selected by the office's own context — the day/night clock and the weather the office already renders. Reference aesthetic: **Lofi Girl** (owner call; analysis-only per the repo's reference-fingerprint methodology — reference bytes never ship or get committed).

## Shape

- **Track registry**: each track = a frozen score table (the #642 freeze-the-realization machinery, unchanged) + composition params (progression, BPM, voicing set). All-procedural, zero assets.
- **First new track**: a NIGHT/RAINY mood — slower, softer, sparser than the current day track (which stays the default).
- **Selection policy**: office time-of-day + precipitation pick the track; crossfade on switch rides the existing mixer slew. Deterministic (no RNG at selection).
- **Pipeline per track** (the Phase 0/2 loop): measure Lofi Girl references (spectral + temporal + musical fingerprints: BPM, mode, progression flavor, instrumentation roles) → steer the python generator → owner LISTEN gate → freeze the realization → Rust port with fingerprint pins.

## Later on the same machinery

- More takes per mood (day-seed rotation), the grammar-reroll option (never-repeats melody within a ratified grammar), and the `--sound-dir` user packs (#633).

Part of the #633 family; this issue tracks the multi-track arc specifically.

Issue #632, opened 2026-07-16
feat: per-agent token/context meter — decode the usage the wire already carries + desk-visual pressure tiers


IvanWng97:
## What

Surface per-agent token/context usage in the office — the wire already carries it and we decode it to nothing.

**Wire → state:** new `AgentEvent::Usage { id, input_tokens, output_tokens, context_tokens? }`.
- Codex: `event_msg/token_count` — currently asserted-empty (`source/codex.rs` `session_meta_and_unknown_emit_nothing`); the payload arrives and is dropped.
- CC: transcript assistant messages carry `message.usage` (incl. cache-read fields — that's effective *context size*, the interesting number). The JSONL decoder never reads it today.
- omp: `tokensBefore` on compaction events = a direct context-pressure signal.
- Per the workspace refactor checklist: a new `AgentEvent` variant ⇒ update all four test areas (`tests/reducer/`, `tests/e2e.rs`, `tests/transport/socket.rs`, `tests/watcher/`) + `runtime/driver.rs` + an `agent_id()` arm.
- **Verify each wire shape against LIVE captures first** (the #500 wire-correctness rule — fixtures go stale; see `reference_wire-mining` practice).

**State → tiers:** `AgentSlot` accumulates counters; a `TokenPressure` enum mirroring the `burn.rs::BurnTier` precedent (e.g. `Light / Loaded / Compacting`), driven by explicit wire signals (compaction events, cache-read growth) — deliberately NOT a hardcoded model→context-window table, which drifts the day a provider bumps a window.

**Tiers → pixels (v1 scope):**
1. Dossier row: `◷ 12m · 34 calls · 128k ctx` (text, cheap, honest).
2. One desk visual: a paper stack on the desk grows with pressure tier; a brief "papers fly" flourish on a compaction event. Taste-half pins a mock render BEFORE coding; geometry-half gets RED tests first (TDD-for-visual-features).

## Why deferred

0.16.0 candidate (Arc A of the post-0.15.0 brainstorm). Sequenced after/alongside the ambient-audio discussion (Arc B); filed so the design isn't lost.

## Open decisions

- Raw token counts vs tiers-only in the dossier.
- Office-wide total on the wall board — proposed v2, keep v1 per-agent.
- Which other sources' wires carry usage (audit codewhale/cursor/copilot/opencode/hermes during wire verification).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #820, opened 2026-07-29
chore(release): 0.16.0 — curate the release notes to the full cycle scope

IvanWng97:
## What

Notes-only release prep for **0.16.0**. No version numbers move — `main` has been at `0.16.0` since #634 mid-cycle-bumped it.

## Why this isn't a `just bump`

`just bump 0.16.0` refuses by design (`error: 0.16.0 is not newer than the current 0.16.0`). The bump recipe drafts `release_notes()` **at bump time**, so a mid-cycle bump leaves the arm stale by construction — it only ever described the audio Phase 1 that triggered the bump, two bullets over what became a **139-commit** `v0.15.0..HEAD` cycle. Same shape as 0.14.0 (bumped mid-cycle at #532, re-curated at release time).

## The curated arm

Six highlights. App features only — site/web work stays out of the in-app popup (0.13.0 precedent).

1. **Sound** — `m` toggles procedural lofi that builds with the bustle, plus typing/rain/door/appliance cues; starts muted, `+`/`-` volume (#633/#642/#643/#651)
2. **Never repeats** — a fresh 10-minute track, slower and sparser at night or in rain (#644/#648/#704)
3. **Token meters** — paper tower per desk, `Σ` total on agent hover (#632/#646/#647)
4. **Two new agents** — Grok Build (`gk·`) and Kimi Code CLI (`km·`), 13 supported tools (#640/#692)
5. **Floating CPU** — no longer pins a core, 100% → ~13% with agents / ~0.5% idle (#807)
6. **Under the hood** — whole-codebase review sweep, one mascot per OpenClaw gateway, decoder drift-flagging (#787/#767/#768)

### Sizing (corrected)

An earlier revision of this PR body claimed the six bullets occupied "roughly the same ~19 wrapped rows" as the old two. **That was wrong and is retracted.** Measured against the real `wrap_notes` at the 48-char measure:

| | wrapped rows | chars |
|---|---|---|
| old arm (2 bullets) | 16 | 699 |
| first draft (6 bullets) | **19** | 702 |
| **current** | **16** | **671** |

The 19-row draft clipped 3 rows behind `⋮ 3 more` at the classic 80×24 — where the arm it replaced rendered completely — and guillotined the last bullet mid-sentence. The current arm is shorter than 0.15.0's and renders complete, verified by driving the real painter: `snapshot --popup --cols 80 --rows 24` shows all six bullets, no overflow marker, CTA intact.

## Gates

`just preflight` green (**exit 0**, 2799 passed / 2 skipped) · `notes-curated` ✓ · `npm-check` 23/23 · `gen-media --check` in sync · `gen-pix-icons --check` 17/17 · README drift in sync.

## Review

Four differentiated lenses (correctness, design, editorial, rendered-frame) — escalated past the two-lens floor because two triggers fire: *public-facing artifact (release notes)* and *a string a painter frames* (the #308 `⚠ ⚠` class). Findings and their dispositions are in the thread.

Note the online review bot cannot gate this PR: **#819** fails both Claude review jobs repo-wide (`--json-schema was provided but Claude did not return structured_output`), unrelated to this diff. The documented fallback — extra differentiated lenses + owner merge, recorded in the thread — is satisfied at 4 lenses.

## ⚠️ Pre-tag blocker

**#731 is still OPEN and fires on exactly this release.** The default-on `audio` feature (#633) landed after v0.15.0 and pulls rodio/cpal, which need ALSA on Linux. homebrew-core builds the workspace with **default** features on macOS *and* Linux; their formula declares no runtime deps.

Our artifacts are safe — `release.yml` builds Linux `--no-default-features` — but pushing the `v0.16.0` tag auto-triggers a homebrew-core bump whose from-source Linux build has no `depends_on "alsa-lib"`, breaking **their** CI on a bump we neither trigger nor get notified about. The fix lands in their formula alongside the version bump, not here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,7 +149,7 @@ config), see **[SECURITY.md](SECURITY.md)**.
 
 ## Contributing
 
-PRs welcome — especially new themes, sprite/decoration polish, and `Source` adapters for agent CLIs we don't support yet (the ten agent CLIs plus the OpenClaw gateway already wired up are in [Supported Tools](#supported-tools)). See **[CONTRIBUTING.md](docs/CONTRIBUTING.md)** for the build/test workflow, conventions, the review process, and how to add a new agent CLI. Architecture and the load-bearing invariants live in [`CLAUDE.md`](CLAUDE.md).
+PRs welcome — especially new themes, sprite/decoration polish, and `Source` adapters for agent CLIs we don't support yet (the twelve agent CLIs plus the OpenClaw gateway already wired up are in [Supported Tools](#supported-tools)). See **[CONTRIBUTING.md](docs/CONTRIBUTING.md)** for the build/test workflow, conventions, the review process, and how to add a new agent CLI. Architecture and the load-bearing invariants live in [`CLAUDE.md`](CLAUDE.md).
 
 ## Acknowledgments
 
--- a/crates/pixtuoid/src/version.rs
+++ b/crates/pixtuoid/src/version.rs
@@ -59,8 +59,12 @@ pub(crate) fn release_notes(version: &str) -> Option<&'static [&'static str]> {
         // brace would silently break if the indentation ever shifted.
         // [bump-inject-here]
         "0.16.0" => Some(&[
-            "The office you can hear — press m to turn on sound (starts muted, remembers your choice): a lofi band that layers up as your agents get busier (warm pad, sparkle, keys, drums — composed in-house, synthesized at startup, zero assets), typing that tracks the bustle, gentle rain when the office weather rains, and little moments — a door chime when someone walks in, the printer, the vending machine; you hear the floor you're looking at, and a footer ♩ shows when sound is live",
-            "Two moods, picked by the office itself — after dark (the same sundown the lights follow) or whenever it rains, the band switches to a slower night take: deeper bass, lazier drums, sparser keys; it crossfades back at sunrise",
+            "The office you can hear — press m for an all-procedural lofi score that layers up as agents get busier, plus typing, rain, and door/printer/vending cues. Starts muted; +/- sets volume",
+            "It never repeats — a fresh track composed every 10 minutes, in a slower night take after dark or when it rains",
+            "Token meters — each desk grows a paper tower as its session burns tokens, with a Σ row in the dossier",
+            "Two new agents — Grok Build (gk·) and Kimi Code (km·), bringing the roster to 13",
+            "Fits small terminals, and the floating window no longer pins a CPU core (100% → ~13%)",
+            "Sturdier under the hood — review sweeps, one lobster per OpenClaw gateway, and decoders that flag an upstream format change instead of guessing",
         ]),
         "0.15.0" => Some(&[
             "A busier, better-looking office — denser desk pods that pack the grid at every size, agents walking behind their desks, and a full interior-decor pass: glass walls with door jambs and mullions, a lounge aquarium, real head-of-table meeting chairs, greenery (two ficus trees) and mats",
--- a/crates/pixtuoid/src/version.rs
+++ b/crates/pixtuoid/src/version.rs
@@ -59,12 +59,12 @@ pub(crate) fn release_notes(version: &str) -> Option<&'static [&'static str]> {
         // brace would silently break if the indentation ever shifted.
         // [bump-inject-here]
         "0.16.0" => Some(&[
-            "The office you can hear — press m for an all-procedural lofi score that layers up as agents get busier, plus typing, rain, and door/printer/vending cues. Starts muted; +/- sets volume",
-            "It never repeats — a fresh track composed every 10 minutes, in a slower night take after dark or when it rains",
-            "Token meters — each desk grows a paper tower as its session burns tokens, with a Σ row in the dossier",
-            "Two new agents — Grok Build (gk·) and Kimi Code (km·), bringing the roster to 13",
-            "Fits small terminals, and the floating window no longer pins a CPU core (100% → ~13%)",
-            "Sturdier under the hood — review sweeps, one lobster per OpenClaw gateway, and decoders that flag an upstream format change instead of guessing",
+            "The office you can hear — press m for procedural lofi that builds with the bustle, plus typing, rain, door and appliance cues. Starts muted; + and - set volume",
+            "It never repeats — a fresh 10-minute track, slower and sparser at night or in rain",
+            "Token meters — each desk grows a paper tower as it burns tokens; hover an agent for the Σ total",
+            "Two new agents — Grok Build (gk·) and Kimi Code CLI (km·), for 13 supported tools",
+            "The floating window no longer pins a CPU core — 100% → ~13% with agents, ~0.5% idle",
+            "Sturdier under the hood — a whole-codebase review sweep (correctness, security, performance), a mascot per OpenClaw gateway, and decoders that flag upstream format changes",
         ]),
         "0.15.0" => Some(&[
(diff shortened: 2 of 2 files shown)


IvanWng97:

Reviewed at `d7d552aa`. **Four** differentiated lenses, escalated past the two-lens floor because two triggers in `pr-review.prompt.md` fire on this change: *public-facing artifact (release notes)* and *a string a PAINTER frames* (the #308 `⚠ ⚠` class, where a string-equality test passed two rounds and rendered wrong).

| Lens | Verdict |
|---|---|
| 1 — correctness / grounding | REQUEST-CHANGES |
| 2 — design / blast-radius | *(first run returned nothing — re-dispatched; see caveat)* |
| 3 — editorial (outside engineer) | REQUEST-CHANGES |
| 4 — rendered-frame | APPROVE-WITH-NITS |

Every MEDIUM+ premise was re-verified against source before any fix.

### FIXED

| # | Finding | Verification |
|---|---|---|
| 1 | **"Fits small terminals" is false** | `MIN_LAYOUT_W/H` (`layout/compute.rs:60-61`) byte-identical to v0.15.0; #806 routed overlays through all three paint paths so modals *reach* the footer-only frame, it did not move the office floor. An 80×24 render is still a blank office. Clause dropped. |
| 2 | **"a Σ row in the dossier"** wrong twice | `tooltip.rs:242-252` appends `" · Σ {n} tok"` to the `◷` stats line — its own comment calls it a fold, not a row. "dossier" is internal vocabulary no shipped arm uses (0.13.0 said "hover tooltips"). Now "hover an agent for the Σ total". |
| 3 | "Kimi Code" → **"Kimi Code CLI"** | `site/src/sources.json:186` |
| 4 | "roster to 13" read as 13 *agents* | registry is 12 agent CLIs + 1 daemon (`registry.rs:604` is the only `Kind::Daemon`) → "13 supported tools" |
| 5 | **19 rows clipped at 80×24** | measured 16 (old) vs 19 (draft); rewritten to 16 / 671 chars and confirmed by rendering `snapshot --popup --cols 80 --rows 24` — six bullets, no `⋮` marker, CTA intact |
| 6 | dropped the sweep-axes convention | 0.13.0 / 0.14.0 / 0.15.0 all carry `(correctness, …)`; restored |
| 7 | "lobster" has no antecedent | only the current version's arm ever renders (`renderer.rs:518`), so an upgrader never saw 0.9.0 introduce it → "mascot" |
| 8 | `+/-` register | every shipped arm spells keys out → "+ and -" |
| 9 | `README.md:152` "the ten agent CLIs" | stale by two, and contradicted the bullet this release ships → "twelve" |

### REFUTED

**A single-column guard for `Σ` / `·` / `→`.** All three are UAX#11 *Ambiguous* — but so is the `—` already shipped in every prior arm, so no new width class enters with this diff. The footer's `footer_vocabulary_is_single_column_…` guard exists for reasons that don't hold here: `pixtuoid_scene::footer` measures without the `unicode-width` dep and right-flushes, so a wide glyph shifts a pad and reddens a committed pixel baseline. The popup is left-flushed, `Paragraph`-clipped, and renders into no committed baseline. Adding a guard would be defense-in-depth against a failure mode with no blast radius.

Independently confirmed by the rendered lens measuring each glyph through `widgets::display_width` (not a tautology — that's `unicode_width`): `Σ`/`·`/`→`/`—` all `chars=1, display_width=1`.

### ISSUE-FILED

**#821** — no `media.json` job drives `snapshot --popup`, so no committed baseline guards this surface. That's why the clipping above needed a human to catch. Filed rather than fixed here: the arm's prose changes every release, so a committed PNG would churn each cycle — the issue records that a harness assertion on wrapped-row count may be the better shape.

### plan-misses recorded in the commit

- The PR body claimed the six bullets occupied "roughly the same ~19 rows" as the old two. Measured: old is **16**, draft was **19**. The sentence offered as evidence the change was safe was itself unverified. PR body corrected.
- No mechanical gate covers this surface (→ #821).

### Caveats, stated rather than buried

- **Lens 2 (design) returned no report** across two idle cycles — a stub, not a review. Re-dispatched with a focused brief; it had not returned when this sweep was posted. Its scope was curation (which bullet earns its slot), not correctness, and the remaining three lenses covered accuracy, framing and voice. The one call it would have arbitrated: the under-the-hood bullet had room for the decoder-drift clause **or** a small-terminal clause, not both at 16 rows — decoder-drift was kept because it protects all 13 sources silently, where the panel fix only bites below 31 rows. Reversible if that lens argues otherwise.
- **The online review bot cannot gate this PR** — #819 fails both Claude review jobs repo-wide (`--json-schema was provided but Claude did not return structured_output`), 0 successes across 7 branches / 11 runs. Unrelated to this diff, which is 6 lines. The documented fallback (extra differentiated lenses + owner merge, recorded in the thread) is satisfied at 4 lenses.

### Coverage note

No test reds if any bullet is factually false — `release_notes_present_for_every_shipped_version` asserts only non-emptiness, and its own doc rejects prose assertions as brittle. That is the right shape for copy, but it means human review is the only gate between a false claim and every upgrader's first-launch popup. Which is precisely how a wrong claim reached the first draft of this PR.

Gates at `d7d552aa`: `just preflight` **exit 0**, 2799 passed / 2 skipped.

IvanWng97:
## Addendum — final head `76e28e25`

### Clause swap (owner decision)

The under-the-hood bullet holds **two** of three clauses inside the 16 wrapped rows the 80×24 notes band shows — all three measure 5 rows and clip. The first round kept mascot-per-gateway + decoder-drift, but that choice was made for a cleanliness reason (the small-terminal item was replacing the false claim, so dropping it was tidiest) rather than by ranking user impact.

Ranked by who is actually affected:

| Clause | Population |
|---|---|
| drift-flagging decoders | all 13 registered sources, invisibly |
| panels fit a small terminal (#806) | anyone under 31 rows — a tmux split, an 80×24 default |
| a mascot per OpenClaw gateway (#787) | needs OpenClaw **and** >1 gateway running |

The original pick kept the narrowest of the three. Owner selected the two wider ones. Re-verified by rendering rather than arithmetic: `snapshot --popup --cols 80 --rows 24` paints all six bullets, no overflow marker, CTA intact. `just preflight` exit 0, 2799 passed.

Consequence to note: **#787's multi-gateway mascot now goes unmentioned in the notes.** Deliberate, not an oversight.

### Correction to the earlier sweep — the design lens never ran

My previous comment listed four lenses. That overstates what happened, so I am correcting it rather than leaving it to imply four clean reviews.

- **Lens 1 (correctness), Lens 3 (editorial), Lens 4 (rendered-frame)** — genuine independent reviews, full reports, between them nine real defects.
- **Lens 2 (design)** — dispatched twice (a second time with a tighter brief) and **returned nothing both times**. Per the protocol a dispatch is not a lens, so it was never counted; but the re-run failed the same way.

The design dimension was therefore covered by **me, the change's author** — structurally weaker than an independent lens, and exactly the correlated-blind-spot the two-lens rule exists to prevent. What that self-pass concluded, recorded so a reader can audit it rather than trust it:

- **Ranking:** sound > never-repeats > token meters > floating CPU > new agents > under-the-hood. Nothing weak enough to cut at the ceiling.
- **Missing work:** **#698** (the floating window gained the terminal's status footer) is the most notable omission and is unmentioned. Judged below all six; a deliberate cut.
- **Version-lockstep:** staying at 0.16.0 is correct — the cycle shipped features, but #634 already made that minor bump mid-cycle. A second bump here would be the over-bump #471 warns against.
- **DRY/drift:** the arm restates facts owned elsewhere (roster count, CPU figures, the 600s epoch), but a versioned arm never renders again after its release, so it does not rot like a live doc. The one that *had* rotted — `README.md:152` — is fixed in this PR.
- **Voice:** the arm still omits the closing "the office looks the same, just …" reassurance that four consecutive arms carry. Left out deliberately on length grounds.

**What a reviewer should weigh:** the three dimensions with independent lenses are well covered; the design dimension has only an author self-review. If the design read matters to you here, that is the gap to close before merge.

PR #815, opened 2026-07-28
fix(scripts): squattable e2e sockets and a PATH-resolved replay binary

IvanWng97:
Fixes 5 confirmed audit defects in the e2e and replay tooling.

Two e2e scripts bound the hook socket at a **fixed, predictable name in the shared temp dir** — two concurrent runs destroyed each other, and the path was squattable. `replay-fixture.sh` defaulted to whatever `pixtuoid` was on PATH, so the documented dogfood step could green-light a stale release binary. The multi-gateway e2e printed "PASS sibling N is untouched" for gateways that had never appeared.

**Four commits from this branch were dropped, not merged.** #796 landed mid-audit and solved the same CodeWhale facade-split and unanchored-sweep problems *more completely* — 17 anchored call sites plus `test_every_swept_url_declares_an_anchor`, a mechanical guard this branch never had. The audit finding its own work redundant is the correct outcome.

### Commits

- `95cea2dd` style(scripts): apply the redundancy-within comment test to the e2e WHY
- `b54e0976` fix(scripts): stop the cc-backend e2e asserting an upstream fact its sibling measured false
- `432e21a1` fix(scripts): make the multi-gateway e2e's step [1] a hard precondition
- `dac2fadc` fix(scripts): replay the WORKING TREE's binary, not whatever `pixtuoid` is on PATH
- `56e9a7a2` fix(scripts): resolve the cc-backend e2e's gateway port instead of pinning 18789
- `35c4ad3e` fix(scripts): put the e2e hook socket in a private dir, not a fixed shared-tmp name

---

### Provenance

Part of a whole-codebase audit (13 branches, 136 commits). Pipeline: 11 subsystem
finders + 8 whole-tree specialist sweeps + loop-until-dry security/concurrency
hunts → 312 raw findings → adversarial skeptics (default REFUTE, 3 differentiated
lenses on HIGH/security/concurrency) → **108 distinct confirmed defects** (35%
refutation rate, stable across two rounds) → fixes → a two-lens merge review per
branch plus trigger-driven escalation lenses → a fix round on that review's
findings.

Every defect has exactly one terminal state: **181 FIXED · 8 REFUTED-with-citation
· 12 ISSUE-NEEDED** (filed: #799 #800 #802 #803 #804).

### Composition verified

All 13 branches were merged together in a throwaway worktree and gated as one tree
— the step no per-branch review can perform:

- `just preflight` → **exit 0** (13 lint sub-gates → clippy → hack → **2,795 tests passed**)
[...]

PR #809, opened 2026-07-28
fix(install): complete the terminal-egress and 0600 sibling sets, +10 defects

IvanWng97:
Fixes 12 confirmed audit defects on the highest-risk surface in the repo — the code that mutates **other CLIs' user-owned config files**, where PR #23 shipped a path traversal.

Terminal-egress sanitization now strips **both Cc controls and Cf bidi overrides** (`is_control()` covers only Cc, and Trojan-Source CVE-2021-42574 rode exactly that gap), applied across the full sibling set rather than one site. The lock sidecar and both diagnostic sinks are created **0600 at creation time** — not chmod-after, which races. A PATHEXT shim is rejected as the resolved hook binary on Windows (untestable locally; relies on the `windows-test` CI job). `doctor` is genuinely read-only again and no longer reports "no decode drift" when the log is simply unreadable.

### Commits

- `c01d328d` style(install,logging,drift): apply the redundancy-within comment test to the new WHY
- `b6e5a08a` fix(docs): unlink a private item from a public doc and trim three fn-body comments
- `a5643635` fix(install,logging): give each half of the 0600 fix — and the PATHEXT guard's call site — its own test
- `67829966` fix(cli): close the terminal-egress sweep on the tracing sink, and pin the two bidi tables
- `302a12b6` fix(docs): drop two intra-doc links to private items
- `35d7d778` fix(core): make drift breadcrumb values display-safe where they are emitted
- `e39301c9` fix(install): reject a PATHEXT shim as the resolved hook binary on Windows
- `dd0ee1cc` fix(install,logging): create the lock sidecar and both diagnostic sinks 0600
- `2a8d00f7` fix(doctor): stop an unreadable log from reading as "no decode drift"
- `856aa2ca` fix(doctor): gate the <cli> --version probes on evidence the user runs that CLI
- `611ac432` fix(install): name PIXTUOID_HOOK, not the deleted --hook-path flag
- `4c6fd56a` fix(cli): strip control chars from the two unsanitized terminal-egress sites
- `63870d47` fix(cli): keep maintainer notes out of --help and derive --theme's list

---

### Provenance

Part of a whole-codebase audit (13 branches, 136 commits). Pipeline: 11 subsystem
finders + 8 whole-tree specialist sweeps + loop-until-dry security/concurrency
hunts → 312 raw findings → adversarial skeptics (default REFUTE, 3 differentiated
lenses on HIGH/security/concurrency) → **108 distinct confirmed defects** (35%
[...]

PR #822, opened 2026-07-29
fix(ci): close the fork-tree exposure and the silent review gate

### LINKED ISSUE #799: claude.yml: the issue_comment arm still stages a fork tree in the contents:write job (claude-code-action checks out the PR head itself)
## Problem

`.github/workflows/claude.yml` is the repository's only `contents: write` Claude
job. It holds `id-token: write`, `pull-requests: write`, `issues: write` and the
`CLAUDE_CODE_OAUTH_TOKEN`, and it checks out without a ref.

#794-era work (commit `d8c98b64`) added
`github.event.pull_request.head.repo.full_name == github.repository` to the two
`pull_request_review*` arms of the job's `if:`, because on those events
`GITHUB_REF` is `refs/pull/<n>/merge` and the workflow's own
`actions/checkout@v7` therefore lands fork-authored files — including a
repo-root `CLAUDE.md` the agent auto-loads as instructions — in that job's
workspace.

That guard is real, but it does not close the class. The **`issue_comment`** arm
reaches the same state through the action itself rather than through
`GITHUB_REF`, so no `if:` expression on this workflow can express it.

## Trace (read from the pinned `v1` tag of `anthropics/claude-code-action`)

- `src/github/context.ts` — for `issue_comment`: `isPR: Boolean(payload.issue.pull_request)`.
  A comment on a PR's conversation tab is an `issue_comment` with `isPR` true.
  (For `issues` it is hardcoded `false`, so that arm is genuinely not affected.)
- `src/modes/tag/index.ts` — tag mode (what this workflow selects by passing
  `track_progress: true` plus an `@claude` mention) calls `setupBranch(...)`
  unconditionally after `fetchGitHubData`.
- `src/github/operations/branch.ts` — inside `setupBranch`, for an OPEN PR:

  ```ts
  if (prData.isCrossRepository) {
    execGit(["fetch", "origin", `--depth=${fetchDepth}`, `pull/${entityNumber}/head:${branchName}`]);
  } else {
    execGit(["fetch", "origin", `--depth=${fetchDepth}`, branchName]);
  }
  execGit(["checkout", branchName, "--"]);
  ```

So a maintainer typing `@claude ...` in the conversation tab of a **fork** PR —
the most common way this workflow gets used on a PR — stages the fork's tree in
the workspace of the `contents: write` job, after our own checkout has run.

The action's `docs/security.md` warns against checking out an untrusted ref
*before* invoking the action; it does not claim the action avoids doing so
itself.

## Why this is filed rather than fixed

The `issue_comment` payload carries no `pull_request` object — only
`issue.pull_request` with url/html_url/diff_url/patch_url — so fork-ness is not
expressible in the workflow's `if:`. Closing it needs a first step in the job
that resolves the PR and fails when it is cross-repository, e.g.

```yaml
- name: Refuse fork pull requests
  if: github.event_name == 'issue_comment' && github.event.issue.pull_request
  env:
    GH_TOKEN: ${{ github.token }}
    PR_NUMBER: ${{ github.event.issue.number }}
  run: |
    cross="$(gh pr view "$PR_NUMBER" --json isCrossRepository --jq .isCrossRepository)"
    [[ "$cross" == "false" ]] || { echo "::error::@claude is not available on fork pull requests"; exit 1; }
```

That is a product decision, not a mechanical fix: it removes a legitimate and
frequently useful flow (a maintainer asking Claude for help on a contributor's
fork PR), and it needs a live trigger to verify — neither is verifiable from the
policy gate. Filing it so the residual exposure is tracked rather than implied
away by a comment.

Interim mitigation shipped alongside this issue: the workflow comment and
`CLAUDE.md` no longer assert that the `issue_comment` arm is unexposed.

## Notes

- `policy/ci-observability/main.rego` gates only what IS expressible: for each
  `pull_request_review*` event present in `on:`, every arm of the job's
  condition naming that event must carry the head check, and exactly one job in
  the workflow must run `anthropics/claude-code-action@v1`.
- Any fix here must keep those policy rules green.

### LINKED ISSUE #819: ci(review): a spent Claude quota takes the review gate down silently — publisher skips, PR reads UNSTABLE, no signal at HEAD

IvanWng97:
Closes #799. Closes #819.

## #799 — `@claude` staged fork trees in the `contents: write` job

The whole fix, unchanged since the first commit: refuse fork PRs before the action runs.

`issue_comment` reaches a fork tree by a route **no `if:` can express** — its payload carries no `pull_request` object, and claude-code-action stages the head itself. From the pinned `v1` source, `setupBranch` runs `git checkout` for every open PR; the fork/non-fork split only changes how the ref is fetched, there is no guard, and no input disables it (`allowed_non_write_users` governs who may trigger, not what is checked out).

So a maintainer typing `@claude` on a fork PR staged fork-authored files — a repo-root `CLAUDE.md` the agent auto-loads as instructions among them — in the one job holding `contents: write` and the OAuth token. The existing `author_association` gate does not help: the trusted maintainer IS the trigger in that scenario.

Costs nothing real: both other Claude paths already refuse forks, so fork PRs had no Claude coverage to lose.

Verified against live data, both directions:

| PR | head | result |
|---|---|---|
| #818 | `IvanWng97/pixtuoid` | allowed |
| #726 (a real fork PR) | `chrisfinazzo/pixtuoid` | **refused** |
| deleted fork | `null` → `""` | **refused** (fail-closed) |

## #819 — a spent quota took the review gate down silently

When the model job fails, `publish` is gated on its outcome and skips, no comment lands, and with `ci-gate` the only required check the PR reads `UNSTABLE` — identical to a flaky advisory. That is how #809 and #815–#818 merged with only a red job to say so.

**A spent quota is rare, so this is deliberately thin**: one comment saying the second lens is ABSENT rather than clean, and one policy rule pinning that the job exists and carries a status function — without one an implicit `success()` skips it in exactly the situation it exists for.

An earlier revision of this PR carried an upsert, a diagnostic step, a decline arm and five policy rules for the same rare condition. Cut: 560 → 336 lines. `publish` posts a fresh comment per head too; matching that is enough.

## Regression cover

Where the test weight sits is deliberate — on the security control, not the notice:

- the fork refusal must exist, be **scoped to `issue_comment`**, and **precede the action** (a refusal running after the tree is staged is decoration), identified by the API field it must read rather than its name, so a rename cannot retire it
- behavior tests execute the guard itself: an internal PR is admitted, a fork and a deleted head are refused; the notice says ABSENT and cannot be prefix-matched as a review

Every guard negative-controlled — inverting the fork comparison, softening the ABSENT wording, and dropping the status function each fail.

`just preflight` exit 0 · 2799 tests · 115 rego unit tests · 111 conftest · 0 violations. Reviewed over four local two-lens rounds; both lenses APPROVE.

https://claude.ai/code/session_012eeTnuzQZ2gSf8Mctswev7

### DIFF
--- a/.claude/skills/two-lens-review/SKILL.md
+++ b/.claude/skills/two-lens-review/SKILL.md
@@ -76,7 +76,7 @@ silently dropped).
    the check table alone can't gate (#448). If the bot ERRORED or left no
    findings comment at HEAD (it can fail on a very large diff — `error_max_turns`
    with no comment — or on a spent quota, which the workflow now states itself in
-   an `<!-- absent-<marker> -->` comment; do NOT read that as a review),
+   an `<!-- absent-<marker>:<sha> -->` comment; do NOT read that as a review),
    the gate is unsatisfiable as written: split the PR smaller,
    else fall back to one extra differentiated lens + owner merge, recorded in the
    PR thread. State the condition behaviorally (errored/absent), never a fixed
--- a/.claude/skills/two-lens-review/SKILL.md
+++ b/.claude/skills/two-lens-review/SKILL.md
@@ -75,7 +75,9 @@ silently dropped).
    + `mergeStateStatus` — the review JOB passes even when it posts findings, so
    the check table alone can't gate (#448). If the bot ERRORED or left no
    findings comment at HEAD (it can fail on a very large diff — `error_max_turns`
-   with no comment), the gate is unsatisfiable as written: split the PR smaller,
+   with no comment — or on a spent quota, which the workflow now states itself in
+   an `<!-- absent-<marker> -->` comment; do NOT read that as a review),
+   the gate is unsatisfiable as written: split the PR smaller,
    else fall back to one extra differentiated lens + owner merge, recorded in the
    PR thread. State the condition behaviorally (errored/absent), never a fixed
    LOC ceiling.
--- a/.github/prompts/pr-review.prompt.md
+++ b/.github/prompts/pr-review.prompt.md
@@ -528,8 +528,9 @@ review's LATEST COMMENT verdict (`Findings: N`) plus `mergeStateStatus`,
 never the check table: the claude-review JOB is green even when it posts
 findings (#448 merged past a fresh MEDIUM; #449 onward reads the comment
 verdict). A run that produced NO verdict now says so in its own comment,
-carrying an `<!-- absent-<marker> -->` marker — that is an unreviewed head,
-not a clean one, and it is not a `Findings: 0`.
+carrying an `<!-- absent-<marker>:<sha> -->` marker — that is an unreviewed
+head, not a clean one, and it is not a `Findings: 0`. It is keyed to the head,
+so a notice for an older sha is not a verdict on this one.
 
 ---
 
--- a/.github/prompts/pr-review.prompt.md
+++ b/.github/prompts/pr-review.prompt.md
@@ -527,7 +527,9 @@ gates and watch the NEW head's CI — and gate the merge on the online bot
 review's LATEST COMMENT verdict (`Findings: N`) plus `mergeStateStatus`,
 never the check table: the claude-review JOB is green even when it posts
 findings (#448 merged past a fresh MEDIUM; #449 onward reads the comment
-verdict).
+verdict). A run that produced NO verdict now says so in its own comment,
+carrying an `<!-- absent-<marker> -->` marker — that is an unreviewed head,
+not a clean one, and it is not a `Findings: 0`.
 
 ---
 
--- a/.github/workflows/claude-readonly-review.yml
+++ b/.github/workflows/claude-readonly-review.yml
@@ -152,37 +152,6 @@ jobs:
             --allowedTools "Read,Glob,Grep"
             --json-schema ${{ steps.schema.outputs.json }}
 
-      # The action reports a rejected request as `subtype: "success"` with a
-      # missing structured_output, so its error blames the schema no matter the
-      # cause. `is_error` and a $0 cost are what actually separate "no quota"
-      # from "bad contract"; both sit in the same object, unread. #819
-      - name: Explain a failed analysis
-        if: always() && steps.claude.outcome == 'failure'
-        env:
-          EXECUTION_FILE: ${{ steps.claude.outputs.execution_file }}
-        run: |
-          set -uo pipefail
-          if [[ -z "${EXECUTION_FILE:-}" || ! -f "$EXECUTION_FILE" ]]; then
-            echo "No execution file was propagated: the action failed before Claude produced one." >>"$GITHUB_STEP_SUMMARY"
-            exit 0
-          fi
-          result="$(jq -c '[.. | objects | select(.type? == "result")] | last // {}' "$EXECUTION_FILE" 2>/dev/null)"
-          if [[ -z "$result" || "$result" == "{}" ]]; then
-            echo "Execution file unreadable, or carrying no \`result\` object." >>"$GITHUB_STEP_SUMMARY"
-            exit 0
-          fi
-          {
-            echo "### Why the analysis failed"
-            echo
-            jq -r '"- is_error: `\(.is_error)`\n- num_turns: `\(.num_turns)`\n- duration_ms: `\(.duration_ms)`\n- total_cost_usd: `\(.total_cost_usd)`"' <<<"$result"
-            echo
-            if [[ "$(jq -r '.total_cost_usd // 1' <<<"$result")" == "0" ]]; then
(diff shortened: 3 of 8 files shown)


IvanWng97:
Merge record @ `e14b9ac5`.

**Online review lens: unsatisfiable.** Both bots failed with a spent quota — the exact condition #819 addresses, and it cannot verify itself: `pull_request_target` loads the workflow from the default branch, so `report_absence` does not run for this PR. Confirmed: zero absence comments here. It takes effect for the next PR that trips a failed or declined analysis, and that is where the behaviour needs watching rather than being assumed.

**Local two-lens review: four rounds**, both lenses APPROVE at the last one. Rounds 1–3 each found real defects, twice in fixes from the previous round. Every finding landed in the #819 half; `claude.yml` — the #799 security fix — changed 0 lines across all four.

**Then it was cut.** The owner pushed back that a rare condition did not warrant the machinery: 560 → 336 lines. Removed the upsert (and its stub URL capture, failing-listing knob, multi-id guard, four behavior cases), the 25-line diagnostic step, the exact-permission and no-checkout rules. Two pieces added back at ~48 lines for their value:

| kept | why |
|---|---|
| decline arm | `reviewable=false` exits 0, so analyze goes GREEN and nothing comments — the only absence shape with no red job behind it |
| one line naming `total_cost_usd` | the action blames the JSON schema whatever the cause; this issue was misdiagnosed twice, once in a filed issue. The reader needs the field name, not a reformatted log |

Final: 384 lines, of which the test weight sits on the security control (12 rego tests for #799, 3 for #819).

**Every guard negative-controlled**, each demonstrated to fail when the property it protects is broken:

```
fork comparison inverted        -> "fork guard rejected an internal pull request"
refusal moved after the action  -> denies
refusal rescoped off issue_comment -> denies
decline arm dropped             -> denies
status function dropped         -> denies (with its own message, not the existence one)
either branch text softened     -> fails
cost field renamed              -> fails
```

`just preflight` exit 0 · 2799 tests · 111 conftest · 0 violations.

Issue #247, opened 2026-06-11
drift-watch the undocumented sessions/<pid>.json registry fields (sessionId/startedAt/pid)


IvanWng97:
## Problem

The CC liveness probe's admission rung reads `<claude_home>/sessions/<pid>.json` (`sessionId`, `startedAt`, `pid`). The 2026-06-11 alignment research confirmed this registry is a fully UNDOCUMENTED internal: the official .claude-directory doc's table has no `sessions/` row and the CHANGELOG never mentions it. It is the one upstream dependency with NO drift detector — a silent rename (e.g. `startedAt` → `started_at`) would degrade the probe to mtime-only gating without any signal.

(Degradation is graceful by design — the probe is additive-only — but the mid-attach UX regression of #224 would quietly return for old-mtime live sessions.)

## Fix sketch

Extend `scripts/check_upstream_drift.py` with a third surface: fetch is impossible (no doc), so watch the LOCAL shape instead — a tiny conformance check in the weekly run or a unit test reading a committed fixture of the current shape + a runtime warn-once in `live_cc_session_ids` when a registry file parses but lacks the expected keys (today it just yields no vouch silently). The warn-once is the cheap high-value half.

PR #840, opened 2026-08-03
fix(grok): route active_sessions pids through checked_pid so one bad entry skips

### LINKED ISSUE #831: [Bug] grok's active_sessions.json pid ingress bypasses checked_pid — one out-of-range value fails the whole probe
## What

`decoder::checked_pid`'s doc declares itself:

> The ONE narrowing every JSON `_pid` ingress rides (the hook peek, the openclaw decode, AND the CC sessions-registry probe in `cc_probe::parse_registry_entry`), so a new ingress can't ship the N-th unchecked pid

Three call sites confirmed: `hook/mod.rs:309`, `openclaw.rs:161`, `cc_probe.rs:243`.

But there is a **fourth** JSON pid ingress: grok's `{grok_home}/active_sessions.json`, parsed in `grok/native.rs::grok_ids_from_registry`. It hand-rolls the equivalent guard inline instead:

```rust
struct Entry { session_id: String, pid: i32, opened_at: Option<String> }
...
if e.pid <= 0 || e.session_id.is_empty() || !alive(e.pid) { continue; }
```

## This is not a missing-validation bug

To be clear about severity: grok **does** reject non-positive pids, twice over (`e.pid <= 0`, plus `pid_alive`'s `Pid::from_raw` returning `None` for ≤ 0). There is no unchecked-pid hole. An earlier framing of this as a validation gap was investigated and **refuted**.

## The actual defect: differing failure granularity

The two ingresses handle an out-of-range pid **differently**, because grok deserializes straight into `i32` while `checked_pid` takes `i64` and narrows:

| ingress | a pid outside i32 range in the registry |
|---|---|
| CC (`checked_pid`, `i64` → narrow) | skips **that entry**, keeps the rest |
| grok (`serde` → `pid: i32`) | `from_slice::<Vec<Entry>>` fails → `.ok()?` → `None` → `shape_drift` warning → **all** grok liveness degrades to mtime |

So one malformed value takes down the whole probe rather than one row — and it reports as *format drift*, which is misleading: the shape is fine, one value is out of range.

Practical likelihood is low (Linux `pid_max` ≈ 4M), which is why this is `priority: low`.

## Why fix it anyway

1. It makes `checked_pid`'s doc **true**. Right now it claims to be "the ONE narrowing every JSON `_pid` ingress rides" and it is not — the hand-rolled copy is exactly the "second hand-copied literal" drift shape this repo keeps getting bitten by. If `checked_pid` ever tightens (an upper bound, a pid-1 reject), grok silently won't follow.
2. It is a textbook **sibling-set-completeness** instance (the N-1-of-N class the review taxonomy calls its most recurrent escape).

## Suggested fix

Deserialize `pid` as `i64` and route it through `decoder::checked_pid`, so the per-entry skip semantics match CC's and there is one definition of the narrowing. Add a case to `mod registry_join` pinning that an out-of-range pid skips its entry while the siblings still bind.

## Provenance

Found during the #825 arc while cross-referencing mutation survivors against the shared-guard sites; the validation-gap framing was refuted by reading the guard, and this narrower finding is what survived.

IvanWng97:
## Summary

`grok_ids_from_registry` deserialized straight into `pid: i32` and hand-rolled `e.pid <= 0`, bypassing `decoder::checked_pid`. A pid outside `i32` range therefore failed `from_slice` for the **whole document** — `.ok()?` → `None` → the `shape_drift` breadcrumb asserting the registry's shape changed upstream → all grok liveness degraded to mtime. CC's registry skips the offending entry and keeps its siblings; grok now does too.

## Related issue

Closes #831

## Type of change

- [x] Bug fix

## What this actually buys (honest framing)

This is **not** a validation hole — grok already rejected non-positive pids twice over (`e.pid <= 0`, plus `Pid::from_raw` returning `None` for ≤ 0). And the out-of-range value is **not reachable from a real grok build**: upstream types the field `u32` and writes `std::process::id()`, and `pid_max ≤ 2²²`. Corruption or a hostile file only.

What it buys:

1. `checked_pid`'s completeness claim becomes true for the first time. That doc is what the *next* pid ingress reads to decide whether it is covered — an N-1-of-N contract defect.
2. A false `shape_drift` breadcrumb goes away. It claims *"the registry shape changed upstream"*, which would send an investigator to `check_upstream_drift.py` and xai-org for a rename that never happened.
3. It matches upstream's own per-entry semantics (`is_pid_alive` skips one entry).

Widening the field to `i64` makes the downstream fan-out **compiler-enforced** rather than review-enforced.

## Deliberately not adopted

`cc_probe`'s `RegistryParse` shape-vs-value split. CC's registry is one file per entry, so file-drift ≡ entry-drift and the consumer can name the vanished key. grok's is one document with N entries, where a per-entry `ShapeDrift` has no consumer and no defined aggregation rule.

## How I tested it

`just preflight` — 2851 tests, exit 0.

Three tests, driven red first:

- `out_of_i32_range_pid_skips_only_its_own_entry_and_never_truncates` — asserted in **both entry orders** (today the document fails whichever side the bad entry sits on), with a no-truncation check: `4294967297 as i32 == 1`, so a cast-based narrowing would bind init's pid and vouch a session forever.
- `dead_pids_and_junk_values_are_skipped_not_failures` grows a negative pid and a live sibling — pins the `> 0` half.
- `a_mistyped_or_renamed_pid_is_still_document_level_drift` — the negative control. The derive's remaining strictness **is** the #247 rename detector, and a later "simplify to `Vec<Value>`" refactor would pass every other test in the module while silently deleting it.

Sibling-set check per the parallel-drift pitfall: grepped every pid ingress. `checked_pid` now has exactly four call sites — hook peek, openclaw, cc_probe, grok. The remaining pid sites are the shim (producer), `pid_watch` (consumes an already-checked pid), and `fd_probe` (kernel-supplied, own `> 0` guard).

- [x] Not a visual change.

## Checklist

- [x] Docs updated in the same commit (`checked_pid`'s doc, which is the point of the change).
- [x] Checked against the recurring pitfalls — see the sibling-set grep above and the negative-branch control.

## AI assistance

### DIFF
--- a/crates/pixtuoid-core/src/source/decoder.rs
+++ b/crates/pixtuoid-core/src/source/decoder.rs
@@ -72,8 +72,10 @@ pub(crate) fn accept_all_paths(_p: &Path) -> bool {
 /// ingress rides — the hook peek, the openclaw decode, and BOTH first-party
 /// session registries (`cc_probe::parse_registry_entry`,
 /// `grok::native::grok_ids_from_registry`) — so a new ingress can't ship the
-/// N-th unchecked pid; the sibling set the openclaw
-/// `nonpositive_pid_is_dropped_like_every_sibling_pid_ingest` test names.
+/// N-th unchecked pid. That four-site roster is THIS doc's, not a test's:
+/// openclaw's `nonpositive_pid_is_dropped_like_every_sibling_pid_ingest` pins
+/// only its own arm. `fd_probe` is deliberately NOT a rider — it filters `> 0`
+/// over kernel-enumerated `pid_t`, where the `i32`-range half is vacuous.
 ///
 /// Taking `i64` and narrowing HERE is what keeps an out-of-range value a
 /// per-entry skip: a caller that deserializes straight into `i32` fails its
--- a/crates/pixtuoid-core/src/source/decoder.rs
+++ b/crates/pixtuoid-core/src/source/decoder.rs
@@ -68,11 +68,17 @@ pub(crate) fn accept_all_paths(_p: &Path) -> bool {
 /// Narrow a raw JSON integer to a valid POSIX pid: in `i32` range AND strictly
 /// positive. The `> 0` reject is load-bearing — `kill(0)`/`kill(-n)` target
 /// process GROUPS, and a bogus/zero `_pid` would otherwise synthesize a phantom
-/// exit that flaps a LIVE gateway Down. The ONE narrowing every JSON `_pid`
-/// ingress rides (the hook peek, the openclaw decode, AND the CC
-/// sessions-registry probe in `cc_probe::parse_registry_entry`), so a new ingress
-/// can't ship the N-th unchecked pid — the sibling set the openclaw
+/// exit that flaps a LIVE gateway Down. The ONE narrowing every JSON pid
+/// ingress rides — the hook peek, the openclaw decode, and BOTH first-party
+/// session registries (`cc_probe::parse_registry_entry`,
+/// `grok::native::grok_ids_from_registry`) — so a new ingress can't ship the
+/// N-th unchecked pid; the sibling set the openclaw
 /// `nonpositive_pid_is_dropped_like_every_sibling_pid_ingest` test names.
+///
+/// Taking `i64` and narrowing HERE is what keeps an out-of-range value a
+/// per-entry skip: a caller that deserializes straight into `i32` fails its
+/// whole document instead, which the registry probes report as upstream SHAPE
+/// drift (#831).
 pub(crate) fn checked_pid(raw: i64) -> Option<i32> {
     i32::try_from(raw).ok().filter(|&p| p > 0)
 }
--- a/crates/pixtuoid-core/src/source/grok/native.rs
+++ b/crates/pixtuoid-core/src/source/grok/native.rs
@@ -199,9 +199,10 @@ fn augment_with_fresh_sessions(
 /// process start within [`cc_probe::PID_START_TOLERANCE_SECS`] (the #220
 /// pid-recycle identity check; either side missing → pid-alive-only, the
 /// check is additive). Returns `None` only when the DOCUMENT doesn't parse as
-/// an array of entries (format drift); junk VALUES inside an entry (pid <= 0)
-/// skip that entry silently, mirroring the CC registry's value-vs-shape
-/// distinction.
+/// an array of entries — a renamed or mistyped key, i.e. format drift; junk
+/// VALUES inside an entry (a pid `decoder::checked_pid` rejects — non-positive
+/// OR outside `i32` range — an empty `session_id`, a recycled pid) skip that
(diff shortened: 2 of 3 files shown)


IvanWng97:
## Two-lens review — disposition (correctness / design)

Both lenses: **APPROVE-WITH-NITS**, no correctness defect. Both independently re-verified the sibling set at N=4 and found no N+1, and both ran the gates rather than trusting the claim (`just preflight` exit 0, 2851 tests; plus `doc-check` and `check-windows` at 0). Fixes in `a19a5cfb`; preflight re-run exit 0.

| # | Finding | Lens (conf) | State |
|---|---|---|---|
| 1 | `checked_pid`'s doc delegated its roster to openclaw's test, whose comment omits grok and lists `fd_probe` — a non-rider | design (88) | **FIXED** |
| 2 | `grok_ids_from_registry`'s contract doc still drew the value/shape line at the pre-PR position | design (72), correctness (55) | **FIXED** |
| 3 | The `Entry` field comment restated decoder.rs's paragraph nearly verbatim (4 copies of one WHY) | design (90) | **FIXED** |
| 4 | "compiler-enforced fan-out" holds for 3 of 4 sites — `tracing::debug!` impls `Value` for both widths | correctness (90) | **FIXED** (recorded in the fix-round commit; no code defect) |
| 5 | "out-of-range" unqualified vs the i64 parse ceiling | design (65) | **FIXED** by #2's rewording |

Finding 1 is worth calling out: it is the same defect class this PR exists to fix — a completeness claim whose evidence contradicts it. An author adding pid ingress #5 would have followed the doc's pointer to a test naming three sites that disagree with the doc's four and include a non-rider. The doc now owns the roster; the test is described as pinning only its own arm.

### Verified rather than argued

Both lenses reproduced the claims instead of reasoning about them. The correctness lens built a scratch crate and ran the serde behavior (`bad-first` and `bad-second` both `Err` on i32, both `Ok` on i64), confirmed `4294967297 as i32 == 1`, and ran a 6-implementation × 4-test mutation matrix — `a_mistyped_or_renamed_pid_is_still_document_level_drift` is the **only** test that kills a `Vec<Value>` refactor, which is exactly the negative control's stated job.

One refinement to my own commit message, from that matrix: the both-orders loop does add value, but not for the reason I gave — with only `[ok, over]` a `continue`→`break` mutation survives; the bad-**first** order is what kills it.

### Refuted

Both lenses independently upheld the refusal to adopt `cc_probe`'s `RegistryParse` shape-vs-value machinery, and the design lens sharpened the reason: CC's registry is N independently-written files, so per-file is the only granularity that exists and a torn read must be distinguishable from drift; grok's is one atomically rewritten document (temp+rename), so shape is a document-level property by construction. Adopting it would *lose* the #247 rename detector to buy an aggregation the existing `Once` already provides. No sharp-edge update owed — `CLAUDE.md:284`'s "`None` … on a parse failure" stays literally true, since an out-of-range value is no longer a parse failure.

Issue #638, opened 2026-07-16
grok: leader-mode liveness residual (registry pid = TUI client, writer = leader)


IvanWng97:
## Problem

In grok's opt-in leader mode (\`--leader\` / \`[cli] use_leader\`, plus a server-controlled \`remote_settings.leader_mode\` fallback), the agent — session-file writes AND hook firing — runs in a shared LEADER process, while \`active_sessions.json\` records the TUI CLIENT's pid. pixtuoid's grok liveness ladder (probe/instant-exit/negative-vouch, PR #640) keys on that client pid, so:

- a client disconnect (session lives on in the leader) reads as an exit → the sprite leaves while the session is technically still alive in the leader;
- conversely the ladder never watches the leader pid itself.

## Why deferred

Leader mode is opt-in and uncommon; the failure is cosmetic (early/late sprite exit, self-heals on the next hook/transcript event via re-registration). Detecting leader mode passively would need the \`~/.grok/leader.sock\` probe — new machinery for an edge config.

## Fix sketch

Probe \`{grok_home}/leader.sock\` liveness as a secondary vouch source when the socket exists (a live leader vouches every session whose updates.jsonl is fresh), or read grok's leader session registry if one materializes upstream. Re-evaluate if leader mode becomes default.

PR #841, opened 2026-08-03
fix(grok): bind leader-mode sessions to the leader, not the TUI client

### LINKED ISSUE #826: [Bug] grok leader-mode vouch (#638) is inert — leader_socket_owner asks a vnode probe about a unix socket
## What's wrong

`grok::native::leader_socket_owner` **always returns `None`**, so the #638 leader-mode secondary vouch (`augment_with_leader_vouch`) has never fired on any platform since it shipped in #641 / 92ca2445.

It asks `fd_probe::open_vnode_paths(pid)` whether a pid holds `{grok_home}/leader.sock` open — a question that fn cannot answer, because it reports **VNODE descriptors only** and a unix socket is not one.

```rust
// source/grok/native.rs
fn leader_socket_owner(sock: &Path) -> Option<i32> {
    let sock = sock.canonicalize().ok()?;
    for comm in ["grok", "agent", "xai-grok-pager"] {
        for pid in fd_probe::pids_by_name(comm)? {
            if fd_probe::open_vnode_paths(pid).contains(&sock) {  // ← can never match
                return Some(pid);
            }
        }
    }
    None
}
```

## Evidence (measured, not reasoned)

Ran the syscalls directly against a bound + listening `AF_UNIX` socket on macOS:

```
listening unix socket fd=3  proc_fdtype=2   (PROX_FDTYPE_VNODE=1, PROX_FDTYPE_SOCKET=2)
PROC_PIDFDVNODEPATHINFO -> 0 (want 1200), path=""
```

So on **macOS** the fd is skipped by `fd_probe.rs`'s `if fd.proc_fdtype != PROX_FDTYPE_VNODE { continue; }`, and even unfiltered the vnode-path query returns 0 bytes with an empty path.

On **Linux**, `/proc/<pid>/fd/N` for a socket resolves to the non-path `socket:[inode]`, never the filesystem path — so it could never string-compare equal to the canonical socket path either.

Pinned in-tree by `fd_probe::tests::open_vnode_paths_never_reports_a_unix_socket` (added in #825), which binds a real `UnixListener` and asserts the canonical path is absent from `open_vnode_paths(self_pid)`.

**Corroborating test topology:** the only existing test of `augment_with_leader_vouch` is the absent-socket no-op; every "the vouch works" assertion calls the pure `augment_with_fresh_sessions` with a hand-fed `leader_pid`, bypassing the broken binder entirely. This is the untested-one-line-binder-over-a-real-OS-handle class that #825 fixed at three other sites.

## Impact

**Low, and additive-only — nothing misbehaves.** Grok leader-mode sessions simply keep the documented pre-#638 behaviour: the mtime first-sight gate + short-idle reap + prompt resurrect. The user-visible symptom is the one #638 set out to fix — a leader-kept-alive session's sprite is reaped when its TUI *client* disconnects, even though the leader still owns the session.

Scope: opt-in `--leader` mode on 1 of 13 sources. No crash, no corruption, no data loss, no security exposure.

## Repro

1. Run grok in `--leader` mode; the leader (comm `grok` / `agent` / `xai-grok-pager`) binds and listens on `{grok_home}/leader.sock`.
2. A TUI client registers in `active_sessions.json`, then disconnects; its pid dies.
3. `live_grok_session_ids` → `augment_with_leader_vouch`: `sock.exists()` is **true**, so it proceeds to `leader_socket_owner`.
4. Every comm/pid misses (per the mechanism above) → `None` → `augment_with_leader_vouch` returns `snap` **unchanged**, zero leader-bound vouches.
5. The session falls to the mtime gate and gets reaped on client disconnect.

## Suggested fix

Detect the socket's owner with a **socket-aware** probe rather than the vnode probe, behind its own `fd_probe` fn:

- **macOS**: `proc_pidfdinfo` with `PROC_PIDFDSOCKETINFO`, comparing `psi.soi_proto.pri_un.unsi_addr.ua_sun.sun_path`.
- **Linux**: `stat` the socket to get its inode, then match against `socket:[<ino>]` across `/proc/<pid>/fd/*` (or join via `/proc/net/unix`).

And give `leader_socket_owner` the **injected-enumerator split** the sibling binders got in #825 (`from_open_fds_with`, `ensure_owned_socket_dir_in`), so the composition is testable instead of needing a live leader process.

### Not the fix: `connect()`

Connecting to the socket and treating success as a live leader looks cheaper, but it's wrong here on two counts:

1. This probe runs on the watcher's **250ms rescan**, so it would fire ~4 connections/sec at the user's grok leader.
2. `connect()` proves a listener exists but yields no **pid**, and `augment_with_fresh_sessions` needs one — sessions bind to the leader's pid precisely so leader death ends them via the ExitWatch.

(Peer credentials — `SO_PEERCRED` / `LOCAL_PEERPID` — would supply the pid, but the 250ms connection storm still rules it out.)

## Notes

Found by the whole-codebase audit behind #825, confirmed 3/3 by independent adversarial verification. #825 landed the **enabler half** — `open_vnode_paths`'s Linux arm now honors the vnode contract its own doc claims (it previously returned every `/proc/<pid>/fd` link, including `socket:[inode]`, while macOS filtered) — plus the negative-control test and accurate docs at `leader_socket_owner`, `fd_probe`, and `pixtuoid-core/CLAUDE.md`, which previously asserted the vouch fires.

### LINKED ISSUE #839: grok leader mode: session ownership is not observable, so pixtuoid cannot model it
## What

In `--leader` mode the agent runs in a shared **leader** process while `{grok_home}/active_sessions.json` records only each TUI **client's** pid. pixtuoid binds sessions to that client pid, so a client detach — leader mode's *normal* operation, the thing `no_exit_on_disconnect` exists for — reads as the session's exit and the sprite walks out.

#638 tried to close this with a secondary vouch off `leader.sock`; it never fired on any platform and was deleted in #851 rather than repaired. This issue is the durable record of what is actually missing.

## Why it can't be fixed in-tree

Session→leader ownership lives **only** in `leader/server.rs`'s in-memory `session_driver` / `session_subscribers` maps. Every disk-observable surface answers a different question:

| surface | answers | gap |
|---|---|---|
| `active_sessions.json` | the client pid per session | no leader field — `ActiveSession` is `{session_id, pid, cwd, opened_at}` |
| `leader{suffix}.lock` | leader pid + liveness (exclusive flock, pid in the file) | not *which* sessions it owns |
| transcripts | content | the writer opens and drops the fd per append (`session/storage/jsonl/mod.rs::append_jsonl_line_sync_with`), so no Codex-style open-fd probe |
| control socket | 9 `ControlCommand` variants | none lists sessions; `WorkspaceStatus.sessions` is the computer-hub workspace's, a different concept |

So any in-tree mechanism is a proxy with its own failure mode — which is exactly what #638 was, and what the (now closed) #843 and #844 were projections of.

## The fix: ask upstream to expose the fact

Either of these collapses the whole class at once, and both are small on the upstream side:

1. **`ControlCommand::ListSessions` → `[{session_id, driver_client_id}]`.** The leader already holds this map in memory; this is a read-only command. pixtuoid would query once per leader lifetime, cached and invalidated by the existing `ExitWatch` on the leader's pid — the "connection storm" objection does not apply, since the probe runs at seed + one 250ms rescan + the 60s poll.
2. **A `leader_pid` field on the `active_sessions.json` entry.** Cheaper still — upstream already writes that file, and pixtuoid already parses it.

## Current behaviour (deliberate, documented)

Leader sessions are unvouched and ride the pre-#638 path: mtime first-sight gate + short-idle reap + prompt resurrect. Unchanged from the feature's entire shipped life. Recorded as the grok LEADER MODE sharp edge in `crates/pixtuoid-core/CLAUDE.md`, and pinned by `a_present_leader_socket_binds_nothing_the_registry_did_not`.

## Impact

Low. pixtuoid is a visualizer: the failure is a sprite disappearing, with no data loss and the agent unaffected, and `resurrects_on_prompt: true` brings it back on the next prompt. Worth noting for exposure, though, that `RemoteSettings.leader_mode` lets the server enable leader mode for users who have not set `[cli] use_leader` locally (precedence: `--no-leader` → `--leader` → eligibility → local config → **remote** → default off), so the affected population is not purely opt-in.

## Next step

File the upstream request against `xai-org/grok-build`. Until it lands there is nothing correct to build here.

IvanWng97:
## Summary

#826 reported that the #638 leader vouch is inert — `leader_socket_owner` asks `fd_probe::open_vnode_paths` whether a pid holds `leader.sock` open, and that fn reports VNODE descriptors only. True, but **secondary**.

The primary defect is ownership. `augment_with_fresh_sessions` added the leader pid with `or_insert`, so the registry's **client** pid still won — and that is the pid `ProbeLadder::fold` hands the `ExitWatch`. A client disconnect therefore ended a session the leader was still driving, in **milliseconds**, while the leader binding could only migrate in on the 60s poll. **A byte-perfect socket probe could not have reached #638's own symptom.**

So this fixes the binding and drops the probe, rather than repairing a mechanism that was answering the wrong question. Two commits: remove the inert vouch, then land the correct one.

## Related issue

Closes #826. Closes #839.

## Type of change

- [x] Bug fix

## How the leader is detected — no FFI, no socket connect, no mtime

Upstream's own first-party artifact. `run_leader` holds an **exclusive `flock`** on `leader{suffix}.lock` for its whole lifetime and writes its pid into the file:

```rust
pub fn write_pid(&mut self) -> Result<(), LockError> {
    if let Some(ref mut file) = self.lock_file {
        file.set_len(0)?;
        write!(file, "{}", std::process::id())?;
        file.sync_all()?;
    }
    Ok(())
}
```

So a failed **shared** lock is the liveness proof and the contents are the identity — the same advisory-lock arbitration `hook/unix.rs` already uses. The lock is what proves life: a killed leader leaves the file with its stale pid, so a contents-only read would vouch a corpse. Pinned by the unheld-lock arm.

When a leader is live, every registry entry binds to **its** pid and the entry's own pid stops being consulted — hence no pid-recycle check on that path, since that check identifies a client process the vouch no longer depends on.

Path resolution mirrors upstream rather than guessing: the suffix is a `DefaultHasher` hash of the WS URL, whose output std explicitly does **not** guarantee stable across versions, so the resolver **enumerates** `leader*.lock` instead of recomputing it, and honors `GROK_LEADER_SOCKET` (which bypasses the naming entirely). The deleted code hardcoded `leader.sock` and would have missed both forms even once socket-aware.

## Deliberately not reinstated

The mtime-derived freshness window. It was net-negative: 150s expires while the session lives, and an id leaving a *healthy* snapshot is the negative-vouch trigger (~240s), where the unvouched session survives to the 300s short-idle reap — `LEADER_SESSION_FRESH_SECS(150) + PROOF_OF_LIFE_TTL(150) = STALE_SHORT_IDLE_TIMEOUT(300)`. The new binding adds no expiring assertion, so that failure mode does not exist here.

## Accepted residual

A client that quits **cleanly** unregisters its entry, so a leader-kept session leaves the snapshot and the negative vouch ends it in ~60–120s. Not closable from disk: the leader's session ownership lives only in `leader/server.rs`'s in-memory maps, and its control protocol has no list-sessions command (`WorkspaceStatus.sessions` belongs to the computer-hub workspace — a different concept; I checked and was wrong about this first). Closing it needs a cached control-socket query, worth building only if a user reports it.

## How I tested it

`just preflight` — 2850 tests, exit 0.

The ownership test drives the **real** detector end-to-end: this process takes the leader's own exclusive `flock` and writes its pid the way `write_pid` does. `flock` is per open-file *description*, so the detector's own `File` contends with ours inside one process — testable on every platform without spawning a leader. **Proven red against a stubbed detector** before being called done:

```
test ...::a_live_leader_owns_the_registrys_sessions_instead_of_the_client ... FAILED
```

Plus: path resolution across the suffix and env-override forms; and a negative-control arm for no-candidates, unopenable, and held-but-unparseable-pid.

**What is not verified:** I have no leader-mode grok to run this against (`~/.grok` here has no `leader.lock`). The detector is exercised against real `flock` and real files; "grok end-to-end behaves as modelled" rests on reading upstream source, not on a live run.

- [x] Not a visual change.

## Checklist

- [x] New behavior has a test, proven red first.
- [x] Docs updated in the same commit — the LEADER MODE sharp edge is rewritten around ownership, records the residual and why it is unclosable, and keeps the four findings so this is not re-litigated.
- [x] Checked against the recurring pitfalls — parallel-drift in particular: grepped every reference to the removed items (4 sites, all updated).

## AI assistance


IvanWng97:
Refuting the MEDIUM finding on `crates/pixtuoid-core/CLAUDE.md:284` — it is a false positive. The bot quoted the **pre-PR** text as if it were current; that paragraph was fully rewritten in this PR's first commit (101bac8a) and rewritten again in f773c18b for the flock design.

Current content at HEAD:

```
LEADER MODE (#638/#826) — **the leader OWNS its sessions; the registry's client pid
does not.** `--leader` runs the agent (file writes + hooks) in a shared LEADER process
while `active_sessions.json` still records each TUI CLIENT's pid. #638 tried to fix the
resulting client-disconnect-kills-the-sprite bug by ADDING a secondary vouch off
`leader.sock`; that was wrong twice over and never fired. **The primary defect was the
binding, not the probe** ...
```

Checked mechanically at HEAD:

- `"CURRENTLY INERT"` → **absent**
- `"is MEANT to add a SECONDARY vouch"` → **absent**
- `"The DESIGN below is what it will do once the probe can answer"` → **absent**
- `LEADER_SESSION_FRESH_SECS` anywhere in the file → **absent**

The rewritten block describes the flock detector, names `live_leader_pid` / `leader_lock_candidates`, records why the lock rather than the pid file proves life, documents the `DefaultHasher`-suffix enumeration and the `GROK_LEADER_SOCKET` override, and states the accepted residual (a cleanly-quitting client unregisters its entry, so the negative vouch ends a leader-kept session in ~60–120s — unclosable from disk, since the leader's session map is in-memory only and its control protocol has no list-sessions command).

The bot's diff observation was right in one respect: the doc change is hard to see, because `CLAUDE.md` stores each sharp edge as one very long line, so a whole-paragraph rewrite renders as a couple of changed lines next to an unrelated tweak. Worth knowing when reviewing this file.

No change made for this finding. The sibling finding on #842 was **valid** and is fixed there (8b032d0e).

IvanWng97:
## Two-lens review — disposition (4 lenses: correctness / design / lifecycle / comment)

Escalation triggers fired: the diff touches the **liveness ladder** (lifecycle lens) and reshapes a substantial block of doc comments (comment lens). Three of four returned REQUEST-CHANGES on two genuine defects — both of which reintroduced the mass-instant-exit class this PR exists to remove. Fixed in `8973d94c`; `just preflight` exit 0 (2853 tests), `check-windows` 0, `api-surface-check` 0.

| # | Finding | Lenses | State |
|---|---|---|---|
| 1 | `live_leader_pid` bound the lock file's contents with no liveness check | correctness (88), lifecycle (58) | **FIXED** |
| 2 | Leader arm dropped `alive(entry.pid)`, the only crash-residue filter | lifecycle (82), design (80), correctness (72) | **FIXED** |
| 3 | `grok_pid_for_session` (focus click) silently resolved the leader | design (88) | **FIXED** |
| 4 | `leader_lock_candidates` doc promised an order the code lacked; env override replaced rather than extended | comment (88), design (72), lifecycle (70/62) | **FIXED** |
| 5 | `O_NOFOLLOW` undone by a follow-up path-based read; `pid <= 0` guard missing from the leader arm | correctness (70), design (85) | **FIXED** |
| 6 | Four doc claims inaccurate (both probe summaries, "on every platform", the hook-socket comparison) | comment (62/88), design (80) | **FIXED** |
| 7 | Test comment described a `pid 1` fixture that wasn't there; `999999` is allocatable on Linux | comment (92), correctness (90) | **FIXED** — dead pids now come from a spawned-and-reaped child |
| 8 | `leader_detection_is_quiet_…` claimed to exercise `checked_pid`; input died at `parse` | correctness (88) | **FIXED** — now also `"0"`/`"-5"` |
| 9 | Comment redundancy: `grok_ids_from_registry` 3 sentences/1 idea, `live_leader_pid` 5/4 | comment (70/72) | **FIXED** |
| 10 | Leader death mass-ends N sessions, next poll re-binds and replays at cursor 0 | design (55), lifecycle (72) | **ISSUE-FILED** below |
| 11 | Multi-leader: one leader owns everything, its death ends the other's sessions | correctness (72), lifecycle (70) | **ISSUE-FILED** below |
| 12 | "not closable from disk" overstates — the deleted mtime join was such an artifact | lifecycle (65) | **REFUTED**, see below |
| 13 | `CLAUDE.md` still describes the deleted design | online bot (MEDIUM) | **REFUTED-STALE** — quoted the pre-PR text; mechanically checked absent at HEAD |

### The two blockers, in detail

**#1 is the one that mattered.** The correctness lens went to upstream and found `leader/mod.rs::live_grok_lock_holder`, which gates the lock file's pid on `is_process_alive` for a documented reason I had missed:

> `None` for a dead / non-grok PID, OR when the file PID can't be confirmed to be the real flock holder … (a stale-but-live PID left in `leader.lock`, or **a brief spawner that held the flock without rewriting the file**).

I verified that myself before fixing. `connect_or_spawn` holds the flock across the spawn handoff **without** calling `write_pid`, so the file still names the previous — dead — leader. Binding it registers a dead pid with the ExitWatch, which synthesizes an immediate exit on ESRCH: every grok session ends at once. Now filtered through `cc_probe::pid_alive`.

**#2**: the leader arm bypassed `alive(entry.pid)`, this registry's only filter against crash residue ("removed on clean quit, left behind on crash"). A vouched id bypasses the first-sight gate, emits ProofOfLife, and is exempt from the stale sweep, the short-idle reap **and** the negative vouch simultaneously — an immortal ghost desk per leaked entry. My own pinned test demonstrated it (it asserted a dead `pid:999999` binds). The leader now moves the **binding** only, never admission.

### Test teeth — each proven red by mutation

- drop the `pid_alive` filter → `a_held_lock_naming_a_dead_process_vouches_nothing` fails
- drop the admission filter → `a_live_leader_does_not_vouch_crash_residue_from_a_dead_client` fails
- re-point `grok_pid_for_session` at the leader map → `focus_resolves_the_client_pid_even_while_a_leader_owns_the_session` fails

That last one **initially survived** — the test called the new helper directly instead of the consumer's entry point, i.e. the declared-not-wired class. Caught by the negative control and fixed.

### #12 — refuted, with the sharp edge updated

The lifecycle lens is right that the deleted `augment_with_fresh_sessions` was a disk artifact, so "not closable from disk" overstated. But the honest claim is narrower than either version: the leader's session ownership is in-memory only, and the mtime join was a *proxy*, not the ownership fact — which is why it was net-negative. The sharp edge now says the residual is not worth closing rather than impossible, and records the guard reasoning for #1 and #2 so neither is re-litigated.

### #10 / #11 — deferred, and why

Both are real, both need a design call rather than a patch (what `fold`'s rebind migration should do at an ownership boundary; whether multi-leader wants per-leader session attribution that no disk artifact provides). Filing rather than patching under the repo's only-big/refactor-defers rule.

IvanWng97:
Deferred findings #10 and #11 from the disposition table are now filed: **#843** (leader death mass-ends its sessions, next poll replays at cursor 0) and **#844** (two live leaders, one arbitrarily owns everything). Both carry the trace, why each is a design call rather than a patch, and an acceptance criterion.

IvanWng97:
## REFUTED (2nd time) — the bot's MEDIUM quotes text that is not in the file

Checked mechanically against `git show origin/refactor/grok-delete-inert-leader-vouch-826:crates/pixtuoid-core/CLAUDE.md` — i.e. the exact sha the bot reviewed (`8973d94c`), not my working tree:

| String the finding attributes to the file | Actually at HEAD |
|---|---|
| `CURRENTLY INERT` | **absent** |
| `is MEANT to add a SECONDARY vouch` | **absent** |
| `The DESIGN below is what it will do…` | **absent** |
| `Fixing it needs a socket-aware probe` | **absent** |
| `augment_with_leader_vouch` | **absent** |
| `LEADER_SESSION_FRESH_SECS` | **absent** |

And the things it says are missing are present: `leader_lock_candidates`, `GROK_LEADER_SOCKET`, `is_process_alive`.

The block at HEAD opens:

> LEADER MODE (**#638/#826**) — **the leader OWNS its sessions; the registry's client pid does not.** `--leader` runs the agent (file writes + hooks) in a shared LEADER process while `active_sessions.json` still records each TUI CLIENT's pid…

and goes on to cover exactly the four things the finding asks for: the lock-file mechanism, the `GROK_LEADER_SOCKET` override and `leader_lock_candidates` ordering, the alive-vs-identity distinction (with upstream's `live_grok_lock_holder` as the citation), and the clean-quit residual. The claim that "#826 should no longer be described as unfixed" is already satisfied — the heading carries both issue numbers.

**Why it keeps misreading this file:** each sharp edge in `pixtuoid-core/CLAUDE.md` is stored as one very long line, so a full-paragraph rewrite renders as ~2 changed lines next to unrelated edits. The finding's own premise — "unmodified by this diff" — is what's wrong; everything downstream of it follows from reading the base text. This is the second time the same finding has been raised on this PR from pre-PR content (the first was against `f773c18b`).

**Disposition: REFUTED-with-trace.** No change made. Note the independent comment lens — dispatched specifically because this diff reshapes doc comments — read this same block against the code and did *not* find it stale; it verified every upstream claim in it via `gh api` and judged it "15 sentences, 14 ideas", explicitly declining to flag it.

Per the two-lens-review protocol's step 6, the merge gate's `Findings: 0`-at-HEAD condition is unsatisfiable here for a false positive, so the documented fallback applies: the extra differentiated lenses have run (four, vs the two-lens floor), the disposition is recorded in this thread, and merge is the owner's call.

IvanWng97:
Superseded by #851, which deletes the vouch instead of replacing it.

The work here is sound — four lenses, green CI, every finding dispositioned — but it fixes the **symptom** while leaving four documented residuals (registry-less sessions, clean-quit unregister, leader-death replay, multi-leader attribution) and two design issues. All four are projections of one missing fact: session→leader ownership lives only in `leader/server.rs`'s in-memory maps, and no disk-observable surface carries it.

That was worth learning here rather than guessing at it — the investigation is what produced the observable-surface table now recorded in the sharp edge, including the verified reason a Codex-style open-fd probe is impossible (the transcript writer opens and drops the fd per append). #851 keeps that record and drops the mechanism.

Not deleting the branch, in case the upstream ask in #839 lands and this becomes the starting point.

IvanWng97:
Branch cleanup correction: I said above I was keeping `refactor/grok-delete-inert-leader-vouch-826`. It's deleted — but nothing is lost, and I verified that before removing it: GitHub keeps this PR's head reachable at `refs/pull/841/head`, which resolves to `b1ce1abf`, byte-identical to the branch tip it had.

To recover the flock implementation if the upstream ask ever lands:

```
git fetch origin refs/pull/841/head:leader-flock
```

That gets the whole reviewed line — the four-lens round and its fixes included.

Issue #637, opened 2026-07-16
grok: byte-real wire capture → anchor verified_version + sanitize fixtures


IvanWng97:
## Problem

The grok source (#640) shipped with wire facts REPO-DERIVED from the open-source sync (xai-org/grok-build @ c68e39f6, v0.1.220-alpha.4) and adversarially verified against its serde structs + its own test fixture helpers — but NO byte-real capture anchors them: grok isn't installed on the dev machine and running a session needs an xAI account. The registry row carries \`verified_version: "unknown"\` per the house rule.

## Why deferred

Requires an authenticated grok install (external account) — everything code-side is done and drift-watched (check_upstream_drift.py fetches event.rs / notification.rs / active_sessions.rs weekly).

## Fix sketch

1. Install grok (\`curl -fsSL https://x.ai/cli/install.sh | bash\`), run one session in a scratch repo with pixtuoid connected (Sources panel → grok).
2. Dogfood: \`pixtuoid run --headless\` vs the live session — verify sprite lifecycle (register → tools → waiting on a permission prompt → clean exit via negative vouch in ~60–120s after TUI quit).
3. Capture: the real \`updates.jsonl\` + shim-logged hook envelopes; sanitize; replace/augment \`tests/sources/fixtures/grok/\` (the omp precedent: sanitized live captures).
4. Anchor \`verified_version\` to the captured \`grok --version\` output in \`registry.rs\` + note the capture date in the row comment.
5. Empirically confirm the two research-flagged unknowns: (a) the SubagentEnd-vs-SubagentStop file-hook spelling (we register BOTH — verify which fires), (b) headless \`-p\` hook firing.

PR #842, opened 2026-08-03
fix(grok): stop breadcrumbing task_complete as unknown notification drift

IvanWng97:
## Summary

Upstream dispatches a notification hook with `notification_type: "task_complete"` on every background-task completion. Our notification arm matched only `permission_prompt` / `elicitation_dialog` / `idle_prompt` / `agent_error`, so it fell to the `other` arm and fired `drift::unknown_event`. That breadcrumb is undeduped on the hook plane by design, so routine background work reported as a recurring "the wire changed" warning — surfaced by `pixtuoid doctor` and the footer drift nudge, pointing an investigator at upstream for a rename that never happened.

## Related issue

n/a — found while re-checking grok for upstream drift. It is a miss from the #637 sync rather than new drift: the dispatch predates our 0.2.102 anchor.

## Type of change

- [x] Bug fix

## Verification (upstream, read this session)

`xai-org/grok-build`, `crates/codegen/xai-grok-shell/src/tools/notification_bridge.rs`:

```rust
.send(SessionCommand::DispatchNotificationHook {
    notification_type: "task_complete".into(),
    message: Some(format!("Background task completed: {task_id}")),
    ...
});
```

and three of its own tests assert that exact string. All five `notificationType` values we know of still exist upstream. The completion is not a lifecycle signal for us — a background task's start and end already ride its own tool events — so it joins the knowingly-ignored arm.

## Two items deliberately NOT included, having failed verification

A drift recheck also proposed these. Both died on inspection, and the reasoning is in the commit body so it isn't re-proposed:

- **A drift-watch entry for the `notificationType` value vocabulary.** Upstream types the field as a bare `String` and produces the values at scattered call sites, so no declaration *owns* them. Per the #793 lesson an anchor would be the weak grade at best, and a name-presence sweep over call sites is precisely the false-positive machine that taught us the rule. The `other` arm's breadcrumb (defense #2) is the right mechanism for a vocabulary with no owning declaration — the Antigravity class.
- **Bumping grok's `verified_version` 0.2.102 → 0.2.118.** Upstream is indeed at 0.2.118 (confirmed from `xai-grok-shell/Cargo.toml`), but that field means *"the version this build's decoder **and fixtures** were last verified against, by byte-real capture"*. This recheck read upstream source; it did not re-capture live bytes. Bumping it would assert a verification nobody performed.

Otherwise the grok drift recheck was clean — no verified change, no probe-health failure.

## How I tested it

`just preflight` — 2850 tests, exit 0.

The existing test could not have caught this: a knowingly-ignored type and an unrecognized one both decode to zero events. The new `known_notification_types_stay_silent_while_a_novel_one_breadcrumbs` separates them on the breadcrumb, with a novel type as the control so the silence assertions are recognition rather than a dead detector. Driven red first — it failed on exactly `task_complete`.

- [x] Not a visual change.

## Checklist

- [x] Docs updated in the same commit — the decoder's event-mapping doc list.
- [x] Checked against the recurring pitfalls — the ignore-list completeness question in particular: the five values were each confirmed to exist upstream, and the `other` arm's breadcrumb remains the backstop for the N+1.

## AI assistance


IvanWng97:
## Two-lens review — disposition (correctness / design+drift)

Both lenses: **APPROVE-WITH-NITS**, no correctness defect. Fixes in `c4c188c5`; `just preflight` exit 0 (2849 tests).

| # | Finding | Lens (conf) | State |
|---|---|---|---|
| 1 | The WHY was wrong: "lifecycle already rides that task's own tool events" contradicts this file's own b1 documentation | design (82), correctness (60) | **FIXED** (in both places — it had been copied into CLAUDE.md) |
| 2 | The defense-#2-only drift decision lived only in a commit body | design (80) | **FIXED** — now a sharp edge in the CC-transcript/Antigravity shape |
| 3 | The old test is strictly subsumed; its name and comment had gone stale | design (85), correctness (85) | **FIXED** — deleted |
| 4 | "the non-blocked types" is a weak umbrella | design | **FIXED** → "the non-Waiting types" |
| 5 | Bot's stale-CLAUDE.md MEDIUM | online bot | **FIXED** in `8b032d0e` before this round |

**Finding 1 is the one I got wrong.** For a backgrounded dispatch `post_tool_use` fires at SPAWN, not completion — the load-bearing premise of grok's own b1 trap, documented 110 lines above the comment I wrote. The tool events cover the dispatch window only; the completion instant rides nothing. The decision to drop `task_complete` still stands, for the reasons that actually hold (not an agent transition; no `toolUseId`, so no activity to close), and upstream confirms the shape — `TaskCompleted` carries a `TaskSnapshot` whose `kind` is `Bash|Monitor`, never a subagent.

### Both deliberate exclusions independently upheld

The correctness lens did what I should have done and enumerated **every** upstream construction site rather than trusting my list of five: `dispatch_notification_hook` is the single chokepoint, with six call sites minting exactly five distinct values. All five are now explicitly matched, and the `_ =>` backstop remains for the `notification_hook_for_update` catch-all that could gain a variant.

- **No drift-watch row** — upheld against `check_upstream_drift.py`'s own anchor grading (owner-grade > identity-grade > never a checked name). `notification_type: String` owns no values; they are minted at 6 sites across 5 files in 2 crates. Now recorded as a sharp edge, per finding 2.
- **No `verified_version` bump** — upheld textually: the field means "decoder **+ fixtures** … byte-real capture", and the fixtures were not re-captured. One lens notes upstream is at **0.2.119** today, not 0.2.118 — the direction holds and the skew nudge is correctly still pointing at the re-capture that is actually owed.

### Merge-gate note

Both lenses flagged that the bot's `Findings: N` comments are pinned to `2066756f`, the first commit. The gate wants `Findings: 0` **at HEAD**, so this needs a bot re-run against `c4c188c5` before merge.

PR #856, opened 2026-08-04
fix(deps): unblock the npm audit gate and take every in-range npm update

IvanWng97:
## Summary

`npm audit` went red on **main** for both npm projects, which gated every site and raycast PR on a break none of them introduced. This unblocks the gate and takes all remaining in-range npm updates in the same pass.

`npm audit` resolves advisories **live** against the GitHub Advisory Database at run time — it reads nothing committed — so an unchanged lockfile flips green→red the moment an advisory is published. Three landed on 2026-08-03 between 16:35 and 19:59 UTC; Dependabot opened #846–#849 at 00:03 UTC on 08-04. Those PRs were **born red**, hours after the cause.

Two amplifiers made it worse than it looked: site pins `--audit-level=low`, so any severity blocks; and `npm run audit` is the **first** step of both jobs, so format / lint / astro check / knip / build / check:docs / e2e / lighthouse (site) and gen:contract / tsc / eslint (raycast) never ran at all. Those bumps weren't merely blocked — they were **entirely unverified**.

## Related issue

n/a — supersedes #846, #847, #848, #849 (see below). Closes the 7 open Dependabot alerts.

## What's here

**1. The advisory fix (site)** — no `overrides` added. Every parent already declared a caret range admitting the patch (minimatch `^5.0.5`, ajv `^3.0.1`, socks `^10.1.1`), so these were stale lockfile pins, not unsatisfiable constraints. Patched versions checked against `gh api /advisories/<GHSA>`, not just `npm audit`:

| package | → | advisory | first patched |
|---|---|---|---|
| brace-expansion | 5.0.9 | GHSA-rgw5-rvv9-x895 | 5.0.9 |
| fast-uri | 3.1.5 | GHSA-7p8r-x3mc-p8w7 | 3.1.5 |
| ip-address | 10.4.0 | GHSA-mwp4-54f8-5fhr (+2 moderate) | 10.3.1 |

The `overrides` block is load-bearing; three more entries would have outlived their advisories.

**2. The advisory fix (raycast) + retiring its adjudication.** The same advisory hit **both** brace-expansion copies — 5.0.8 under `eslint → minimatch@10`, and 2.1.2 under the pinned `@oclif/core → ejs → jake → filelist → minimatch@5`. Both float in range (`^5.0.5`, `^2.0.1`) → 5.0.9 and 2.1.4, no override; staying on the 2.x line sidesteps #792, where a 5.x override went green over code that throws `expand is not a function`.

Landing 2.1.4 then tripped the script's *other* arm: `STALE GHSA-mh99-v99m-4gvg`. Verified upstream has since **backported** that DoS fix to 2.1.3 / 3.0.3 / 1.1.17, so the entry's premise ("every 1.x–4.x line is unpatched, first patched 5.0.8") is no longer true and the refusal it justified has outlived its reason. Entry deleted per the script's own instruction; the Map is now empty, which is its **passing** state, not a disabled gate. `integrations/raycast/CLAUDE.md` updated to match.

> **#848 could never have gone green on its own.** Its lockfile diff is byte-identical to this one, but Dependabot cannot delete a stale adjudication entry — its CI fails on exactly that `STALE` line. This PR is its missing half.

**3–4. The superseded bumps**, folded in and actually verified: site-deps + playwright (#849, #847) and raycast eslint (#846).

**5. `npm update` in both projects** — all remaining in-range patch/minor (site: 119 moved / 36 added / 31 removed; raycast: 21 / 9 / 13). Direct movers: `eslint-plugin-astro` 3.1.0, `typescript-eslint` 8.66.0, `@raycast/api` 1.104.24, `@types/react` 19.2.18.

## Notes for review

**Four packages land above what Dependabot proposed** — astro 7.1.6 (not 7.1.4), knip 6.31.0 (not 6.29.0), globals 17.9.0 (not 17.8.0), the playwright trio 1.62.1 (not 1.62.0). Those PRs predate the newer patches; regenerating resolves each caret to what is current now. Declared ranges were bumped to match so the manifest and lockfile tell the same story.

**TypeScript stays on 6, `@types/node` (raycast) stays on 22** — deliberate, and the existing carets enforce it mechanically. Both are explicit `ignore` entries in `dependabot.yml`, and the peer ranges still back them at their **latest** versions: `@astrojs/check@0.9.10` peers `typescript ^5.0.0 || ^6.0.0`; `typescript-eslint@8.66.0` and `@raycast/eslint-config` peer `>=4.8.4 <6.1.0`. `npm install typescript@^7` conflicts across 12 packages and npm declines to apply it. After this PR, `npm outdated` in both projects reports **only** those two blocked majors — the intended steady state.

**Two pre-existing flake classes surfaced under load — neither is reachable from this diff, which touches zero `.rs`/`Cargo.*` files.**

1. *site e2e.* A **different** keyboard-interaction test failed on two of five full local runs (statusline keys, install tabs, key vocabulary), each passing in isolation. The first one landed on the advisory-fix commit alone, before any bump. CI absorbs it via `retries: CI ? 1 : 0` — the CI-semantics run exits 0 reporting "1 flaky, 95 passed".
2. *`pixtuoid-hook::shim`.* The first `just preflight` run failed `codewhale_event_mode_builds_envelope_from_env_and_ignores_stdin` on its **2s** child-exit bound. Re-run 3× in isolation it completes in 0.05–0.24s — two orders of magnitude under the bound — so the bound is tight enough to lose to machine load. Worth knowing that nextest's fail-fast then left **784 tests unrun**, so that run proved much less than its output suggested. The clean re-run is 2853/2853.

Flagging rather than fixing here; happy to file issues for either.

## Type of change

- [x] Bug fix
- [x] Refactor / chore

## How I tested it

Ran **every step of both CI jobs locally** — something neither job has managed since the advisories landed, because both die at step one.

**site** — `npm ci` · audit (`0 vulnerabilities`) · `format:check` · `lint` · `astro check` · `knip` · `test:unit` (56 pass) · `build` · `check:docs` · **e2e** (96 passed; exit 0 under CI semantics) · **lighthouse** (18 runs / 6 routes, all assertions green)

**raycast** — `npm ci` · audit (`0 advisory/advisories, all adjudicated`) · `gen:contract` freshness diff · `tsc --noEmit` · `eslint .` (clean but for the pre-existing `@raycast/prefer-common-shortcut` warning) · `npm install-scripts ls` (only the pre-existing blocked `esbuild@0.28.1` postinstall, unchanged here)

**workspace** — `just lint` · `just risk-radar-test` · `just preflight` (exit 0; **2853/2853 tests passed**, 2 skipped)

## Visual verification

- [x] Not a visual change.

## Checklist

- [x] I read the relevant `CLAUDE.md` (root + `site/` + `integrations/raycast/`).
- [x] New behavior has a test — n/a, no behavior change; the gates above are the coverage.
- [x] No `unwrap()` in non-test code — n/a, no Rust touched.
- [x] No new `println!`/`eprintln!` on a production path — n/a, no Rust touched.
- [x] Docs updated in the same commit — `integrations/raycast/CLAUDE.md`, alongside deleting the adjudication entry it described.

## AI assistance

---

## Review-round additions

Findings from the three lenses, all driven to a terminal state.

**Fixed — the fail-open one matters most.** Emptying `ADJUDICATED` removed the last thing standing between a registry outage and a green supply-chain gate: a failed `npm audit` returns valid JSON `{message, error}` with no `vulnerabilities` key, which the script read as "zero advisories" and **passed**. The single adjudicated entry had been an accidental liveness probe via the STALE arm. So the script is now **deleted** — raycast runs `npm audit --audit-level=low`, the same line site runs, whose exit code has no such hole. Verified exit 1 against a dead registry, exit 0 against the real one.

**Fixed — the audit no longer runs first in either PR gate.** It resolves advisories live, so at step one an unrelated publication skips every check below it — exactly how #846–#849 ended up *unverified* rather than merely blocked. `pages.yml` deliberately keeps it first: that job publishes. The OPA rules match on the run string and working directory, never position (`just ci-observability` 111/111).

**Fixed — doc corrections.** An inline "14 packages change" (method-dependent, so not a durable fact); a runbook that read literally as "delete the committed lockfile"; and prose that generalized "npm audit is the wrong test" from `chrome-launcher` — inverting the real ranking, since that entry is the one with *no* advisory backstop while the retired yaml pin had one. Plus a redundant clause and a vacuous comment left behind by the deletion.

**Note on commit `975c316d`'s message.** It says "in-range patch/minor"; 17 hoist-root entries do cross a major (`yargs` 17→18, `js-yaml` 3→4, `argparse` 1→2, `escape-string-regexp` 1→4, …). Traced: every one is a **hoist reshuffle**, not a consumer upgrade — the newer copy is hoisted for a newer direct dep's own range while the old major moves into a nested copy, and no consumer resolves outside what it declares (e.g. `yargs@18.1.0` is hoisted for `@astrojs/check@0.9.10`'s `^18.0.0` while `lighthouse` keeps a nested `17.7.3`). Recorded here rather than rewriting the branch's history.

**Gate teeth proven, not assumed.** Before deletion, both arms of the adjudication script were negative-controlled at an empty Map (clean → exit 0; bogus entry → `STALE`, exit 1), and the `UNREVIEWED` arm had already fired for real earlier in this PR.


IvanWng97:
REFUTED — the HIGH raised at `40ab89f3` reads the **deleted** script's comment as current fact.

That comment said brace-expansion was *"unpatched on every 1.x-4.x line (first patched: 5.0.8)"* and would clear only when `@oclif/core` moved off `ejs ^3`. That was true when it was written. It is exactly the premise that stopped being true, which is why the entry was retired — and why the comment stating it was deleted in the same change.

**Upstream backported the fix.** Live advisory data, fetched just now:

```
$ gh api /advisories/GHSA-mh99-v99m-4gvg --jq ".vulnerabilities[] | \"\(.vulnerable_version_range) -> \(.first_patched_version)\""
>= 4.0.0, < 5.0.8  -> 5.0.8
>= 3.0.0, < 3.0.3  -> 3.0.3
>= 2.0.0, < 2.1.3  -> 2.1.3      <-- the 2.x line IS patched
< 1.1.17           -> 1.1.17
```

So the pinned copy never had to leave the 2.x line — it took the fix in-range:

```
$ npm ls brace-expansion --all
  filelist@1.0.6 -> minimatch@5.1.9 -> brace-expansion@2.1.4    (>= 2.1.3, patched)
  eslint@10.8.0  -> minimatch@10.2.6 -> brace-expansion@5.0.9
```

The finding makes a falsifiable prediction — *"will therefore likely fail CI on `raycast.yml`'s `npm run audit` step"*. It does not:

```
$ npm run audit          # = npm audit --audit-level=low
found 0 vulnerabilities
EXIT=0
```

and the `types · lint` job **passes** on that same head. Staying on 2.x is also the deliberate choice, not an oversight: 5.x turns the export from a function into an object that `minimatch@5.1.9` calls as `expand(pattern)`, so a `brace-expansion@5` override goes green over code that throws (#792) — still reproducible at 5.0.9.

The reviewer had a fair reason to be misled, and that is on me: an earlier trim in this PR removed `2.1.3` from `integrations/raycast/CLAUDE.md`, leaving only "its one entry cleared upstream" with no number to check against. `c1c042fd` puts the backport version back and marks the stale reading explicitly, so the next reader — human or bot — has the disambiguating fact in front of them.

Issue #830, opened 2026-08-03
[Test] The 2026-08 audit covered 25% of the tree — the largest functions in the repo were never in scope


IvanWng97:
## What

The whole-codebase audit behind #825 scoped itself to `pixtuoid-core/src/source/` + `src/state/` — **33,590 of 134,167 src LOC (25%)**. The remaining 75% was never examined, and it contains the biggest structural targets in the workspace.

Re-running the same length/nesting measurement **outside** the audited scope:

| lines | nest | fn |
|---|---|---|
| **602** | — | `pixtuoid-scene/src/layout/compute.rs:83 compute_with_seed()` — longest in the workspace, 46% longer than `Reducer::apply` |
| **546** | **11** | `pixtuoid/src/tui/mod.rs:685 run_tui()` — deeper *and* 4.4× longer than the deepest fn the audit found |
| 486 | 8 | `pixtuoid-web/src/lib.rs:347 overlay_json()` |
| 480 | 7 | `pixtuoid-scene/src/pose/mod.rs:257 derive_with_routing()` |
| 348 | 8 | `pixtuoid-scene/src/pixel_painter/background/mod.rs:707 paint_floor_to_ceiling_window()` |

`run_tui` and `overlay_json` are churn ranks #5 and #7 respectively.

## Why it matters beyond size

The audit's headline structural claim was scope-mislabelled — "the deepest function in the tree" described only the audited subset. Any future reader of #825 would inherit that error. More usefully: the defect class that arc found (**a well-tested pure core joined to the OS by an untested one-line binder**) is not specific to `source/`; nothing has looked for it in the binary, scene, or web crates.

## Suggested scope for a follow-up pass

Per `.github/prompts/pr-review.prompt.md`'s whole-codebase orchestration, the subsystem cells not yet run:

- **scene**: painter / motion+layout / theme+misc
- **binary**: install / runtime+tui / widgets+floating
- **hook + web + tooling**
- **site + integrations/raycast** — the non-Rust consumers, with site pages **RENDERED**, not source-read (the prompt is explicit that this surface is structurally unaudited otherwise, per #453)

Modalities never run in the 2026-08 pass, per its own completeness critic:

- coverage measurement (every "untested" claim was reached by grep + reading; `codecov.yml` already defines five per-crate components and nobody pulled a number)
- unsafe/FFI audit (64 sites)
- public-API / semver-surface sweep
- dependency / supply-chain freshness

## Provenance

[...]

PR #457, opened 2026-07-03
Whole-codebase audit (post cleanup-arc-1): 14 fixes + baseline regen

IvanWng97:
Second whole-codebase re-review after cleanup-arc-1. A 21-finder adversarial audit surfaced 40 findings → 19 survived verification → **14 fixed** (7 correctness + 7 design-debt). Done on an isolated worktree, kept off the shared checkout a concurrent site session owned.

## Correctness bugs
| Sev | Where | Fix |
|-----|-------|-----|
| **HIGH** | `tui/mod.rs` | onboarding **SKIP no longer strips a pre-0.12 upgrader's installed hooks** — freeze to the real state (`connected \|\| has_hooks`) via a pure, tested `sources::freeze_for_skip`, never routing through install/uninstall |
| MED | `layout/approach.rs` | obstacle **render anchor == walk goal** — `stand_point` gains `approach_point`'s reachability filter (verifier probe found 136 pop cases); pinned by a real-layout equality test |
| MED | `tui/mod.rs` | a **connected source whose CLI vanished (`NoCli`) stays disconnectable** — carry the `connected` bit on `ConnectionRow` |
| LOW | `source/copilot.rs` | **Copilot `subagent_type` spoof closed** — name-only dispatch (`copilot_tool_detail`, mirroring Reasonix) |
| LOW | `pose/mod.rs` | **seated snap-back/exit departs from the seat** (no ~10px pop); entry/exit/snap-back measure the rendered polyline |
| LOW | `tui_renderer/mod.rs` | **floor-transition too-small gate** paints a footer-only frame + lands the transition (shared helper) |
| LOW | `floating/window.rs` | **daemon-only office animates at 30fps** (throttle consults `daemons`) |

## Design-debt (0.12.0 pre-release refactors)
- `cascade_exit` gains an explicit `StampRoot::{Yes,No}` param (the documented footgun materialized at its 4th caller).
- Coffee-machine hit-test pinned to a shared `PANTRY_COFFEE_COLS_*` painter const (+ steam cross-check test); small-case reconciled.
- `NIGHT_FLOOR_DIM_STRENGTH` named, vestigial `FloorMeta::sunlight_boost` dropped; `door_anim_max_ms` doc corrected.
- risk-radar `json-contract` + `ci-gates` seams cover the DTO home (`sources.rs`) + the gate scripts.

## Gates
- ✅ `just preflight` — lint → clippy → hack → **2041 tests, 0 warnings**
[...]

PR #862, opened 2026-08-04
refactor(tui): lift run_tui's key dispatch and two click predicates out

IvanWng97:
Refs #830

## Context

#830's point is that the 2026-08 audit covered 25% of the tree and the largest functions were never in scope. I re-ran its measurement live (posted to the issue) — every number had shifted under PR #845, but **`run_tui` held its rank as the deepest function in the workspace**: 458 lines at nesting depth 11.

The depth came from a `select!` → `loop` → `while polled` → `match event::read()` → `match dispatch_key(..)` stack.

## What

- **`apply_key_action(action, &mut KeyCtx) -> bool`** takes the 183-line arm block whole. It returns the **quit** decision — the one piece of control flow the event loop still owns — so no arm reaches back into the loop's locals. `KeyCtx` bundles the four independently-mutated pieces of state (`ui`, `renderer`, `audio_ctl`, plus borrowed config/scene) because the borrow checker needs them split at the call site, not inside.
- **`star_clicked`** and **`version_popup_url_clicked`** are structural twins: both were inline `terminal::size()` → build `Rect` → hit-test ladders four levels down inside the mouse arm. The popup one additionally re-spelled ratatui's `Rect::contains` as four hand-written comparisons; it now calls `contains`, like its sibling already did.

## Result

`run_tui`: **458 lines → 256**. The `KeyAction` arms move from depth 11 to depth 2 inside their own function.

The measurement tool still reports "nest 10" for `run_tui`, but the residual is `KeyCtx { .. }` struct-literal and argument-list braces, not control flow. Stating that plainly rather than quoting a flattering number.

## Honest limitation

Both new predicates call `crossterm::terminal::size()`, which is not injectable, so they remain **untested glue** — the same defect class #830 asks us to hunt. Extraction does not fix that. It makes the geometry decision nameable and gives the injection seam somewhere to land later.

## What's left in #830

Assessment posted to the issue. Short version:
- **`derive_with_routing`** (380 / nest 7) — a clean four-branch split, worth doing, but it is the motion core where an error shows as a visible sprite glitch, so it wants the `beautify-decoration` render-and-eyeball loop rather than test-suite trust alone.
- **`compute_with_seed`** (482 / nest 5) — **not** a good move-only target. Straight-line geometry with heavy variable threading; splitting it would invent intermediate structs per phase and scatter tightly-coupled geometry behind wide parameter lists. That fails the repo's own "would deleting this concentrate complexity or just move it?" deepening lens.
- The four unrun modalities (coverage measurement, unsafe/FFI over 64 sites, public-API/semver sweep, dependency freshness) are untouched.

## Verification

`just preflight` green — 2856 tests, exit 0.

https://claude.ai/code/session_01TDXH8k2NaWmrY6gYv25rBZ

### DIFF
--- a/crates/pixtuoid/src/audio/mod.rs
+++ b/crates/pixtuoid/src/audio/mod.rs
@@ -121,8 +121,10 @@ impl AudioController {
     }
 
     /// [`new`] with the boot-spawn injected, so a test can pin the boot decision
-    /// without opening an output device.
-    fn new_with(
+    /// without opening an output device. `pub(crate)` for the same reason it
+    /// exists: `tui`'s key-action tests need an UNMUTED controller, and an
+    /// unmuted `new` boot-spawns the real device.
+    pub(crate) fn new_with(
         muted: bool,
         volume: f32,
         config_path: std::path::PathBuf,
--- a/crates/pixtuoid/src/tui/CLAUDE.md
+++ b/crates/pixtuoid/src/tui/CLAUDE.md
@@ -34,7 +34,9 @@ tui/
 ├── mod.rs          crossterm event loop + terminal lifecycle (raw mode / alternate screen) + key
 │                   dispatch (the `w` debug-overlay toggle — dev-only, debug builds; `t` theme picker, `Tab` dashboard, `s`
 │                   Sources panel, floor nav, version popup, quit chord) + connect_source/disconnect_source.
-│                   run_tui is the EVENT LOOP + side effects only — the per-surface UI state lives in ui_state.rs
+│                   run_tui is the EVENT LOOP only — each KeyAction's side effects are applied by
+│                   apply_key_action(action, &mut KeyCtx) -> quit, which IS unit-tested (the loop is not);
+│                   the per-surface UI state lives in ui_state.rs
 ├── ui_state.rs     UiState — run_tui's per-surface UI state as ONE struct (onboarding/version/help/
 │                   pause/theme-picker/dashboard/connection + DriftScan): owns the open/close
 │                   transitions, projects the dispatch-facing ModalState (`UiState::modal` — the one
--- a/crates/pixtuoid/src/tui/mod.rs
+++ b/crates/pixtuoid/src/tui/mod.rs
@@ -560,6 +560,260 @@ pub(crate) struct TuiSession {
     pub first_run: bool,
 }
 
+/// Whether a left-click at `(col, row)` landed on the wall's star/repo link.
+///
+/// The caller gates on `renderer.cached_layout().is_some()` FIRST — the wall
+/// display only paints with a layout — so a too-small frame or a floor-slide
+/// transition can't phantom-launch a browser. Split out of `run_tui`'s mouse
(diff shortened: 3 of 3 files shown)


IvanWng97:
## Review disposition — every finding at a terminal state

Three local lenses (correctness / design / UX-journey) plus the online bot. The design lens returned **REQUEST-CHANGES**; all three of its must-fix items are addressed and verified at head `e20b20e7`.

### Must-fix (design lens)

| # | Finding | State | Verification |
|---|---|---|---|
| F5 | doc named a `select!` that does not enclose the arms | **FIXED** | `grep select!` → one hit, the real one at `:1043`. Confirmed the claim was wrong twice: `select!` is a SIBLING after `while polled`, both inside `loop`. |
| F6 | `KeyCtx` doc said "four" `&mut` (it's three) + false borrow-checker rationale | **FIXED** | doc now reads "three `&mut` surfaces … it exists so the arm list takes one argument instead of nine" |
| F1 | 9 mutants newly escaped `.cargo/mutants.toml`'s name-based filter | **FIXED via the TEST remedy**, which both config files prescribe over widening the list | 4 tests, negative-controlled: `apply_key_action -> true` reds 4/4, `-> false` reds 1, `delete !` reds 1, TogglePause swap reds 1 |

### Strongly recommended

- **F3** — both predicates now take terminal size as a parameter (`star_clicked(col,row,term)`, `version_popup_url_clicked(col,row,scale,term)`), the lens's exact signatures. They were not merely untested but **untestable**: under `cargo test` there is no tty, so both returned `false` unconditionally.
- **F2** — `respawn` injected into `KeyCtx`; `AudioController::new_with` is `pub(crate)` so a harness can build an UNMUTED controller without opening a device.

### Nits

**F4** (imperative MUST wording + the scene-rect-vs-full-bounds asymmetry), **F7** (inline measurements cut), **F10** (`tui/CLAUDE.md:37`) — all fixed. **F8** (`&mut KeyCtx` → by value) and **F9** (`Rect::new` at the two PRE-EXISTING sites) — not taken; F8 is neutral, F9 is outside this diff.

### Other lenses

- **Correctness** — APPROVE-WITH-NITS. Verified ratatui's `Rect::contains` against the pinned source (`ratatui-core-0.1.2/src/layout/rect.rs:355`): exactly the four comparisons replaced, with `saturating_add` making it strictly safer. All 35 arms mechanically diffed, zero behavioral difference.
- **UX journey** — APPROVE-WITH-NITS. All eight journeys PASS, including the three this lens exists to protect: R0619-02 (`apply_choices` outcomes still consumed), #457 (`skip_freeze` still reads rows before close), R0619-01 (version popup still dismissable for a no-CLI user).

### One correction to the record

The commit message's "honest limitation" paragraph covered the two *predicates* but was silent on the *applier* — which was the one the extraction actually made reachable. Both lenses caught that; it is now the thing with tests.

CI 47 pass / 0 fail. Bot auto-review **Findings: 0** and security review **Findings: 0**, both at `e20b20e7`.

Issue #832, opened 2026-08-03
[Tech-debt] 42 LOW audit findings from the 2026-08 source/+state/ pass — comment-rot, sibling-set gaps, DRY


IvanWng97:
## What

The whole-codebase audit behind #825 confirmed **57 findings** after adversarial verification (2 HIGH, 13 MEDIUM, 42 LOW). The HIGH/MEDIUM set was fixed in #825 or filed as #826-#831. This tracks the **LOW remainder** — individually small, collectively a real drift load, and each one already carries evidence.

By factor family: comment-rot 13 · sibling-set-completeness 8 · DRY 5 · misleading-identifier 4 · oversized-fn 3 · unnecessary-fallback 3 · mutation-teeth 2 · doc↔code 2.

Several are NOT cosmetic — a comment that is FALSE about the code is the shape that makes a real defect read as deliberate. Highlights:

- `source/jsonl/mod.rs:361` — `JsonlWatcher::run`'s doc promises it terminates when the event channel closes; **no arm, break or return observes closure**. The task spins its 60s scan forever after the reducer is gone.
- `source/copilot.rs:89` — `copilot_child_key`'s primary branch filters an empty string, the **sibling fallback does not**, so an empty `data.toolCallId` mints a child `AgentId` keyed on `""`.
- `source/claude_code.rs:546` — `cc_derive_label`'s project-dir fallback re-hand-rolls `cwd_basename_label`'s format and **drops its `ellipsize` cap**, so an uncapped label reaches `AgentSlot.label`.
- `source/cursor.rs:199` — the doc says the fn does NO subagent-dispatch detection; its **first branch is exactly that**.
- `state/reducer/mod.rs:1406` — `sweep_stale`'s doc still describes the `"cc#"` string-shape sniff that #221 deliberately replaced with `LabelProvenance`.
- `source/mod.rs:55` — `ToolDetail::Task`'s public doc names the `Task` tool arm dropped in 0.12.0; the recognised name is `Agent`, and both sibling docs say so.
- `source/jsonl/tests.rs` — **eight** test docs cite the code under test by LINE NUMBER; after the `jsonl/` directory split every citation is stale and now lands on a different, plausible-looking construct.
[...]

PR #867, opened 2026-08-04
test(source): triage the source/ + state/ mutation survivors — 77 measured, 58 real, 42 pinned (#828)

### LINKED ISSUE #828: [Test] Triage the 63 mutation survivors in source/ + state/ — none has had the documented reproduce-first step run
## What

A broad `cargo-mutants` pass over `crates/pixtuoid-core/src/source/**` + `src/state/**` (excluding `hook/` and `jsonl/`, which #825 drove to zero) scored **1040 of 1115 mutants (93%)** before being stopped:

| outcome | count |
|---|---|
| caught | 855 |
| **missed** | **63** |
| timeout | 5 |
| unviable | 124 |

The 63 survivors have **not** been triaged, and — importantly — **not one has had the repo's own documented procedure run against it**.

## Why triage can't be done by reading

`.cargo/mutants.toml:8-20` states that package-scoped mutation is a *systematic blind spot* for `pixtuoid-core`, because core behaviour is covered on purpose by the **binary's** headless harness, and instructs:

> for a survivor in core/scene, REPRODUCE IT FIRST — apply the mutation and run `-p pixtuoid` too. Only write a new test if the mutation genuinely survives the whole workspace.

Every one of these 63 is package-scoped. The same file names `SceneState::insert_daemon` as the worked example: making it a no-op reds 8 of 9 `harness::mascot` tests under `-p pixtuoid` while all 53 of core's own daemon tests stay green — **and it still reports as a survivor**.

So the `state/mod.rs` cluster below is likely false-positive, but "likely" is exactly the word the procedure exists to eliminate. They are **unmeasured**, not dismissed.

## Where they are

| file | missed |
|---|---|
| `source/decoder.rs` | 19 |
| `state/mod.rs` | 9 |
| `source/grok.rs` | 8 |
| `source/mod.rs` | 7 |
| `source/proc_start.rs` | 5 |
| `source/codex/native.rs` | 3 |
| `source/{omp,hermes,copilot/native,claude_code/native,antigravity/native}.rs` | 2 each |
| `source/copilot.rs`, `source/acp.rs` | 1 each |

Known leads already characterised:

- `source/mod.rs` — `cc_registry_dir`, `codex_pid_for_session`, `grok_pid_for_session` all survive `-> None`. Same **untested one-line binder** class #825 fixed at three sites; these are instances 4-6.
- `source/acp.rs:102` — `replace match guard !t.is_empty() with true` in `decode_session_update`.
- `source/copilot.rs:367` — `replace > with >=` in `decode_copilot_line`.

## Suggested approach

1. Re-run the scope cleanly on an idle machine (timing-sensitive classification needs it — a loaded box produced 9 bogus timeouts during the #825 work, the tell being a *provably equivalent* mutant "timing out").
2. For each survivor, reproduce with `--test-workspace true` / `-p pixtuoid` **before** writing anything.
3. Then drive each to one terminal state: killed by a new test, or pinned in `exclude_re` with a *verified* equivalence proof — never a prose assertion (`.cargo/mutants.toml`'s own header explains why).

## Provenance

Measured during the #825 arc. Raw outcome files were preserved from the run.

IvanWng97:
## Summary

Triages the `source/` + `state/` mutation survivors with the reproduce-first step
`.cargo/mutants.toml` mandates but that had never been run against them. **58 of
the 77 survivors are real gaps; 19 were killed by the binary's harness all
along.** 42 real gaps get a test, 16 are measured noise recorded with their
derivations.

Three of the issue's own numbers turned out to be wrong, and all three are
measurements, not opinions:

| | issue | measured |
|---|---|---|
| survivors in scope | 63 | **77** |
| `state/mod.rs` false positives | "likely" all 9 | **6 of 9** |
| largest false-positive group | not mentioned | **`Source::name`, 12 across six `native.rs`** |

The 14-survivor gap is exactly two files the issue's table never lists —
`omp/native.rs` (8) and `grok/native.rs` (6). The original run stopped at
1040/1115, and those 75 unscored mutants were these two files.

## Related issue

Closes #828.

## Type of change

- [x] Refactor / chore — test hardening; **no production code changed** (12
      files, +464/-1: tests, `.cargo/mutants.toml`, docs — the single deletion is
      a stale doc line)

## What the 58 became

**42 killed by new tests**, each re-measured after the fact rather than asserted:

- `decoder.rs` **19/19** — all in `rfc3339_to_epoch_secs`, whose only pins were
  well-formed shapes plus four garbage strings. Adds the separator bytes at
  4/7/10 (incl. the lower-case `t`), the day/hour/minute/second bounds,
  `Z`-must-be-final, the exact `±HH:MM` grammar, and a NEGATIVE offset with
  nonzero MINUTES. Re-scored 117/117.
- one-line binders **6/7** — `cc_registry_dir` and `grok_pid_for_session` had no
  test reaching a HIT; the #825 class, instances 4-6.
- `grok.rs` **6/7**, `omp/native.rs` **3/3**, `state/mod.rs` **3/3**,
  `acp.rs`/`copilot.rs`/`omp.rs` **4/4**.
- `GrokSource::run` **1/1** — `tests/watcher/sources.rs` drives the `run` of the
  other five transcript sources; grok was the one missing, so mutating its body
  to `Ok(())` survived the whole workspace. On description alone this looks like
  the `runtime/driver.rs` "no test call site" class already pinned in
  `.cargo/mutants.toml`; excluding it would have hidden a real gap in the source
  with the most complex `run` of the six.

**16 measured noise** — 10 excluded, 6 documented as known-noise rows. Every
exclusion names the exact replacement VALUE, not the function, because each of
these fns has a killable sibling under the same name. The 6 that get no
exclusion are cfg twins whose non-compiled variant cargo-mutants describes
IDENTICALLY to the compiled one (`proc_start::imp`, `live_grok_session_ids`), so
any regex would also hide a currently-caught mutant — the `mascot_position` call,
and this file's own rule that such an exclusion is worse than a noisy row.

## Two reusable facts, now in `docs/mutants-exclusions.md`

- **`--timeout-multiplier` must clear nextest's 180s `terminate-after`.** Below
  it, cargo-mutants' own timeout fires first and files a genuine infinite-loop
  detection as a `timeout` row instead of a named test failure.
- **CPU contention can only produce false timeouts or false CAUGHTs — never a
  false pass.** So `missed` verdicts are load-robust and only `caught`/`timeout`
  need a quiet re-run. A second mutation run sharing the machine produced 20
  bogus timeouts here; all 20 resolved to `missed` when re-run idle, and the 19
  `caught` verdicts came back byte-identical.

## How I tested it

- `cargo nextest run -p pixtuoid-core` — 1043/1043 green
- `just preflight` — exit 0: lint (13 checks), clippy `-D warnings`, hack
  (`--no-default-features` across 14 crates), 2879 tests
- Full-scope mutation re-run against the merged tree: **1106 mutants in 3h — 24
  missed, 953 caught, 129 unviable, 0 timeouts.** All 24 audited against the
  ledger: 18 known false positives + 6 documented noise rows, **0 unexpected and
  0 regressions** (every one of the 42 killed gaps stays dead)

## Visual verification

- [x] Not a visual change.

## Checklist

- [x] New behavior has a test — this PR *is* tests.
- [x] No `unwrap()` in non-test code; no non-test code changed.
- [x] Docs updated in the same commit (`docs/mutants-exclusions.md`, and
      `crates/pixtuoid-core/tests/CLAUDE.md` whose tree listing enumerated five
      sources for `sources.rs` — grok's absence there is exactly why its mutant
      survived).
- [x] Checked against the recurring pitfalls. #2 (parallel-implementation drift)
      is the one this PR turns on: after finding grok missing from
      `tests/watcher/sources.rs` I grepped every `impl Source for` — 6 of 6
      transcript sources are now driven there; the 7th, `HookRouter`, is in the
      `hook/` subtree this issue excludes and #825 drove to zero.

## AI assistance


IvanWng97:
## Picks up the two mutation-teeth items #832 handed over

[#832's closing comment](https://github.com/IvanWng97/pixtuoid/issues/832#issuecomment-) routed two LOW findings here rather than fixing them, on the grounds that they are #828's subject. Both are covered:

| #832 finding | pinned by |
|---|---|
| `copilot.rs` — `session.shutdown` mutation teeth | `a_shutdown_with_no_fresh_tokens_emits_no_usage_reading` (`copilot.rs:374`) |
| `acp.rs:102` — the `!t.is_empty()` match guard | `an_absent_or_empty_session_update_tag_never_breadcrumbs` (`acp.rs:140`) |

Both surfaced independently as confirmed real gaps in the Phase-B (`--test-workspace true`) pass, so they are measured here rather than taken on #832's word — the conclusions happen to agree.

## Rebased onto #861

The branch now includes #861 and #865. Worth calling out because #861 touched four files this PR carries exclusions or doc claims about — `hermes.rs`, `copilot.rs`, `grok/native.rs`, `source/mod.rs`:

- `resolve_hermes_home` still has **4 in-test call sites**, so the derivation recorded in `docs/mutants-exclusions.md` still holds. #861's change there is a one-line swap to `platform::nonempty`; the wrapper/`_resolved` split the exclusion rests on is unchanged.
- Suite green at 1046/1046.
- A targeted mutation re-measure of the 141 mutants in those files is running to confirm #861's new code introduces no survivor and no exclusion went stale. Result to follow.

IvanWng97:
## The #861 rebase surfaced two fresh survivors — both now pinned

The targeted re-measure of the 141 mutants in the files #861 touched came back with **5 survivors, 2 of which were new**. Both sat in code #861 had landed hours earlier, and neither was reachable by any existing test:

| new survivor | what was unpinned |
|---|---|
| `grok/native.rs:18` `MAX_SESSION_REGISTRY_BYTES = 1024 * 1024`, `*` → `+` | the cap's MAGNITUDE — a 2 KiB cap passed every test |
| `grok/native.rs:50` `bytes.len() == MAX_…`, `==` → `!=` | WHICH cause the drift breadcrumb names |

Pinned in `fbc9fd64`; `grok/native.rs` is back to 3 survivors, all pre-accounted (the `cfg(not(unix))` stub + the two `Source::name` false positives).

### One thing worth a maintainer's eye, not a silent fix

Writing the second test surfaced a problem in the diagnostic itself. `shape_drift` caps its detail at `MAX_DECODED_FIELD_CHARS` (**80**), and the message spends 24 chars on its prefix **before** interpolating the full path:

```
"active_sessions.json at {path} {cause}; liveness degraded to mtime gating"
```

So the cause is only readable while the path stays short. The default macOS `$TMPDIR` is ~55 chars on its own, which truncates the cause away **entirely** — the first version of this test failed for exactly that reason, and it only passes now because it roots its tempdir in `/tmp`.

For a real `~/.grok` (~42 chars) about 14 characters of cause survive — just enough to tell `was TRUNCATED ` from `does not parse`, but any deeper grok home loses the distinction the branch exists to draw. Left as-is and documented on the test rather than changed here: it is #861's code and reordering the message is that author's call.

### Unrelated pre-existing note

`first_sight::watcher_derives_parent_id_for_subagent_path` reports `LEAK` under nextest. It is on `main` and untouched by this PR — flagging only so it is not read as a regression from these tests.

Issue #866, opened 2026-08-04
flaky: shim env-mode test missed nextest's exclusivity filter — the suffix match covered 2 of its 3 siblings


IvanWng97:
## Problem

`just preflight` (and `just test`) fail on macOS with one test, on `origin/main` as well as on any branch:

```
FAIL [ 2.03s] pixtuoid-hook::shim codewhale_event_mode_builds_envelope_from_env_and_ignores_stdin
env-mode shim did not exit within 2s — it must NOT read stdin
error: recipe `test` failed with exit code 100
```

CI is green (`ci: success` on main @ 5ddc09c0), so this is local-only — the documented pre-push gate cannot currently be used as a gate on macOS.

## The discriminator

The axis is **single-crate vs workspace**, not parallelism and not load:

| command | result |
|---|---|
| `cargo nextest run -p pixtuoid-hook` | 32/32 pass — on `origin/main` AND on a feature branch |
| `cargo nextest run --workspace` | this one test fails, reproducibly |
| `cargo nextest run --workspace --test-threads 4` | still fails — rules out saturation |
| `cargo test --workspace` (libtest, not nextest) | passes |

Two hypotheses were tested and REFUTED: concurrent CPU load from another gate running alongside, and nextest parallelism starving the 2s budget. Reducing to 4 threads still fails, and the single-crate nextest run passes at full parallelism.

## Why it is not a branch's fault

Reproduced on a clean `origin/main` worktree: `MAIN_FULL_EXIT=100`, same test, same 2.036s, 2865 tests. Confirmed independently of any feature-graph change.

## Fix sketch

The test asserts the env-mode shim exits without reading stdin, and the 2s budget is a proxy for "did not block on stdin". Something about a workspace-wide nextest run changes what the spawned shim inherits as stdin, in a way a single-crate run does not. Worth checking what nextest hands a test process for stdin under each invocation, and whether the assertion can be made on the shim's actual fd behaviour rather than on a wall-clock timeout — a timing proxy is what makes this environment-sensitive in the first place.

Environment: macOS (Darwin 27.0), nextest via `just test`.

IvanWng97:
## Correction — the original report's central claim is WRONG

I wrote this up as reproducible. It is not. Post-merge the same suite passes:

| run | result |
|---|---|
| pre-push `just preflight` (full gate) | 2928/2928 passed |
| `cargo nextest run --workspace` ×3 | 2928/2928 passed, each time |

[...]

Issue #161, opened 2026-06-07
flaky: shim missing_socket_exits_zero_without_blocking exceeds its latency bound under full-suite load


IvanWng97:
## Problem

`pixtuoid-hook::shim missing_socket_exits_zero_without_blocking` panicked ("shim must not block when the socket is absent") during a full `just preflight` run on clean main (d3320af), then passed 3/3 consecutive runs in isolation. The test asserts a wall-clock latency bound on the shim's exit; under the fully parallel 1052-test suite on a loaded machine the bound is exceeded — a scheduler artifact, not a shim regression (the shim's 200ms WRITE_TIMEOUT contract is socket-write-bound, and there was no socket to write to).

## Why deferred

Surfaced mid-release-bump; not blocking (re-run passes). Needs a deliberate choice about the right fix shape.

## Fix sketch

Either widen the latency bound with a comment tying it to suite-load reality, or measure the shim's elapsed time around the *process* spawn only (excluding test-harness setup), or mark the test as requiring low parallelism (nextest threads-required override). Don't delete the bound — it pins the "never block CC" contract (#5 root invariant).

Observed: 2026-06-07, `just preflight` during `just bump 0.6.0`, exit 100 at 1051/1052.

PR #869, opened 2026-08-04
fix(test): give the shim env-mode test the exclusivity its two siblings have

IvanWng97:
Refs #866. **Net change vs `main` is one line.**

```diff
-filter = 'test(_exits_zero_without_blocking) | test(_within_watchdog_bound)'
+filter = 'test(_exits_zero_without_blocking) | test(_within_watchdog_bound) | test(_and_ignores_stdin)'
```

## The gap

`.config/nextest.toml` already grants `threads-required = "num-cpus"` to the tests that spawn the shim and poll a wall-clock deadline, because that class flaked before — **#161**, at a 1s bound. The filter matched by **suffix**, and `codewhale_event_mode_builds_envelope_from_env_and_ignores_stdin` matches neither `_exits_zero_without_blocking` nor `_within_watchdog_bound`. So it was the one test of that shape running unprotected.

That is also why #866 only ever saw it under `--workspace`: a 32-test `-p pixtuoid-hook` run has nobody to contend with; a 2865-test workspace run does.

## Why additive, and why the history on this branch is messy

The first attempt (`b91e0f4f`) rewrote the filter as an explicit **name list** and added a 25-line roster-pinning test. Review caught a **HIGH**, correctly: the substring form had *also* been covering `shim_pipe.rs`'s two `cfg(windows)` tests — `missing_pipe_exits_zero_without_blocking` and `stalled_daemon_shim_exits_zero_within_watchdog_bound` — and the three-entry name list silently dropped both. **The same defect this branch exists to fix, reintroduced in the other direction.** The pin couldn't catch it either: `file!()` resolves to `shim.rs`, so it only ever read half the population.

`378a3e11` reverts all of that. Appending one alternation **cannot** remove coverage — on unix the match set goes 2 → 3, and the two original alternations are byte-identical, so the Windows pair is untouched by construction. That property is exactly what the rewrite gave up.

## Not raising the 2s budget

This file's own note already answers it — raising it *"would only lengthen the failure path — they exit on a positive signal, so exclusivity is nearly free."* Applies here identically.

`threads-required = "num-cpus"` verified against nextest's docs rather than inferred from the repo comment: it reserves every logical CPU for the test, *"effectively giving it exclusive use of the machine."*

## Honest limitation: no reproduction

| attempt | result |
|---|---|
| full workspace @ `5ddc09c0` (the sha #866 reported) | 2865 pass, exit 0 |
| full workspace @ `main`, ×3 | 2868 pass, exit 0 |
| the test alone, ×5 | 0.07–0.15s vs a 2s budget |
| under 20 CPU spin-loops (load 16) | 0.075–0.142s |
| under fork/exec storm (**load 107**) | 0.084–0.149s, 0/6 fail |

The justification is the missing override, which is verifiable directly from the config; **#161 is the evidence that the class bites.** Not a repro-driven patch, and I'd rather say so than imply otherwise.

Note CI green is weak evidence here either way: what was missing is flake-protection, not correctness, so every job passes with or without it. That blindness is why the gap survived.

## Verification

`just clippy` exit 0 · full workspace 2887 passed.

https://claude.ai/code/session_01TDXH8k2NaWmrY6gYv25rBZ

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -15,8 +15,14 @@ path = "junit.xml"
 # These assert wall-clock bounds on a CHILD PROCESS's spawn+exit; under the
 # fully-parallel suite on a loaded machine, exec/dyld scheduling jitter alone
 # blows the bound, so give them the machine to themselves.
+# Listed by NAME, not by suffix: the shape is "spawns the shim and polls a
+# deadline", which no name pattern infers — a suffix filter silently missed one.
 [[profile.default.overrides]]
-filter = 'test(_exits_zero_without_blocking) | test(_within_watchdog_bound)'
+filter = '''
+  test(missing_socket_exits_zero_without_blocking)
+  | test(stalled_listener_shim_exits_zero_within_watchdog_bound)
+  | test(codewhale_event_mode_builds_envelope_from_env_and_ignores_stdin)
+'''
 threads-required = "num-cpus"
 
 # Same load-sensitivity as the shim bounds above: this pair spawns a real
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -15,14 +15,8 @@ path = "junit.xml"
 # These assert wall-clock bounds on a CHILD PROCESS's spawn+exit; under the
 # fully-parallel suite on a loaded machine, exec/dyld scheduling jitter alone
 # blows the bound, so give them the machine to themselves.
-# Listed by NAME, not by suffix: the shape is "spawns the shim and polls a
-# deadline", which no name pattern infers — a suffix filter silently missed one.
 [[profile.default.overrides]]
-filter = '''
-  test(missing_socket_exits_zero_without_blocking)
-  | test(stalled_listener_shim_exits_zero_within_watchdog_bound)
-  | test(codewhale_event_mode_builds_envelope_from_env_and_ignores_stdin)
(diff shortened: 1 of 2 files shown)


IvanWng97:
Closing without merging.

The gap is real — `_and_ignores_stdin` is the one child-process-timing test the suffix filter never covered — but I could not reproduce the failure anywhere: green at the reported sha `5ddc09c0`, green ×3 on `main`, and the test held at 0.07–0.15s against its 2s budget under 20 CPU spin-loops and under a fork/exec storm at load 107. No repro, no evidence it bites today.

Against that, this PR's own cost was already negative: rewriting the suffix filter into a name list **silently dropped the two `cfg(windows)` tests** (`missing_pipe_exits_zero_without_blocking`, `stalled_daemon_shim_exits_zero_within_watchdog_bound`) that the substring form had covered — the exact defect this PR set out to fix, reintroduced in the other direction, and caught only by review.

Note the CI green on `b91e0f4f` is not evidence of absence: what was dropped is flake-protection (`threads-required`), not correctness, so every job passes either way. That is the same blindness that let the original gap persist.

If this recurs, the minimal fix is one additive alternation — `| test(_and_ignores_stdin)` — which cannot drop existing coverage, rather than the rewrite attempted here.

PR #876, opened 2026-08-05
fix(gen): gate the wasm on GZIP bytes, the size a user actually downloads

### LINKED ISSUE #871: gen-wasm-check caps RAW bytes while Pages serves gzip — and the raw cap now blocks density art
`justfile:847` sets `CAP=1048576` against the **raw** `.wasm`. GitHub Pages
serves it **gzipped**. Measured on `34695274`:

| metric | size | vs 1 MiB |
|---|---|---|
| raw | 1,006,437 B (983 KB) | **96.0%** — 41 KB left |
| gzip (what Pages sends) | 394,829 B (386 KB) | 37.7% — 638 KB left |

This stopped being cosmetic when #868 landed the `<base>@<N>x` density
contract. Adding `@4x` variants for the furniture whitelist costs ~55 KB of
raw sprite text (measured: all base sprite GRID bytes = 9,160 B, furniture
subset 3,532 B, ×16 for 4x area) against 41 KB of headroom. **The second
variant blows the cap.** The same content gzipped is ~15 KB — sprite text
compresses to 27% (24,211 B → 6,596 B measured over the bundled pack).

So the gate blocks the art workstream on a metric no user ever downloads.

Fix: measure gzip. Budget is the owner's call — 400 KB strict / 430 KB
buffered / report-only. Note the raw number is not useless (it is parse +
compile cost in the browser), so keeping it as a *reported* second number
rather than deleting it is reasonable.

Found while sizing the art workstream after #868.

### LINKED ISSUE #877: justfile recipe bodies are outside shellcheck AND untested — a reintroduced silent-pass has nothing to catch it
Surfaced by the correctness lens on #876, confidence 90. Regression surface,
not a live defect — the current code is sound.

`SHELL_SOURCES := "scripts/*.sh .githooks/* policy/ci-observability/*.sh"`
(`justfile:33`) excludes the justfile itself, so `just shellcheck` and
`just shfmt-check` never see recipe bodies. There is also no test over any
recipe. So a shell defect inside a recipe rests entirely on review.

That is not hypothetical. #876's first draft wrote:

    WIRE=$(gzip -9 -c "$W" | wc -c)

POSIX sh has no `pipefail`, so the pipeline reports `wc`'s status: a broken
gzip measures zero bytes and PASSES the cap. Reproduced — the pipe form
prints `wasm 0 / 524288 bytes gzipped (0% of cap)` and exits **0**. The gate
goes green exactly when its own measurement tool breaks. It was caught by
reading, before merge, and the fix (compress to a `mktemp` file) is one a
future edit could "tidy" straight back into a pipe.

The repo's own idiom is right there — `compare-selftest` and `drift-selftest`
both exist and are wired into `lint`. Neither transfers cleanly: both hang a
`--selftest` flag on a Python script, whereas this is inline justfile shell,
so matching the idiom means parameterizing a recipe CI depends on.

Options, roughly by cost:
1. Extend `SHELL_SOURCES` to cover extracted recipe bodies (the
   `policy/ci-observability/action_behavior_test.sh` harness already extracts
   workflow step scripts with `yq` — the same shape would work for `just
   --evaluate`/`--dump`).
2. A `gen-wasm-check` selftest recipe that drives the gate with a stub `gzip`
   and asserts non-zero, plus `CAP` low and asserts non-zero.
3. Accept and rely on the WHY comment at the `mktemp` (which names the trap
   directly above the code that avoids it, so a reverting edit has to delete
   a warning — visible in any diff).

Deliberately deferred from #876 rather than dropped: the fix is a testing-
infrastructure change, out of scope for a gate-metric fix, and option 3 is
already in place.

IvanWng97:
`gen-wasm-check` capped the RAW `.wasm` at 1 MiB. GitHub Pages serves it
gzipped, so the gate was budgeting a number nobody downloads.

Measured against the live site, not estimated — the deployed etag `f5b65`
= 1,006,437 raw confirms Pages is serving the `34695274` build, so these
compare like for like:

| | bytes |
|---|---|
| Pages actually sends (gzip) | **398,811** |
| local `gzip -6` | 395,618 |
| local `gzip -9` | 394,829 |
| raw — what the gate capped | 1,006,437 |

## Why now

This stopped being cosmetic when #868 landed the `<base>@<N>x` density
contract. A full density pass over the 16-piece **furniture subset** costs
~55 KiB of raw sprite text (3,532 grid bytes × 16 for 4× area) against
**41 KB** of raw headroom — the *second* variant reds the gate. The same
content is a few KiB over the wire. The gate was blocking an entire
workstream on the wrong metric.

## The cap

**512 KiB.** Not 400 KB, which I proposed first and my own numbers kill:
389 KB today + the art phase ≈ 401 KB, red on the first complete landing.
Across *every* sprite at 4× the ceiling is ~407 KiB (9,160 grid bytes × 16
= 143 KiB raw, at the **measured 14.6%** compression — `desk@4x` is
4,266 raw for 624 B marginal inside the wasm). That leaves ~105 KiB spare,
and still reds on a debug build or a skipped `wasm-opt`, which are MB-scale.

Raw is now **reported, not gated** — and the review found the reason that
makes this safe, which was in the tree but cited nowhere: `site/lighthouserc.json`
asserts `total-blocking-time` and `user-timings:pixtuoid-revealed` at
`error` level, and `site.yml` fires on `site/**` where the wasm lives. Parse/
compile cost is **measured on the runner**, not proxied by a byte count.

## Review

Four lenses, all APPROVE-WITH-NITS. Dispositions in the thread. The three
that changed the diff:

- **Deleting the raw gate is not a coverage loss** — red-teamed and refuted
  with numbers, so no second cap was added.
- **The cap is deliberately loose and nothing said so**, meaning nobody
  would ever tighten it. The declaration block now names the retirement
  trigger and procedure.
- **The reading key I added for the raw ratio was inverted** — caught in the
  review round I was applying, and fixed before merge.

Also unwritten: three claims I had asserted without earning them (a cause I
never established, an inline number that rots, and a superlative I never
measured).

Verified after every round: gate green, `CAP=1024` reds, a failing `gzip`
stub reds, and the `trap` leaves zero files behind on both the success and
over-cap exit paths. `just preflight` green, 2947 tests.

Closes #871
Closes #877

### DIFF
--- a/crates/pixtuoid-scene/CLAUDE.md
+++ b/crates/pixtuoid-scene/CLAUDE.md
@@ -347,11 +347,15 @@ src/                (the pixtuoid-scene crate root; default pack at ../sprites/d
 │                   CAN SELECT IT — `pixtuoid-web` paints at `RenderScale::ONE` by design (the
 │                   chunky look), and only `densest_art` reads variants. Budgeting the art
 │                   phase: `gen-wasm-check` gates the GZIP size, because raw and served bytes
-│                   disagree by ~2.5x on this payload and sprite text is the most compressible
-│                   part of it — a variant that looks expensive on disk is cheap over the wire.
-│                   Run the recipe for the live numbers rather than trusting a figure written
-│                   here; the cap and its reasoning are in the justfile above it. GitHub Pages
-│                   serves gzip and does not offer brotli.
+│                   disagree by more than 2x on this payload and sprite text compresses far
+│                   better than the wasm code around it — a variant that looks expensive on
+│                   disk is cheap over the wire.
+│                   Run the recipe for the live raw/gzip figures rather than trusting one
+│                   written here; the cap and its reasoning are in the justfile above it. What
+│                   the CDN actually transfers is a SEPARATE check the recipe cannot make —
+│                   `curl -sIL -H 'Accept-Encoding: gzip, br, zstd'` answered gzip as of
+│                   2026-08, with brotli an open GitHub feature request, not a published
+│                   contract; if it ever lands, this gate becomes a ~20% overestimate.
 │                   Also: `pack.toml` declares the variant, so dropping the `include_str!`
 │                   alone makes `build_pack` fail on a missing frame — the loader would have to
 │                   skip variant animations too.
--- a/crates/pixtuoid-scene/CLAUDE.md
+++ b/crates/pixtuoid-scene/CLAUDE.md
@@ -345,14 +345,13 @@ src/                (the pixtuoid-scene crate root; default pack at ../sprites/d
 │                   `the_drawn_size_is_the_same_whichever_density_the_art_came_from`).
 │                   VARIANT ART IS EMBEDDED FOR EVERY TARGET, INCLUDING WASM, WHERE NOTHING
 │                   CAN SELECT IT — `pixtuoid-web` paints at `RenderScale::ONE` by design (the
-│                   chunky look), and only `densest_art` reads variants. Numbers, measured
-│                   2026-08-04, for whoever budgets the art phase: `gen-wasm-check`'s 1 MiB is
-│                   a RAW cap standing in for ~350-400 KB gzipped (its rationale is in the
-│                   justfile above the recipe). The artifact is 1,006,260 raw / 394,722 gzip
-│                   locally; GitHub Pages serves it gzip at 390,240 and does not offer brotli.
-│                   Sprite files are palette-key text: `desk@4x` is 4,087 raw, 518 gzip, 388
-│                   brotli. So ~27 further pieces are ~110 KB raw, ~14 KB gzip. Note only that
-│                   raw and served-bytes disagree by an order of magnitude on this payload.
+│                   chunky look), and only `densest_art` reads variants. Budgeting the art
+│                   phase: `gen-wasm-check` gates the GZIP size, because raw and served bytes
+│                   disagree by ~2.5x on this payload and sprite text is the most compressible
+│                   part of it — a variant that looks expensive on disk is cheap over the wire.
+│                   Run the recipe for the live numbers rather than trusting a figure written
+│                   here; the cap and its reasoning are in the justfile above it. GitHub Pages
+│                   serves gzip and does not offer brotli.
 │                   Also: `pack.toml` declares the variant, so dropping the `include_str!`
 │                   alone makes `build_pack` fail on a missing frame — the loader would have to
 │                   skip variant animations too.
--- a/justfile
+++ b/justfile
@@ -897,7 +897,7 @@ gen-wasm-check:
 # + a release build of the snapshot example.
 [group('gen')]
 [doc('Fail if any committed README section or rendered image has drifted')]
-gen-check: compare-selftest gen-readme-check gen-wasm-check
+gen-check: compare-selftest wasm-check-selftest gen-readme-check gen-wasm-check
     #!/usr/bin/env sh
     set -eu
     test -x .venv/bin/python3 || { echo "needs the venv: python3 -m venv .venv && .venv/bin/pip install -r requirements-dev.txt"; exit 1; }
@@ -1116,6 +1116,33 @@ setup-tools:
     # would otherwise be the only gate). Idempotent. CI re-runs `just preflight`
     # regardless, so a skipped local hook still meets the same checks at merge.
     git config core.hooksPath .githooks
+
+# The size gate's own negative control, because nothing else can be one: the
+# justfile is outside SHELL_SOURCES, so shellcheck never reads a recipe body, and
+# a size cap that stops measuring reports success for any artifact at all. This
+# pins the FAIL-OPEN class specifically — the first draft of the gzip gate wrote
+# `gzip … | wc -c`, which under POSIX sh (no pipefail) reports wc's status, so a
+# broken gzip measured zero and PASSED. Driving the real recipe with a gzip that
+# exits 1 is what that form cannot survive.
+# Not covered, deliberately: the over-cap and empty-artifact arms, which would
+# have to mutate the committed wasm to exercise. Their failures are loud; the
+# fail-open one is the silent class worth a test.
+[group('meta')]
+[doc('Self-test the wasm size gate: prove it still reds when its measurement breaks')]
+wasm-check-selftest:
+    #!/usr/bin/env sh
+    set -eu
+    stub=$(mktemp -d)
+    trap 'rm -rf "$stub"' EXIT
+    printf '#!/bin/sh\nexit 1\n' > "$stub/gzip"
+    chmod +x "$stub/gzip"
+    if PATH="$stub:$PATH" just gen-wasm-check >/dev/null 2>&1; then
+        echo "wasm-check-selftest: FAIL — gen-wasm-check passed with a broken gzip;"
+        echo "  the size measurement is fail-OPEN. Did the gzip call become a pipe?"
+        exit 1
+    fi
+    just gen-wasm-check >/dev/null
+    echo "wasm-check-selftest: OK (reds on a broken measurement, greens on a real one)"
 
 # The pixel comparator is the primitive under `gen-check` and the smoke job, and
 # it had no test of its own — an always-green comparator reports success for any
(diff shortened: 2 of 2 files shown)


IvanWng97:
## Review round — 4 lenses, disposition

Two lenses delivered full reports (best-practice, design), both **APPROVE-WITH-NITS**; correctness + comment lenses are re-running after returning nothing on their first pass. Every finding below has a terminal state.

### FIXED

| # | finding | fix |
|---|---|---|
| F1 / N1 | 512 KiB is deliberately loose (129 KiB slack; the art phase needs ~21 KiB of it) and nothing recorded that the looseness is temporary — nobody would ever tighten it. Also: the comment called raw "a real constraint" and then gated it at infinity | Declaration block now names the retirement trigger + procedure, and cross-references `site/lighthouserc.json` — raw is **already** hard-gated there (`total-blocking-time` and `user-timings:pixtuoid-revealed` are `error`-level, and `site.yml` fires on `site/**` where the wasm lives). Verified all three myself. A byte-count proxy would be the weaker instrument; that reasoning existed nowhere in the tree |
| F3 | Redundancy-within: the gzip/#871 rationale appeared in BOTH the declaration block and the new body block, 30 lines apart. "3 sentences, ~1.5 ideas" | Body block collapsed 8 lines → 2; only the unique idea survives |
| F4 | `$RAW` printed bare, with no reference point — against this recipe's own adjacent argument that a number you can only read when it breaks gives no warning | Now prints its ratio. Falling = new poorly-compressible code; rising = new sprite text |
| F5 | "Run the recipe for the live numbers" over-promises: the recipe gzips locally and never touches the CDN, so it cannot substantiate the two Pages/brotli claims that follow it | Split into "run the recipe for raw/gzip" vs "the CDN is a separate check", with the actual curl |
| N3 | Pages-serves-gzip-no-brotli stated as a standing fact. It is undocumented third-party behavior + a 2019 GitHub feature request still open in 2024 | Time-qualified, with the consequence noted (~20% overestimate if brotli lands) |
| — | Two things I introduced myself: `(different implementation and level)` was a cause I never established (I had measured the gap, not its origin), and "roughly a third of the wire size" is an inline number that rots as the payload grows — while citing the rule against exactly that | Both removed |

### REFUTED, with trace

- **N2** — `ci-builds.yml:27` "the size cap IS deterministic". The two lenses disagreed; the one that **measured** it (21-byte spread between gzip implementations, against 126 KiB headroom) called it not a finding. The comment's claim is about BUILD non-determinism, which this diff does not touch. Left as-is.
- **F6, first half** — "~55 KB raw for a furniture density pass" was disputed as ~78 KiB. It **reproduces exactly** at 3,532 grid bytes × 16 = 55 KiB; the discrepancy is a grouping difference (`OPTIONAL_FURNITURE_ANIMATIONS` includes the cat/dog/lobster pet animations, which I had excluded as not-furniture). Not an arithmetic error — an under-specified scope. Changing a correct number would have introduced the error.

### CORRECTED (my numbers were wrong)

- **F2, confidence 88, and it is right.** I assumed 27% compression for sprite text; measured against the one real hand-authored variant it is **14.6%** (`desk@4x`: 4,266 raw → 624 bytes marginal gzip cost inside the wasm). So the full density pass costs ~21 KiB gzip, not ~39 KiB — the cap is looser than my own reasoning claimed. Error was in the safe direction (over-budgets, nothing reds unexpectedly), but the recorded chain was wrong and F1's text now reflects it.

### Explicitly checked clean by the design lens

Raw-gate deletion is **not** a coverage regression — do not add a cap back. A debug build carries DWARF + the name section and blows a gzip cap by an order of magnitude; the compressible-bloat scenario needs ~870 KB of hand-added raw to matter. No stale raw-cap prose anywhere, no consumer parses this recipe's stdout, no duplicate payload-budget policy, layering N/A.

### On the owner's question — "is this best practice?"

Verified online rather than asserted: bundlesize (gzip), size-limit (brotli), Lighthouse CI (`transferSize`) all budget **compressed** size; `gzip-size` hardcodes `level: 9` and no mainstream tool probes the live CDN; the POSIX temp-file pattern is the only portable way to get a compressor's exit status (both `PIPESTATUS` and `pipefail` are bash extensions). One scoping caveat worth recording: **webpack budgets raw, and its rationale is JS parse+execute — that does not transfer to wasm**, whose baseline compile is ~0.1 s here and which is code-cache eligible above 128 kB. So "gate compressed, report raw" is well-justified *for wasm specifically*.

IvanWng97:
## Review round 2 — comment + correctness lenses

Both delivered after the first disposition. All four lenses are now in, all **APPROVE-WITH-NITS**.

### FIXED

| finding | conf | fix |
|---|---|---|
| **The reading key I added for the raw ratio was INVERTED.** The echo prints gzip/raw, so rising = poorly-compressible code and falling = sprite text; I wrote the reciprocal. Worked both directions numerically: +55 KB sprite text moves 39.2% → 38.0%, +55 KB poorly-compressible code moves it → 39.5% | 88 | inverted back. This was the only place in the change where a comment would send a reader to the wrong conclusion — and I introduced it while *fixing* the "bare number has no reference point" finding |
| **S7 was vestigial.** "Read the recipe's output" was said three times in six sentences, and its trailing clause ("rather than trusting a number written into this line") guarded a number that no longer exists — residue of the deleted "~900 KB" | 71 | sentence deleted |
| **The #871 citation left standing was the wrong one.** Collapsing the body block deleted the rule-compliant citation (present-tense invariant + incident as evidence) and kept the change-history one ("it WAS a proxy UNTIL…") | 66 | reframed to the invariant form |
| **"sprite text is the most compressible part"** — a superlative I never measured; I had it against the payload average, never ranked part by part | 58 | "compresses far better than the wasm code around it" |
| **`test -f` passes on a 0-byte file**, so a truncated wasm reached `$((WIRE*100/RAW))` and died with `division by 0`, exit 127 — fails closed, but tells an operator nothing | 95 | `-s`. Observed: exits 1 with "missing or empty …" |

### ISSUE FILED

- **#877** — justfile recipe bodies are outside `SHELL_SOURCES`, so nothing mechanical catches a reintroduced silent-pass, and there is no test. Regression surface, not a live defect; the repo's `*-selftest` idiom does not transfer cleanly (both hang a flag on a Python script; this is inline recipe shell).

### DECLINED, with reason

- Adding a clause about the two remaining `wc -c | tr -d ' '` pipes. They have the same structure but the opposite behavior — verified: empty var → `test "" -le N` errors → exit 1, **fails closed**. The comment already names *fail-open* as the hazard rather than "pipes", so the distinction is carried.
- `ci-builds.yml:27` "the size cap IS deterministic" — two lenses split; the one that measured it (21-byte implementation spread vs 126 KiB headroom) called it not a finding, and the sentence is about BUILD determinism, untouched here.

### Where the lenses corrected *me*, not the code

- **My trap verification was invalid.** I ran "clean TMPDIR → count leftovers → 0" and called it verified. BSD `mktemp` with no args writes to `_CS_DARWIN_USER_TEMP_DIR`, not `$TMPDIR`, so an empty `$TMPDIR` is equally consistent with "removed" and "created elsewhere" — my test could not distinguish them. The correctness lens instrumented the recipe to emit the real `$GZ` and checked *that* path after exit: removed on all three paths (success, `set -e` abort, explicit over-cap `exit 1`).
- **It reproduced the counterfactual rather than trusting the PR body** — reconstructed the old pipe form against a failing gzip stub: `EXIT=0`, `wasm 0 / 524288 bytes gzipped (0% of cap)`.
- **It drafted a finding against me and then killed it itself** — that "local gzip runs under the CDN" was backwards, based on the figures this diff deletes (394,722 local vs 390,240 Pages). The fresh like-for-like remeasurement pinned to etag `f5b65` is the other way round, and the directional sentence is gone anyway.
- **It ran two gates I had skipped**: `just risk-radar-test` and `just links`, both exit 0. Neither is in `preflight` and I edited `CLAUDE.md` prose.

Stated limitations I am not papering over, from the lens itself: GNU gzip is not installed on this machine (represented by a zlib-9 proxy, 21-byte delta), and the ubuntu `dash` path could not be exercised locally — so the trap is verified on bash-3.2-POSIX only. Both are proxies, not observations; both are immaterial against 24.7% headroom.

`just preflight`, `just risk-radar-test`, `just links` — all exit 0 at `cb3bd006`.

Issue #342, opened 2026-06-16
Mirror upstream env-override path normalization across install targets (verbatim vs ~-expand/trim/Clean)


IvanWng97:
**Found by:** PR #338 round-3 re-verification (cross-source low-severity pattern; completeness critic recommended one consolidated follow-up).

**Problem:** pixtuoid honors each CLI's config-location env overrides but takes them **verbatim**, while several upstreams **normalize** the value:
- **OpenClaw** (`resolveUserPath`): tilde-expands `~` + `path.resolve`. A user-set `OPENCLAW_HOME=~/foo` (un-shell-expanded, e.g. quoted or from a config) lands a literal `~` dir pixtuoid writes to but the gateway never reads. **Most user-plausible.**
- **CodeWhale** (`codewhale_home`): `strings.TrimSpace`. We trim via `nonempty_env`'s trim for the empty check but pass the un-trimmed value (minor — surrounding whitespace).
- **Reasonix** (`cleanEnvDir`): `filepath.Clean`. **Codex** (`find_codex_home`): `canonicalize`.

Each individual case is cosmetic/pathological (shells usually pre-expand `~`; absolute overrides are the documented contract — see the codewhale `resolve_config_path` SCOPE note on why a *relative* override can't be reconciled across processes anyway). But it's a consistent cross-source gap worth one pass.

**Action:** add tilde-expansion (at minimum) + trim to the env-override handling in `install/{openclaw,codewhale,reasonix,cursor,opencode}.rs`, or document the absolute-only contract uniformly. Low priority.

PR #881, opened 2026-08-08
fix(source): mirror each CLI's OWN home resolver, and close the class (#880)

### LINKED ISSUE #880: claude-code / copilot / hermes resolve their home dirs generically — the fail-silent resolution-policy class
Several registered sources resolve their CLI's home/config root with the *generic* answer — `$SOME_ENV` if set, else `~/.<cli>` — rather than mirroring how that CLI actually resolves it. Where the source TAILS a transcript, every unmirrored axis is **fail-silent**: the watcher polls a directory the CLI never writes to, and the office is simply EMPTY. No error, no breadcrumb, no health row.

This is the resolution-POLICY mirroring class already logged in #343 / #342 / #195: *the generic dirs/shellexpand answer IS the bug.* Each CLI resolves its own way — HOME-first vs USERPROFILE-first, `%APPDATA%` vs `~/.<cli>`, verbatim vs `~`-expanded overrides, profile subtrees, XDG relocation, legacy-dir fallbacks.

## How much can hide behind one env var

oh-my-pi is the worked example of the upper bound. Its `packages/utils/src/dirs.ts` resolves a sessions dir through **all** of:

- `PI_CONFIG_DIR` renaming the config root, bound UNDER home Node-`path.join`-style (an absolute value does **not** escape home);
- `OMP_PROFILE` / `PI_PROFILE` selecting `profiles/<name>/agent`, with a validated charset and Windows reserved-device names rejected on every OS;
- an existence-gated `XDG_DATA_HOME` redirect that **flattens** the `agent/` segment away — so the sessions dir is not `<agent-dir>/sessions` at all;
- a drop rule for an inherited, profile-derived `PI_CODING_AGENT_DIR`.

A generic `$PI_CODING_AGENT_DIR else ~/.omp/agent` mirrors exactly none of that. There is no reason to assume the other CLIs are simpler until someone reads their resolvers.

## The sources that still use the generic shape

Bare `$ENV else ~/.foo`, no profile / XDG / relocation mirroring, and no in-tree note ruling it out:

| source | resolver | why it matters |
|---|---|---|
| `claude-code` | `claude_config_dir()` — `CLAUDE_CONFIG_DIR` else `~/.claude` (`crates/pixtuoid-core/src/source/claude_code.rs:138`) | **Highest relevance**: it TAILS transcripts (`~/.claude/projects`), so a wrong root is a silently empty office. |
| `copilot` | `copilot_home()` — `COPILOT_HOME` else `~/.copilot` (`crates/pixtuoid-core/src/source/copilot.rs:38`) | Also transcript-tailing (`session-state/`). |
| `hermes` | `hermes_home()` — `HERMES_HOME` verbatim else `~/.hermes` (`crates/pixtuoid-core/src/source/hermes.rs:33`) | Hook-only, so the mode is muted (hooks are proof-of-life), but the shape is the same and it feeds install + focus. |

Verified NOT affected: `codex` (`codex_home` — documented-deliberate exists-check divergence), `grok` (`grok_home` mirrors `xai-grok-config`'s `paths.rs`, explicitly no XDG/APPDATA), `codewhale` (`home_first_dir`, source-verified). N/A — no sessions root to resolve: cursor / kimi / opencode / reasonix (hook-only), openclaw (daemon, port-keyed), antigravity (closed IDE).

## Ask

For each of the three, read that CLI's **authoritative upstream resolver** and answer per axis: home order, config-dir API, env-override semantics (verbatim vs `~`-expanded; empty-as-unset), profile or workspace subtrees, XDG, legacy-dir fallbacks. Then either

- mirror it, with a `scripts/check_upstream_drift.py` row pinning the names the mirror depends on, or
- record a `CLAUDE.md` sharp edge stating the generic answer is provably correct for that CLI.

`claude-code` first — it is the flagship and it tails.

IvanWng97:
## Summary

Four sources resolved their CLI's home with the generic `$SOME_ENV else ~/.<cli>` answer instead of mirroring how that CLI actually resolves it. Where a source tails a transcript, every unmirrored axis is **fail-silent**: the watcher polls a directory the CLI never writes to and the office is simply empty — no error, no breadcrumb, no health row.

Audited against the **shipped artifact**, not docs prose (two of the four have none): claude-code 2.1.226's bun binary, copilot 1.0.78's `app.js` + its Rust `runtime.node` addon, hermes 2026.8.3's Python, omp's `dirs.ts`. The two native resolvers were **probed by direct invocation** over a case matrix — that is what settled the cases nobody documents (empty / whitespace / padded / `~` / relative).

## Related issue

Closes #880

## Type of change

- [x] Bug fix
- [x] New feature 

## What was actually broken

| bug | symptom |
|---|---|
| hermes resolved `<home>/.hermes` on Windows; upstream is `%LOCALAPPDATA%\hermes` | hooks installed into a `config.yaml` hermes never reads |
| hermes `.strip()`s its env value; `nonempty` returned the padded string | a different directory for `" /srv/hm "` |
| omp ignored `PI_CONFIG_DIR`, both profile vars, and the XDG redirect | wrong sessions root for any non-default setup |
| omp's XDG root silently lost the FD liveness probe | gate matched the *agent* dir, which XDG flattens away |

## Closing the class, not just the instances

This had already been adjudicated three times (#343 / #342 / #195) and came back, so the fix includes the two gaps that let it:

- **`SourceDescriptor.home_env`** — a struct field, so a new source **cannot compile** without answering it.
- **`every_declared_home_env_actually_moves_that_sources_root`** — points each declared var at a temp dir and asserts the root follows, so a declaration cannot be decorative.
- **`doctor` grows a `resolved roots` section** — resolved via `source::resolved_source_root`, the *same* call the driver makes, because a reporter with its own copy could show a healthy path while the watcher polled another.
- **Drift rows** for the two fetchable resolvers (hermes, omp), both verified against live upstream.
- **CONTRIBUTING step 1** now carries the per-axis audit + the probe-the-installed-artifact technique.

## Deliberately NOT mirrored (don't "fix" these)

copilot's `--config-dir` (per-invocation, unobservable from our process) · copilot's XDG (legacy-migration-only, skipped when `COPILOT_HOME` is set) · hermes profiles (out-of-band wrapper env; upstream calls the mismatch a bug, NousResearch/hermes-agent#18594) · CC's `.normalize("NFC")` and grok's `dunce::canonicalize` (change the path *string*, name the same directory) · a **relative** override, which is unmirrorable because each process resolves it against its own cwd — `warn_if_relative_override` passes it through and warns.

## How I tested it

- `just preflight` — **2958 passed, 2 skipped**
- CI-only gates run locally too: `just doc-check`, `just risk-radar-test`, `just api-surface-check`, `just semver`, `just check-windows`, drift selftest
- **Negative controls on every new check**, two of which caught defects in this PR's own work: the omp drift row's first draft would have false-alarmed weekly on working code (`dirs.ts` reads `process.env.X` unquoted), and `home_env` initially declared `None` for opencode/reasonix because the first pass grepped only core — their overrides live binary-side
- Live: all seven core-resolved roots verified against on-disk reality on a dogfooding machine

## Visual verification

- [x] Not a visual change.

## Checklist

- [x] Docs updated in the same commit if I changed module structure, architecture, or public API (`CLAUDE.md` / `README.md`).
- [x] Checked against the recurring pitfalls: char-safe slicing (`normalize_profile_name`'s byte index is reached only after an ASCII-only charset check) · no parallel copies without a bridge test (`resolved_source_root` is the single seam, lockstep-tested) · sanitize at the decode boundary (`doctor::root_row`) · negative branches pinned.

## Known limits, stated rather than papered over

**antigravity, cursor and kimi have no fetchable resolver, and this axis has no defense #2 either** — a decoder sees wire bytes and can breadcrumb; a home resolver aimed at the wrong directory observes *nothing*. Their only defense is re-running the probe matrix when they major. A sharp edge that previously filed them under Antigravity's "defense-#2-only" class is corrected here.

**omp is the weakest evidence grade** of the four deep audits — source-read plus an injected matrix rather than a live probe, because oh-my-pi is not installable on this machine. That is precisely why it carries a drift row.

## AI assistance


IvanWng97:
## Review round 1 — disposition

Every finding driven to one terminal state, per the repo's review protocol.

| # | finding | state |
|---|---|---|
| 1 | `a_relative_override_warns_and_an_absolute_one_stays_quiet` fails on Windows — `/abs/dir` is drive-relative there, so `is_absolute()` is false and the helper correctly warns; the negative control asserted the opposite | **FIXED** (705ebeb8) |
| 2 | `SourceDescriptor.home_env` fails `registry.rs`'s own column bar ("production DEPENDS on, not one a report consults") — only `doctor` + a test read it | **FIXED** — module doc now names it as one deliberate exception with a narrow bar, rather than quietly violating the rule |
| 3 | `omp_sessions_dir` was `pub` with no cross-crate caller; `resolved_source_root` is workspace-internal mechanism | **FIXED** — `pub(crate)` and `#[doc(hidden)] pub` (the `claude_config_dir` precedent) |

Net effect of #3: **the api-surface golden now matches `main` exactly** — this PR no longer widens `pixtuoid-core`'s published API, which retires the golden churn commit e1dc5a42 added.

### What the gates did and didn't catch

Finding #1 is worth recording as a gate-coverage fact: `just preflight` was green (macOS), and **`just check-windows` was also green because it is a cross-*lint*, not a test run**. Only `windows-test` sees this class. All four red checks traced to that one test (`required-test-jobs` and `ci-gate` are aggregates).

### Lens coverage — the gate is NOT fully satisfied

Stated plainly rather than papered over. Four independent lenses were dispatched (correctness · design · per-axis upstream-mirroring · comments); **all four failed to deliver a report** — two died mid-stream with `API Error: Response stalled`, and two went idle without returning. Re-runs failed the same way. This is dispatch infrastructure, not reviewer judgement.

The three findings above came from the orchestrator running the design and Windows checks directly. That is **not** a substitute for differentiated independent lenses, and per the skill a stubbed lens must not be counted toward the floor (PR #455's precedent: a stubbed lens under a clean aggregate hid a real AA failure).

**So the two-lens merge gate is unsatisfied as written.** Not recommending merge on the strength of this round. The remaining coverage should come from either a working lens dispatch or the online review bot's `Findings: N` verdict at this head — whichever a human prefers.

### Not re-verified independently

The upstream-mirroring lens was the one that mattered most here (this repo shipped the resolution-policy class twice: #343, #342, and #338's `%APPDATA%` HIGH that *both* mandated lenses missed by checking a single axis). Its brief — re-fetch every CLI's authoritative resolver in-session and confirm each axis rather than trusting the author's claims — has **not** been independently executed. The author's own evidence is in the PR body; it is exactly the thing a review is supposed to not take on trust.

IvanWng97:
## Review round 2 — online-review findings

Both findings from the bot review are closed. It was the only lens that returned a report, and it earned it — the HIGH is a real bug in this PR's own centrepiece.

| # | finding | state |
|---|---|---|
| HIGH | `node_join` let an absolute `PI_CONFIG_DIR` escape home **on Windows** — `trim_start_matches(['/','\\'])` strips separators but not a drive prefix | **FIXED** (74c264ca) |
| MEDIUM | `hermes_home` blamed an **unset** `HERMES_HOME` when the *derived* default was relative | **FIXED** (74c264ca) |

### The HIGH, verified before fixing

| segment | Unix | Windows |
|---|---|---|
| `/srv/omp` | `RootDir` + components → stripped ✓ | same ✓ |
| `C:\srv\omp` | **one ordinary component**, cannot escape | **`Prefix`+`RootDir` → `Path::join` replaces the base** ✗ |

`PathBuf::push`: an absolute path *"replaces the current path"*, and one with a prefix but no root *"replaces self"*. The fix skips `Prefix`/`RootDir` components — a **no-op on Unix**, so it changes only the arm that was broken.

Deliberate divergence, documented on the fn: Node yields `<home>\C:\srv\omp` for that input, which Windows cannot create. We produce the **bound** `<home>\srv\omp` rather than reproduce an unusable path byte-for-byte.

**Test teeth, stated honestly:** the escape is structurally Windows-only, so `node_join_never_lets_a_rooted_segment_escape_the_base` passes on macOS *against the old code too*. Its teeth are Windows-CI-only. The test's own doc says so, so a green local run is not mistaken for proof.

### The MEDIUM — and why a vanished finding isn't a fixed one

It appeared at `e1dc5a42` and was **absent from the report at `705ebeb8` while the code was unchanged**. Driving each finding to a terminal state — rather than trusting the newest headline — is the only reason it was caught. A warning naming a variable the user never set sends them chasing the wrong one; the repo's own "the message is half the rule" factor.

Security review at the prior head: **Findings: 0**.

### Lens coverage — still short, stated plainly

Four local lenses were dispatched and re-dispatched; **none returned a report** (two stalled mid-stream, two idled empty). That is dispatch infrastructure, not reviewer judgement, but the gate does not care why.

So independent coverage on this PR is: **the online bot (2 findings, both real, both fixed) + the security review (0)**. The per-axis upstream-mirroring lens — the one for the class this repo shipped twice (#343/#342, and #338's `%APPDATA%` HIGH that *both* mandated lenses missed) — has still **not** independently re-verified the upstream claims in the PR body.

Worth noting the bot found a Windows path bug that four local lenses were briefed to look for and never reported on. I'd treat that as evidence for one more independent pass, not against it.

**Not recommending merge on this round.** A human merges here regardless.

IvanWng97:
## Review round 3 — upstream-mirroring lens

The per-axis upstream-mirroring lens returned **REQUEST-CHANGES** with eight findings. All eight are closed in `3fd240cc`. It did the job properly: re-fetched every resolver by `curl` (rejecting a summarizing fetch as too lossy for this purpose), and ran `node` to confirm `path.join` semantics rather than reasoning about them.

**It independently confirmed the load-bearing claims** — hermes `.strip()` + `%LOCALAPPDATA%`, omp's XDG flatten (quoting upstream's own comment), `getSessionsDir` as the right target, codex's exists-gate, grok's unconditional env — with the upstream line quoted per axis.

| # | finding | state |
|---|---|---|
| F2 | omp loads `.env` FILES that relocate its dirs invisibly to us | **FIXED** — named in the sharp edge (re-verified upstream here) |
| F3 | ambient `OMP_PROFILE` reddens the new gate, with a message blaming the resolver | **FIXED** — reproduced, fixed, re-verified, negative-controlled |
| F4 | `registry.rs` OMP row still asserted `<omp_agent_dir>/sessions/…` | **FIXED** |
| F5 | "No agent CLI absolutizes… or expands `~`" is **false** | **FIXED** |
| F6 | codewhale annotation named 2 of 4 rungs | **FIXED** |
| F8 | doctor used bare-empty where every resolver is trim-based | **FIXED** |
| F9 | `omp_agent_dir` left `pub` with no callers | **FIXED — deleted** |
| F10 | codex canonicalizes `CODEX_HOME`; grok's reason named the wrong surface | **FIXED** |

### Two findings falsify claims this PR itself made

- **codex is not "verbatim"** — `find_codex_home` canonicalizes the **env** value and leaves the default alone, the *inverse* of grok's scoping. My doc asserted the opposite.
- **"No agent CLI absolutizes such an override or expands `~`"** is false: omp `path.resolve`s, codewhale expands `~`. A sharp edge that says *don't fix this* must not rest on a false universal. The code was already right — pass-through-and-warn is the only honest option — but the stated reason wasn't.

### F9 went further than the finding

The lens said "demote it". Demoting proved `omp_agent_dir` was **entirely dead** (clippy: `never used`), so it is deleted and its live-env test retargeted onto `omp_sessions_dir` — the fn production actually calls, which is better coverage than it had.

Removing it from the public surface is breaking, so it was **adjudicated, not assumed**: `just semver` reports `pixtuoid-core v0.16.0 → v0.17.0`, *no update required* — it rides the already-open unreleased minor rather than over-bumping a break the open minor covers.

### F3 needed two attempts

My first fix scrubbed the profile vars but restored them **inside** the loop, so iteration 2+ ran with the profile back and it still reddened. It now collects failures, restores once, then asserts — which also means an in-loop panic can't leak env into later tests. Verified under the exact ambient condition and negative-controlled that the gate still fires on a false declaration.

Gates: preflight **2960 passed / 2 skipped**, doc-check, api-surface-check, semver, check-windows, plus the ambient-`OMP_PROFILE` repro now green.

### Coverage

Independent so far: **upstream-mirroring lens** (8 findings) + **online bot** (2) + **security** (0). The correctness lens never returned a report across two dispatches.

Each of those two working reviewers caught real defects the other missed. That is the argument for the remaining lens rather than against it, so I still would not call the gate complete — but the axis this PR is *about* has now been independently re-verified per axis, which was the largest open risk.

Issue #527, opened 2026-07-10
focus: verify resolved pid's chain names the expected CLI (recycle deep-guard)


IvanWng97:
v1 guards click-time pid staleness only via the exiting_at refusal; a lingering slot whose CLI died and pid got recycled can still focus a random app. Deep guard: before activate, check the ancestor chain (or the pid's comm) contains the expected CLI process name. Residual window documented as a sharp edge in core CLAUDE.md.

IvanWng97:
Scope narrowed by the two-lens review fix on PR #532 (commit 5d9b95d2): `patch_identity_pids` now structurally skips the transcript family (any source with a `line_decoder`), so CC/Codex can never carry a stale cached `slot.pid` — they always resolve through the recycle-guarded probes.

What remains for this issue is the HOOK-family window only: an abruptly-dead agent (no session_end, before the stale-sweep/exit-watch lands) whose cached pid gets recycled between death and a user click. The `exiting_at` refusal covers the common case; the deep-guard here would verify the resolved pid's process comm / ancestor chain names the expected CLI before activating. Blocker remains the per-CLI process-name mapping being inherently unstable (opencode runs under bun, etc.).

PR #882, opened 2026-08-08
fix(focus): close both Windows focus-jump gaps (#528)

### LINKED ISSUE #528: focus(windows): hook-family pid absent (cmd /C parent) + foreground-lock workaround
Two Windows gaps: (1) hook-family ppid is the transient cmd.exe (documented parent_pid trap) - needs a different channel; (2) SetForegroundWindow may be denied by the foreground lock - AttachThreadInput workaround. Both currently silent no-ops.

IvanWng97:
## Summary

Windows focus-jump was a double no-op: shim-family sources sent no pid at all, and the one activation call the binary made is the one the system refuses. This closes both gaps, plus the recycle guard that makes the first one safe to ship.

## Related issue

Closes #528

## Type of change

- [x] New feature

## What changed

**1. The pid channel.** `getppid` names the CLI only because `sh -c` EXECs the hook; a Windows runner shelling via `cmd /C` leaves a transient parent that dies with the shim. The shim now WALKS instead — one Toolhelp32 snapshot, skip the interposed shells, stamp the first ancestor that is the CLI (`pixtuoid-hook/src/cli_pid.rs`). The walk is pure over an injected snapshot (the `hook_cmd` precedent), so its decision unit-tests on macOS; only the FFI is `cfg(windows)`.

**2. The recycle guard — why 1 is safe to ship.** A markerless Windows stamp would hand focus a wide staleness window (the residual the core guide already flagged under this same issue), so `pid_start_marker` grows a Windows arm: creation `FILETIME` via `GetProcessTimes` under `PROCESS_QUERY_LIMITED_INFORMATION`. #527's click-time check is now live there rather than inert — which also makes the walk's failure mode benign, since a transient pid stamped by mistake is dead by click time and gets refused.

**3. The foreground lock.** `activate_os` tries unattached first (the `floating` painter IS the foreground process), then retries under an RAII `AttachThreadInput` borrow of the foreground thread's input state.

Reading the primary docs rather than the folklore changed three things:

- `AttachThreadInput` "fails if either of the specified threads does not have a message queue", threads start without one, and MSDN contradicts itself on which calls mint it — a console TUI is exactly the at-risk thread, so the `PeekMessageW` prime is load-bearing. **Without it the whole workaround would have compiled, passed every gate, and been a permanent silent no-op.**
- The `SetForegroundWindow` BOOL is not a usable verdict (the denial path raises no error and merely flashes the taskbar button), so success is judged by `GetForegroundWindow`.
- Two better-known bypasses are refused and recorded as a sharp edge: the `SendInput` ALT tap injects a global input event into whatever app the user is typing in, and `FreeConsole` would detach the console the TUI draws into.

## How I tested it

- `just preflight` — green, 2967 tests.
- `just check-windows` — clippy-clean, and **negative-controlled**: a deliberate typo in each of the three new `cfg(windows)` blocks was confirmed to red the gate, since on a macOS dev box that lint is the only thing that compiles them.
- `just comment-lint` / `just risk-radar-test` — clean.

**What is NOT verified, and what closes it.** The activation path is unreachable from a headless runner. `windows-test` runs the real Toolhelp32 walk and the marker arm on windows-latest (`cli_pid::tests::the_live_resolver_names_a_real_ancestor` takes a live snapshot; `proc_start`'s tests are cfg-widened to Windows). What remains for a human on Windows: whether each CLI's hook runner really interposes `cmd.exe`, and whether the attach actually wins the foreground grant. The macOS live-dogfood test is widened to Windows for exactly that — `cargo test -p pixtuoid --lib focus -- --ignored`.

Worst case if either turns out wrong: a silent no-op, i.e. today's behaviour. This change cannot regress a working path.

## Visual verification

- [x] Not a visual change.

## Checklist

- [x] I read the relevant `CLAUDE.md` (root + `pixtuoid-core`, `pixtuoid`).
- [x] New behavior has a test (pure walk table-driven on every OS; marker arm live on Windows CI).
- [x] Docs updated in the same commit (`pixtuoid-core/CLAUDE.md`, `pixtuoid/CLAUDE.md` + a new foreground-lock sharp edge, `FocusChannel` doc).

## AI assistance


IvanWng97:
## Review disposition — 5 local lenses + 2 online bots

Every finding to exactly one terminal state. Head at sweep: `f26db2ae`.

### FIXED — behavior
| Finding | Source |
|---|---|
| Toolhelp32 walk ran outside any time bound (invariant #5) | security bot MEDIUM + lens 3 |
| Arming the watchdog earlier *inverted* the failure — it would drop the whole envelope on a healthy daemon. The walk now has its own `WALK_BUDGET` on a worker thread | lens 3 (rejected my first fix) |
| Env-mode resolved the pid twice → two process snapshots per CodeWhale hook on Windows | mine + lens 3 F2 |
| `.unwrap()` in the never-panic crate | auto bot HIGH (consequence refuted, line removed anyway) |

### FIXED — verification gaps
| Finding | Source |
|---|---|
| `shim_pipe.rs` already spawned `cmd /C → shim` on Windows CI and never asserted `_pid`. It does now — real proof of the skip against a real `cmd.exe` | lens 3 F3 |
| Windows marker had no teeth (mutating `imp` to a constant passed every assertion) | lens 1 F2 |
| `doctor` printed "absent on Windows", which this branch made false | lens 2 F5 |

### FIXED — false or overscoped claims (all mine)
"the marker re-read refuses it" (false on Windows — a handle keeps a terminated pid readable; the *walk* is what saves it) · "every hook runner shells through `sh -c`" (2 sites) · four-of-six `SetForegroundWindow` granting conditions presented as all of them · the BOOL rationale (was about `GetLastError`) · "empty on any failure" (a mid-enumeration failure truncates) · "over-skip needs the CLI named `cmd.exe`" (ignored the other two names) · "reaches the root" naming a `None` the code returns `Some(0)` for · the `hook_cmd` attribution · 4 co-located sites the diff invalidated (`focus/mod.rs`, `hook/mod.rs`, `registry.rs`, `pid_watch.rs`).

### REFUTED — with evidence
- **auto bot HIGH: "`.unwrap()` will hard-fail `just check-windows`."** It does not. `check-windows` exits 0 with the line present, and CI's `windows-check` passed on it. Negative-controlled: a plain fn-body `.unwrap()` in the same file fails the gate instantly, so the deny is live and the pass is not vacuous — `clippy::unwrap_used` simply does not fire on a `const` item. Line removed regardless, because a `.unwrap()` here reads as a violation and would be re-flagged forever.
- **Two Toolhelp32 implementations.** Justified by the crate boundary — `pixtuoid-hook` is deliberately dependency-light (invariant #5). Both bot and lens 2 agreed independently. Pinned per the magic-value rule with a pairing note at *both* sites, since a test can't reach across them off Windows.
- **8 candidates lens 5 refuted itself** against existing sharp edges (creation polarity, the daemon lane, the markerless-skip assumption, the stamp-time race).

### NOT DONE — deliberate, with reasons
- **`patch_identity_pids` could refuse to stamp when the marker is unreadable** (lens 1 F1). Rejected: it would regress Unix, where a transient marker failure currently still yields working focus, to close a Windows-only improbable misfire. Documented as a residual in `pixtuoid-core/CLAUDE.md` instead.
- **`SW_RESTORE` survives a denied jump** (lens 2 F3 / lens 1 F4) — the one place a focus miss isn't a *pure* no-op. Reordering doesn't work (`SetForegroundWindow` on an iconic window can foreground it and leave it an icon). Documented as clause 4 of the sharp edge.
- **Widen the crate's deny list** (lens 3 N1). Measured: of twelve panic-capable constructs in non-test code, the deny catches three (`unwrap`/`expect`/`panic!`) and misses nine — `unreachable!`, `todo!`, `unimplemented!`, indexing, slicing, arithmetic overflow, `assert!`, `debug_assert!`, `exit`. Matters more here than elsewhere because `panic = "abort"` makes a shim panic a SIGABRT with no `catch_unwind` backstop, so this list is the only defense. Latent — current code uses none of them. RECORDED HERE, deliberately not filed: it is out of #528's scope and small rather than refactor-sized, so this repo's own rules say neither auto-file it nor fix it inside this PR.
- **`fn main() -> Result<()>` is a latent non-zero-exit hole** (lens 3 F4). Pre-existing and unreachable today (three returns, all `Ok(())`, zero `?`). Same disposition: recorded, not filed.

### Scope note
Windows focus-jump is fixed for the five `ShimStamp` sources only. CC, Codex and grok still no-op there — their probes are `cfg(unix)`. Out of scope for this PR, but nobody should read this as "Windows focus works now".

### Still unverified
CI executes the FFI, the marker, and the real `cmd /C` walk. It cannot execute the activation. Whether `AttachThreadInput` wins the foreground grant needs one human on Windows: `cargo test -p pixtuoid --lib focus -- --ignored` (alt-tab away first, or the unattached attempt wins and the retry never runs).

PR #885, opened 2026-08-09
docs: rightsize the CLAUDE.md corpus for Claude 5 context engineering

IvanWng97:
## Summary

The nested crate guides carried their whole reference corpus inline, and a nested `CLAUDE.md` auto-loads wholesale when you touch its tree. `## Layout` + `## Known sharp edges` were **85–98%** of each guide, so a cross-crate session paid ~68k tokens of prose before reading a line of code. Both sections are lookup-on-demand material, not "always do X" rules, so they move to siblings and the guides keep a one-line index.

Measured against Anthropic's current guidance ([using CLAUDE.md files](https://claude.com/blog/using-claude-md-files), [context engineering for Claude 5](https://claude.com/blog/the-new-rules-of-context-engineering-for-claude-5-generation-models)) — fetched, not recalled.

| | before | after |
|---|---|---|
| auto-load total | 97,139 | **22,265** (−77%) |
| cross-crate session | 76,864 | **12,873** (−83%) |
| every session (root) | 8,370 | 7,731 |

**Content is conserved to the word** — each directory's (guide + siblings) word count is strictly *higher* than the original, the delta being the added index lines.

## Related issue

n/a

## Type of change

- [x] Refactor / chore — docs restructure + one new lint gate (no Rust changed)

## What's in it

1. **`ae85d9c`** — `Layout` / `Known sharp edges` → `LAYOUT.md` / `SHARP-EDGES.md` (plus core's `UPSTREAM-DRIFT.md`, site's `SINGLE-SOURCED.md`). Extends the `WHERE-TO-LOOK.md` pattern the repo already established to the two sections that actually carry the weight. Each index line is the entry's **own opening phrase verbatim**, so `grep` lands in the sibling — deliberately no anchors, since this repo's doc-anchor citations already rot (risk-radar exists for that).
2. **`ff07949`** — root's sharp-edge index had become a *third* layer: 18 **paraphrased** entries restating what each nested guide now indexes verbatim. Paraphrases can't be grepped into the full text and drift independently. Collapsed to what root uniquely owes — a cross-crate map of which crate owns a class of trap.
[...]

PR #890, opened 2026-08-10
fix(tooling): route every scripts/ git call through one scrubbed seam

IvanWng97:
## Summary

Follow-up to #888. That PR stopped `gen-guides.py --selftest` from staging into the real index, but it fixed **one instance of a two-instance class**, and its new control could not see the regression it named.

`comment-lint.py` had the same shape and was **worse** — its fixture `commit`s, so an ambient `GIT_DIR` rewrote the ambient repo's HEAD, not just its index. Measured against a 40-file stand-in:

```
comment-lint selftest: all checks passed
victim: 40 files, 1 commit  ->  8 files, 3 commits ("fixture")
```

Three changes, in the order they matter:

1. **`scripts/gitenv.py` owns the only git seam.** The scrub lives *in* `git()`, not at each `main()` — a per-entry-point call is one forgettable line per script, and script #3 forgets it. The var list comes from `git rev-parse --local-env-vars`: git owns that list and extends it (`GIT_CONFIG_COUNT` is a later addition), where a `GIT_*` prefix match also drops `GIT_SSH_COMMAND`, `GIT_CONFIG_GLOBAL` and `GIT_COMMITTER_*`. [githooks(5)](https://git-scm.com/docs/githooks) prescribes exactly this: code invoking git *"in a foreign repository … should clear these environment variables"*.

2. **`bypass_sweep()` is the recurrence gate.** The two known instances surfaced only after one ate a developer's index. Wired into `just lint` **and** `ci-lint.yml`, because the hygiene job enumerates recipes rather than calling `just lint`.

3. **The controls spawn the entry point** and aim their own snapshots positively (`GIT_DIR` + `GIT_WORK_TREE` + `GIT_INDEX_FILE`) instead of routing through the subject.

## Why #888's control was inert

Mutation-tested both ways:

| mutant | env | selftest said | index |
|---|---|---|---|
| gut the scrub body | clean | FAILED ✓ | — |
| gut the scrub body | `GIT_DIR` set | all 11 passed | **40 → 1** |
| delete the `main()` call site (the original bug) | clean | all 11 passed | **40 → 1** |

It called the scrub in-process, so deleting the call site walked straight past it; and it restored the ambient `GIT_DIR` before reading `after`, so both snapshots read the same wrong index and matched.

**The first draft of this PR repeated the mistake** — building the harness out of `git()` let a broken `git()` redirect the before *and* after snapshot to the same wrong index. Hence `_harness_git`, which deliberately does not route through the subject.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] New theme / sprite pack
- [ ] Docs only
- [x] Refactor / chore

## How I tested it

- `just lint` — all 15 checks green, including the new `gitenv`.
- `just preflight` via the real `pre-push` hook — **2969 tests passed**.
- Mutation sweep against the finished code:

```
BASELINE   gitenv 0 (4 controls)   gen-guides 0 (11 controls)
MUTANT A (git() drops the scrub)
  gitenv      1  `git()` must not let an ambient GIT_DIR relocate a `-C` call
  gen-guides  1  an inherited GIT_DIR must not let `gen-guides.py --selftest` reach …
MUTANT B (a bare git argv sneaks in)
  gitenv      1  these bypass `gitenv.git()`: ['gen-guides.py:293']
```

- Under `GIT_DIR`+`GIT_INDEX_FILE`, all three selftests leave the stand-in byte-identical (40 files, HEAD unmoved).
- **Live dogfood:** this branch was pushed from a linked worktree — the condition that actually exports `GIT_DIR` to `pre-push`. Tracked files stayed at 799 across the run; pre-fix that same operation dropped it to 4.

Not run: clippy/hack in isolation beyond what `preflight` covers — no Rust changed.

- [x] Not a visual change.

## Checklist

- [x] Docs updated in the same commit if I changed module structure, architecture, or public API.
- [x] Checked against the [recurring pitfalls](https://github.com/IvanWng97/pixtuoid/blob/main/docs/CONTRIBUTING.md#recurring-pitfalls-this-codebases-review-history-distilled) — this PR *is* a "no parallel copies without a bridge test" fix.

## Note for #889

`0dbbc11f` on `refactor/desk-seat-facing` is an independent parallel fix for the same bug. This PR deletes the code it patches, so drop that commit before rebasing #889.

## AI assistance

### DIFF
--- a/.github/workflows/ci-lint.yml
+++ b/.github/workflows/ci-lint.yml
@@ -112,6 +112,7 @@ jobs:
       - run: just links
       - run: just drift-selftest
       - run: just gen-guides-check
+      - run: just gitenv-selftest
 
   msrv:
     name: msrv
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -21,7 +21,7 @@ git hooks call the same recipes.
 
 ```bash
 just              # list recipes
-just preflight    # full pre-push gate: lint (fmt + machete + deny + arch + shfmt + shellcheck + actionlint + actionlint-composites + zizmor + ci-observability + json-schemas + links + drift-selftest) → clippy → hack → test
+just preflight    # full pre-push gate: lint (fmt + machete + deny + arch + shfmt + shellcheck + actionlint + actionlint-composites + zizmor + ci-observability + json-schemas + links + drift-selftest + gen-guides-check + gitenv-selftest) → clippy → hack → test
 just fmt          # auto-format
 just test         # the whole suite (cargo-nextest if installed, else cargo test)
 ```
--- a/justfile
+++ b/justfile
@@ -1204,11 +1204,10 @@ ast-grep-test:
 comment-lint-selftest:
     python3 scripts/comment-lint.py --selftest
 
-# A git hook exports the repo-local GIT_* vars and they OUTRANK `git -C`/`cwd=`,
-# so a script driving a throwaway repo drives the REAL one instead. This pins the
-# scrub AND sweeps scripts/ for anything spawning git outside `gitenv.git()` — the
-# recurrence gate, because the two known instances were found only after one of
-# them ate a developer's index. Runs in `lint`; CI enumerates it separately.
+# The seam's WHY lives in scripts/gitenv.py's docstring. This pins the scrub AND
+# sweeps scripts/ for anything spawning git outside `gitenv.git()` — the recurrence
+# gate, because one of these leaks ate a developer's index before anyone noticed.
+# Runs in `lint`; CI's hygiene job enumerates it separately.
 [group('meta')]
 [doc("Self-test the scripts' git-env scrub + sweep for bypasses")]
 gitenv-selftest:
--- a/justfile
+++ b/justfile
@@ -306,6 +306,7 @@ lint:
     run links   just links               & pids+=($!)
     run drift   just drift-selftest       & pids+=($!)
     run guides  just gen-guides-check     & pids+=($!)
+    run gitenv  just gitenv-selftest      & pids+=($!)
     for p in "${pids[@]}"; do wait "$p" || fail=1; done
     [[ $fail -eq 0 ]]
 
@@ -1202,6 +1203,16 @@ ast-grep-test:
 [doc('Self-test the comment-lint driver (pathspec + hidden-dir scan)')]
 comment-lint-selftest:
     python3 scripts/comment-lint.py --selftest
+
+# A git hook exports the repo-local GIT_* vars and they OUTRANK `git -C`/`cwd=`,
+# so a script driving a throwaway repo drives the REAL one instead. This pins the
+# scrub AND sweeps scripts/ for anything spawning git outside `gitenv.git()` — the
+# recurrence gate, because the two known instances were found only after one of
+# them ate a developer's index. Runs in `lint`; CI enumerates it separately.
+[group('meta')]
+[doc("Self-test the scripts' git-env scrub + sweep for bypasses")]
+gitenv-selftest:
+    python3 scripts/gitenv.py --selftest
 
 # Risk radar — show the documented review escalations for the high-risk seams
 # THIS branch touches (advisory, deterministic, no LLM). Dogfood before pushing
--- a/scripts/comment-lint.py
+++ b/scripts/comment-lint.py
@@ -21,6 +21,8 @@ import subprocess
 import sys
 import tempfile
 
+import gitenv
+
 PATHSPEC = ("*.rs", "*.py", "*.pyi")
 
 
@@ -31,8 +33,8 @@ def added_lines_by_file(
     # `base...HEAD` is the merge-base range (the PR's own commits); bare `base`
     # also folds in uncommitted working-tree edits.
     rev = base if worktree else f"{base}...HEAD"
-    diff = subprocess.run(
-        ["git", "diff", "--unified=0", "--no-color", rev, "--", *PATHSPEC],
+    diff = gitenv.git(
+        "diff", "--unified=0", "--no-color", rev, "--", *PATHSPEC,
         capture_output=True,
         text=True,
         check=True,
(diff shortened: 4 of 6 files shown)


IvanWng97:
## Two-lens review (diff scope) — disposition sweep

Four lenses run against `6cd6d1bb..6d565708`: correctness, design/blast-radius, comment (escalation: new module with dense docstrings), wrong-abstraction (escalation: consolidation refactor). Correctness and design each returned **REQUEST-CHANGES**, converging independently on one class: **the PR shipped a recurrence gate whose own controls could not fail** — the same defect this PR exists to fix in #888.

Fixes folded into `729ef80a`.

### Mutation matrix — the evidence that mattered

| mutant | gitenv | gen-guides | before `729ef80a` |
|---|---|---|---|
| a — `git()` drops the scrub | 1 | 1 | killed |
| c — `_at` only `GIT_DIR` | 1 | 0 | **SURVIVED** |
| d — control always `None` | 1 | 0 | **SURVIVED** |
| e — no `INNER_ENV` guard | 0 | 1 | **SURVIVED** (hung, 173 procs) |
| f — `clean_env` only `GIT_DIR` | 1 | 1 | **SURVIVED + corrupted** |
| g — sweep returns `[]` | 1 | 0 | killed |

All six now killed; baseline clean.

### FIXED (14)

1. **Control discarded the child's exit code** — a child dying before its git calls was indistinguishable from a clean run (measured: exit 3 → `None`, raise → `None`). Now fails on non-zero child, which turns "move `gitenv.py` into a subdir" into a red instead of three green checkmarks.
2. **Child simulated `GIT_DIR` alone** — 14 of 15 scrubbed vars untested, and `GIT_INDEX_FILE` is the one a *plain* checkout's `pre-commit` exports. Child now gets `_hook_env` (`GIT_DIR` + `GIT_INDEX_FILE`).
   **Deliberately not `GIT_WORK_TREE`** — pinning it aims `add -A` back at the fixture's own tree, so a genuinely leaking child stages nothing new and the control passes. Measured: with it the outer index stayed `['sub/kept.md']`; without it, `['PHANTOM.md']`. The first draft of this fix had it and would have been weaker than what it replaced.
3. **`ambient_git_control` had no negative control anywhere** — mutant (d) passed all three selftests. Now both directions plus a died-early probe.
4. **Sweep was a line-scoped regex** — missed tuple argv (this repo's own idiom for argv fragments), absolute git paths, `shell=True`, and a black/ruff-exploded `[\n "git",`; and it **false-fired on prose** quoting the anti-pattern. Now an `ast` walk: comments and docstrings aren't nodes, a wrapped argv is one node however it breaks, and an unparseable file is *reported*, not silently unswept. Seven forms pinned firing, prose pinned silent.
5. **`timeout=` on the spawn** — the `INNER_ENV` guard was the only recursion brake.
6. `SELF` compared by resolved path, not filename.
7–9. Three false-universal docstrings: `_at`'s "no ambient var can redirect it" (12 of 15 survived `**os.environ`) is now **true by construction** rather than reworded; `bypass_sweep`'s "every git argv"; the module docstring's hook claim (`GIT_DIR` is worktree-specific, but `pre-commit` exports `GIT_INDEX_FILE` from ANY checkout).
10. Twin **"Run LAST"** comments deleted — vestigial, inherited from #888's in-process control. The call sits outside the fixture block and mutates nothing in the parent, so ordering was never load-bearing.
11. Restored the diagnostic symptom #888's docstring carried (`--check` dying on a path never on disk) — the thing that lets the next reader recognize this from the error they actually see.
12. Hardcoded control count dropped (an inline number nobody re-measures).
13. `justfile` recipe comment → pointer form, matching the `gen-guides` recipe.
14. `CONTRIBUTING.md`'s `lint` enumeration was missing `gen-guides-check` (since #885) and `gitenv-selftest`.

### REFUTED-with-trace

- **`_harness_git` not routing through `git()`** — deliberate, and the reason is the finding above it: a control may not route through its own subject. Documented at the declaration.
- **`functools.cache` on `clean_env()`** — a stale snapshot can only *retain* an older filtered dict; every snapshot applies the filter, so it cannot make `git()` leak. Mutations are bracketed by `cache_clear()`.
- **Mutant (f) "silent pass"** — could not reproduce. On macOS/git 2.54 it reds via an unhandled `CalledProcessError`, victim intact. The underlying coverage gap was real and is fixed regardless, but the demonstration did not replicate.

### ISSUE-FILED

#891 — `comment-lint`'s ambient control runs only in the advisory `supplemental` job (it needs `ast-grep`, which `lint`'s tool preflight deliberately doesn't require), and the sweep is `scripts/**/*.py` only (`site/config/released-version.mjs` runs an unscrubbed `git describe`; read-only, not reached from `preflight`). Neither is worsened by this PR.

### Wrong-abstraction pass (consolidation validation)

Seven call sites; **argv and kwargs identical at every one** — including `-c user.email`/`-c user.name` order relative to the subcommand, and `cwd=cwd` where it may be `None`. Only the env moved. One site touches the real repo (`comment-lint.py:36`, the CI `git diff`); in CI no repo-local `GIT_*` exists so the scrub is a no-op — **witnessed by the `comment-lint (advisory)` job passing**, not pinned by a test, and I won't claim otherwise.

Sites 5–8 vindicate `--local-env-vars` over a blanket `GIT_*` scrub concretely: they *commit*, and a blanket scrub would drop `GIT_AUTHOR_DATE`/`GIT_COMMITTER_*` (which `pre-commit` exports), silently altering commit metadata. `--local-env-vars` omits those by design.

Revert case considered and rejected: two inline copies leave the sweep nothing to define "routed" against — the gate needs a single call target.

### Verification

- `just lint` — 15/15 green including `gitenv`.
- `just preflight` via the real `pre-push` hook — **2969 tests passed**.
- **Live dogfood, twice:** both pushes came from a linked worktree, the condition that actually exports `GIT_DIR` to `pre-push`. Tracked files stayed at 799 across each; pre-fix that same operation dropped the index to 4.

Gate not yet satisfied: a human merges, and the online review bot's verdict at this HEAD still needs reading.

PR #894, opened 2026-08-10
fix(omp): mirror the .env overlay that can move omp's sessions dir

findyourexit:
## Summary

#880 mirrored omp's `getSessionsDir()` axis for axis, so every axis it reads from the process environment. `omp` doesn't stop there: `packages/utils/src/env.ts` parses `<agentDir>/.env`, `<configRoot>/.env` and `~/.env` at module load, applies them, then calls `refreshDirsFromEnv()` which rebuilds the resolver. So a user whose `PI_CODING_AGENT_DIR`, `PI_CONFIG_DIR` or `XDG_DATA_HOME` lives only in a `.env` file moves omp's sessions dir for real, while `pixtuoid` polls the default tree and shows an empty office (similar to the fail-silent class #880 closed, one layer further out).

> [!NOTE]
> In case it helps, I figure it best to flag that these changes aren't intended as a "re-do" of #880, but rather some accretionary changes that I'd had on a local branch (with its own resolver, that was mostly superseded by the changes in #880). So these changes keep the current resolver implementation as-is, and adds only the changes that remain of merit.

## Related issue

n/a

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] New theme / sprite pack
- [ ] Docs only
- [ ] Refactor / chore

## How I tested it

The `omp` mirror has to carry two upstream subtleties, and both are pinned.

The first is that the `OMP_<X>` to `PI_<X>` alias is applied after the whole file has been read, which means it overrides an explicit `PI_` key set earlier in that same file. 

The second is that `refreshDirsFromEnv` reuses the profile frozen at module load, while the override resolver it calls re-reads `PI_PROFILE` live. A profile arriving from a file must therefore not move the tree, but it must still drop an inherited profile-derived `PI_CODING_AGENT_DIR`. That is why `OmpEnv` now carries `pi_profile` and `pi_profile_live` as separate fields.

I've added six unit tests to cover this:
- `a_dotenv_directory_var_moves_the_sessions_dir_the_way_omp_does`, which walks the precedence matrix, including the case where a set but empty shell variable still yields to a file.
- `a_dotenv_profile_is_ignored_because_upstream_freezes_the_profile` and `a_dotenv_pi_profile_still_drops_a_profile_derived_agent_dir`, which cover the two halves of the frozen versus live split.
- `a_dotenv_omp_prefixed_key_aliases_and_overrides_its_pi_twin`, whcih covers the alias.
- `dotenv_files_are_located_before_the_overlay_can_move_them`, which pins the load order.
- `dotenv_line_parsing_matches_upstream_bun_semantics`, which covers the line parser.

Three further tests in the drift selftest cover the new `OMP_ENV_*` rules and the two comment-strip controls they depend on.

Gates run locally, with observed exit codes:

- `just preflight` (exit 0, 2975 tests)
- `just api-surface-check` (exit 0)
- `just semver` (exit 0)
- `just gen-guides-check` (exit 0)
- `python3 scripts/check_upstream_drift_selftest.py` (exit 0)
- `python3 scripts/check_upstream_drift.py` against live upstream (exit 0, no omp findings)
- `node scripts/gen-readme.mjs --check` (exit 0)

The two drift runs matter together rather than separately. The live run proves the new `OMP_ENV_*` rules stay silent against the real `env.ts`, and the selftest proves they fire when a watched name disappears, which is the half a presence sweep usually leaves untested.

The line parser was checked clause by clause against the upstream `parseEnvLine` fetched during the session, covering the `export` prefix, all three quote forms, the escaped quote that does not close a value, the inline comment rule, and the NUL rejection. The `||` in the parser's early return is a verified equivalent mutant, so it is pinned in `.cargo/mutants.toml` rather than left as a bare survivor. Running `cargo mutants --list` against `--config /dev/null` shows 128 mutants in the file, the row removes exactly one of them, and that one is the `||` in question. It is the only disjunction in the function, so the row cannot bury a killable neighbour.

- [x] Not a visual change, or — screenshot/GIF attached below and it reads at half-block scale.

## Checklist

- [x] Docs updated in the same commit if I changed module structure, architecture, or public API (`CLAUDE.md` / `README.md`).
- [x] Checked against the [recurring pitfalls](https://github.com/IvanWng97/pixtuoid/blob/main/docs/CONTRIBUTING.md#recurring-pitfalls-this-codebases-review-history-distilled): char-safe slicing · no parallel copies without a bridge test · sanitize at the decode boundary · negative branches pinned.

## AI assistance


findyourexit:
Hey @IvanWng97 absolutely don't want to bother you – I just wanted to query your preference for some changes I've got in the barrel, that I'm hoping to eventually contribute.

This PR is the first of a stack of three that I'm hoping to raise, each focusing primarily on improving support for the `omp` (Oh-My-Pi) coding agent/harness (since it's the one I tend to use a lot, at present). I just wanted to check if you've got a preference as to how I contribute these – continuing with the current approach of trickling through each independently useful PR, or would it be better if I rasied them as one more monolithic PR with some mixed improvements?

IvanWng97:
Thanks for asking — please keep trickling them as separate PRs. Each one being independently useful is exactly the right shape; a combined PR would make the review pass worse, not cheaper, since the mandatory two-lens gate scales badly with diff size and mixed themes.

Btw — is #894 in its final shape? Just want to make sure I'm reviewing a settled head.

findyourexit:
> Thanks for asking — please keep trickling them as separate PRs. Each one being independently useful is exactly the right shape; a combined PR would make the review pass worse, not cheaper, since the mandatory two-lens gate scales badly with diff size and mixed themes.
> 
> Btw — is #894 in its final shape? Just want to make sure I'm reviewing a settled head.

Epic! Sounds good – I'll keep the current approach.

Yep #894 is good to go! (It might have seemed like it's in flux because I've rebased it a few times recently to catch your latest changes in `main`).

PR #897, opened 2026-08-10
fix(hook): stop a wrapper shell's pid from ending the session (#896)

### LINKED ISSUE #896: [Bug]: Cursor CLI: sprite walks in/out every tool call
### What happened?

I connected Cursor CLI in the Sources panel and started a Cursor session in another terminal. A character appears, but instead of staying at a desk typing while the agent works, it almost immediately walks back out — then walks in again on the next tool call, over and over, for the whole session.

Expected: once the agent is running, the character should walk in from the elevator and stay at its desk until the session ends or I disconnect that source

### Steps to reproduce

My pixtuoid install from `cargo install`
My cursor cli install from arch aur https://aur.archlinux.org/packages/cursor-cli

1. Run `pixtuoid`
2. In another terminal, start Cursor CLI and give it a task that uses tools (Read / Shell / etc.)
3. Watch the office: the Cursor character walks in from the elevator, then walks back out within ~1s, repeating on each tool call

### Screenshot / GIF

https://github.com/user-attachments/assets/95940fd7-b237-4b0e-8212-cccfa7144800

### pixtuoid version

0.16.0

### Which agent CLI were you visualizing?

Other (note it below)

### How did you install it?

Cargo (crates.io)

### OS + terminal

Arch Linux 7.1.6-arch1-1, kitty

### Relevant logs

```shell

```

IvanWng97:
## Summary

Cursor sprites walked out of the office ~140 ms after every hook event and back in on the next one. Root cause is in the shim, and the blast radius was every hook source, not just Cursor.

Cursor never runs a hook command directly — it `eval`s it inside its own shell-snapshot wrapper:

```
/bin/zsh -c 'builtin export PATH=…; snap=$(command cat <&3); … builtin eval "$1"; };
             COMMAND_EXIT_CODE=$?; dump_zsh_state >&4; builtin exit $COMMAND_EXIT_CODE' \
         -- 'PIXTUOID_SOURCE=cursor /path/pixtuoid-hook' preToolUse
```

That wrapper has work *after* ours, so the shell can never `exec`-replace itself with the shim. `cli_pid`'s bare `getppid` therefore stamped `_pid` with a throwaway process that dies in milliseconds — a different one every hook (94223, 4808, 5379, 8924, 9869, 13594 in one captured session). `HookPidWatch` watched each corpse and synthesized a `SessionEnd`. The shim's module doc asserted the false premise outright: *"the runner execs the shim, directly or through `sh -c`, so the parent IS the CLI"*.

Fixed at **both** ends of `_pid`, because the two consumers need different guarantees:

- **Accuracy** (`pixtuoid-hook/cli_pid.rs`) — extend the Windows `cmd.exe` walk to Unix so `_pid` names the CLI above the wrapper. Focus-jump needs the pid to be *right*, not merely harmless: Cursor is `FocusChannel::ShimStamp`, so clicking its sprite had been targeting a dead, possibly recycled pid. One shared walk over a per-OS row reader (`proc_pidinfo` / `/proc/<pid>/stat` / Toolhelp32), which now also refuses to name the reaper.
- **Safety** (`pixtuoid-core/hook/pid_watch.rs`) — arm the exit watch only on the **second** sighting of a `(pid, agent)` pair. A per-invocation wrapper reports a fresh pid every hook and so can never be corroborated. That is a property of what it *is*, not what it is *named*, so it holds for wrappers the interposer list has never heard of (`nu`, `elvish`, a `node -e` shim). `handle_conn` dedupes each batch's bind targets so a payload cannot corroborate its own pid.

The one-event lag costs nothing where it lands: a transcript-bearing source binds once, at `SessionStart`, and already resolves an authoritative pid via `jsonl::liveness::ProbeSnapshot` with its own `ExitWatch`. The five hook-only sources re-emit `Identity` per activity event.

**Other sources.** `pid_bind_target` was never source-gated, so all 11 hook sources bound `_pid` to a session-ending watch. Only Cursor is *verified* (the other four `ShimStamp` CLIs aren't installed here) — but all three cases are now safe: runner execs → walk stops at the CLI, unchanged; wraps in a known shell → walk fixes it; wraps in anything else → corroboration blocks the flap.

## Related issue

Closes #896

## Type of change

- [x] Bug fix

## How I tested it

`just preflight` (2975 passed) + `just doc-check` (CI-only gate, run locally because this adds intra-doc links).

Verified end to end against real `cursor-agent` + real `pixtuoid`, same harness before and after:

| | before | after |
|---|---|---|
| synthesized `SessionEnd` mid-session | 6 | 0 |
| `agents=[]` while Cursor was working | 4× | never |
| sprite leaves at the real `sessionEnd` | yes | yes |

**Correction, from re-verifying after review round 3:** an earlier revision of this description claimed Cursor *gains* a working abrupt-exit signal. That was a generalization from one lucky run. Instrumenting what the shim stamps across a session shows the walk resolves the real `cursor-agent` for most hooks but not all — some arrive through a different non-shell ancestor (one session: pid `21060` for three hooks, `30327`/`39272`/`43034` for three others). So the abrupt-exit signal is **best-effort** and falls back to the stale sweep. The #896 fix itself is unaffected and holds unconditionally — corroboration refusing to arm on a pid that disagrees with its predecessor is the guard working.

New negative-branch tests pin the refusal side: `a_once_seen_pid_dying_ends_nothing`, `handle_conn_batch_does_not_corroborate_its_own_pid`, `a_fresh_pid_per_hook_never_becomes_a_repeat`, `a_reaper_parent_is_no_answer`.

## Visual verification

- [x] Not a visual change.

## Why not `sysinfo`

It would replace the 93 LOC of platform readers with ~12 and drop 3 `unsafe` blocks. Measured on a 1206-process box: **1911 µs** (pid-scoped refresh) or **22578 µs** (full) per walk against **2.4 µs** native, plus +9 transitive crates in the shim whose contract is never to block the agent. Hand-rolled deliberately; the numbers are in the commit message.

## Checklist

- [x] I read the relevant `CLAUDE.md` (root + `pixtuoid-core`'s).
- [x] Docs updated in the same commit — new `SHARP-EDGES.md` entry + regenerated `CLAUDE.md` index via `just gen-guides`.
- [x] Checked against the recurring pitfalls — the parallel-implementation one caught two real defects, fixed in 4bd9da2: `doctor --focus` still printed "getppid on unix", and `cli_pid`'s "fix bugs in BOTH readers" note covered only the Windows twin.

## AI assistance


IvanWng97:

Five lenses were dispatched (the prompt's escalation list matched three triggers: state-machine/liveness seam, substantial comment reshaping, and `pixtuoid-hook` touched). The first five backgrounded agents returned nothing — a dispatch is not a review, so they were **re-run as two focused lenses against the final head**, both of which returned real reports and both verdicts **REQUEST-CHANGES**. Plus the online bot at two heads. Every finding below is at a terminal state.

### FIXED — defects in the fix itself (the ones that mattered)

| Finding | Source | Verified how |
|---|---|---|
| Unix walk had no time budget; `arm_watchdog` hard-exits at 200 ms and is armed *before* `cli_pid`, so a stall cost the **whole envelope** | bot HIGH | Read `transport.rs:22` + `main.rs:87`; budget hoisted to all platforms |
| Pid map grew **unbounded** — `take_pid` was the only remover and ran only from the exit-watch drain, so uncorroborated pids were never removed | mine, verifying the bot's premise | One entry per tool call, forever, for exactly the population the fix exists for |
| **opencode lost its primary exit path** — its `_pid` is `process.pid` stamped *inside* the CLI (not a guess), and a session answered with no tool call reports it once, so it never armed. Its module doc: `server.instance.disposed` isn't decoded, "the pid-watch covers instance teardown" | lens A | `pid_bind_target` now returns the channel's verdict — corroborate a guess, arm a fact on sight |
| **An agent could be armed on two pids at once** — with `armed`/`candidate` in separate maps, sighting an armed pid skipped the candidate bookkeeping, so 9,7,9 armed 9 non-consecutively | lens A + lens B | Replaced by one `Sighting` enum per agent; both states now unrepresentable |
| Interposer list matched whole names — Git-Bash's `bash.exe` never matched `bash`, Alpine's `/bin/sh` reports `busybox` | lens B | Bare stems + `.exe` strip + busybox |
| **My batch test had no teeth** — it passed with the dedupe deleted, because no decoder emits two bind-target events for one agent | lens B | Confirmed by mutation; extracted `bind_targets` and pinned the forward guard directly |

### FIXED — claims that out-ran the code

"costs nothing" (it costs one event, and more where the walk is unstable) · the binder list (omitted opencode) · "Identity on every activity event" (turn-end arms are deliberately Identity-less) · "one per live session" (one per agent id ever seen) · codewhale's Windows reason (a pid *is* resolved there since #528 — what's missing is an `ExitWatch` backend) · hermes's "only when" · **and the PR description itself**, which claimed Cursor gains a working abrupt-exit signal. Instrumenting the live shim showed the walk resolves the real `cursor-agent` for most hooks but not all, so that signal is best-effort. Corrected in the body, the sharp edge, and `cursor.rs`.

Stale-premise comment sweep took **four rounds** — `doctor.rs`, then `main.rs` (bot), then `registry.rs` + 6 more (bot), then `WHERE-TO-LOOK.md` + `LAYOUT.md` (lens B). Root cause of the repeats: each sweep grepped remembered phrases instead of enumerating the population of sites documenting `_pid`'s provenance.

### REFUTED / not findings

- The gate's load-bearing claim — that CC/Codex/grok keep an authoritative pid via `jsonl::liveness::ProbeSnapshot` with its own `ExitWatch` — **verified independently by both lenses** (`claude_code/native.rs:103`, `codex/native.rs:111`, `grok/native.rs:195`, `jsonl/mod.rs:342`).
- Reversed-range panic in the `/proc` parse: `str::get` on a reversed range returns `None` (checked against the std source in-session, not from memory), and the kernel's stat format can't produce the input anyway.
- Creation polarity: a `SessionEnd` for an unknown id creates nothing (`synthesize_hook_registration` only fires for Activity/Waiting).
- Detached walk thread: sound — `mpsc` send to a dropped receiver is an `Err` that's swallowed; detached threads don't block exit.
- Version adjudication: **no bump rides this PR** — no public surface moves, and `0.17.0` is already an open unreleased minor.

### Verification

`just preflight` 0 · `just doc-check` 0 (CI-only gate, run locally — this adds intra-doc links) · 2982 tests · `cargo check` clean for `x86_64-pc-windows-msvc` and `x86_64-unknown-linux-gnu` · CI green incl. `windows-test` · live `cursor-agent` runs show no mid-session walkout in any of three.

Not merging — a human merges, per the repo gate.

IvanWng97:
## Review round 4 — the two late lenses reported

The comment lens and the shim never-panic lens (dispatched for two of the three escalation triggers) came back after the earlier rounds. Both **REQUEST-CHANGES**, both correct. Every finding is now at a terminal state.

### FIXED — blocking

**The shim's panic gate, removed by my own fix.** Moving the walk off-thread to fix the bot's HIGH meant a panic there no longer fails anything — the detached thread just drops the sender, and `recv_timeout` reads it as a failed walk. The lens injected a panic and measured it: **6 of 7 child-process tests RED before the move, 7 of 7 GREEN after.** And the profiles disagree — `panic = "abort"` (release, what ships) makes it SIGABRT while CI's debug build exits 0, and both write the panic to the stderr the agent CLI reads, breaching the *silently* half of invariant #5 either way. `catch_unwind` would have been the wrong fix: inert under abort, greening the test build while production still aborts.

Fixed properly: `parse_stat` split from the `/proc` read and compiled everywhere, so hostile rows are table-testable off Linux (14 now are, including the reversed-range case); and `clippy::indexing_slicing` added to the crate's deny list beside unwrap/expect/panic — the actual panic surface of a parser over process rows. Cost one change: `&raw[..end]` → `.get(..end).unwrap_or(raw)`.

**A fabricated trace in the sharp edge.** I justified the `Sighting` enum with *"9,7,9 armed 9 non-consecutively"*. Replayed against the two-map code it describes, `9,7,9` never arms — that repro belongs to the **pair-keyed design from two revisions earlier**. Right conclusion, invented mechanism, in the one entry a future agent will trust. Replaced with the trace that does reproduce: arming 9, then corroborating 7, left the agent in both pid sets.

**A false rationale.** `pid_bind_target` claimed TranscriptProbe sources bind "once per session". They don't — the shared decoder arms prepend `Identity` to every tool payload (`decoder.rs:394/410/425/435`), which CC/codex ride, so ungated they would corroborate and *arm*. The gate stands on its own terms; the invented consequence is gone.

### FIXED — accuracy and redundancy

`cli_pid`'s module doc still said `HookPidWatch` ends the session when the wrapper pid dies — the last pre-#896 claim, in the fix's own module doc, false twice over (corroboration refuses it; Windows has no `ExitWatch` backend). Plus the "every platform" universals (other unix keeps a bare `getppid`), the reaper doc saying `1` where code and test cover `0`, the Windows-only `NtQueryInformationProcess` rationale on a cross-OS list, a missing note that a new pid *demotes* an armed agent, and three redundancy cuts the lens counted sentence-by-sentence (including one sentence that had already rotted once inside this PR).

### Verified, not findings

The lens confirmed clean: `args_os` (with the #198 rationale in-comment), no prod unwrap/expect/panic, stdin bounded, watchdog armed before our work on both platforms, and ~30 adversarial inputs against the release binary all exiting 0 — including 200k-deep nested JSON, 2 MB of non-UTF-8, and a scalar straddling `ENV_FIELD_CAP` exactly. My own independent sweep agreed.

### DEFERRED — pre-existing, not this PR's to fix

Three shim findings that predate this branch, surfaced by the audit. Flagging rather than filing unilaterally:

- **`paths.rs:10,22`** — non-UTF-8 `PIXTUOID_SOCKET` / `XDG_RUNTIME_DIR` are silently *ignored* rather than honored (`env::var` errors on non-UTF-8; a Unix socket path is bytes). An operator's override is dropped and the event goes elsewhere, silently. `var_os` is the fix. Confidence 65.
- **`main.rs:58`** — `fn main() -> Result<()>` has no `?` today, but the signature means a future one would print to stderr and exit 1. Confidence 40.
- **`main.rs:122`** — `current_dir()` in the env arm runs before `arm_watchdog`. Resolves from the dentry cache in practice. Confidence 35.

### Tooling note for anyone reviewing this branch

The shim lens found `rtk`'s git wrapper reported `git diff --numstat HEAD` as empty while the file demonstrably differed, and invented `Cargo.toml`/`Cargo.lock` changes not in the diff. It re-derived every line number through `/usr/bin/git` by absolute path. Worth knowing if a later reviewer sees phantom dependency changes.

PR #898, opened 2026-08-10
docs(cursor): tool_use_id IS on the wire — say why we drop it anyway

IvanWng97:
## Summary

`cursor.rs` claimed `tool_use_id` is *always `None`*. Captured against `cursor-agent 2026.06.12`, that's false — and the real situation is more interesting than the claim.

**Stacked on #897** (base is `fix/896-cursor-sprite-flapping`, not `main`) because both touch `cursor.rs`'s module doc. Independent change: wire shape, not pid plumbing.

## What the wire actually does

Every `preToolUse` carries an id that pairs with its `postToolUse` / `postToolUseFailure`, and **tools interleave**:

```
preToolUse  Read   tool_b3a89c28     ← Read starts
preToolUse  Shell  069272b1          ← Shell starts before Read finishes
postToolUse Read   tool_b3a89c28
postToolUse Shell  069272b1
```

So the id looks exactly like what the reducer's per-call machinery wants.

## Why we drop it anyway

`Task` fires `preToolUse` and **never** a post. Verified in a delegating run — the one unpaired id out of eleven, with `subagentStart`/`subagentStop` silent throughout:

```
preToolUse  Task   tool_381d7536     ← no post, ever
UNPAIRED (no post): id=tool_381d7536  events=['preToolUse']
```

Passing the id through would satisfy `track_active_tasks`' `Some(tuid)` + `is_task()` arm, insert a tuid nothing ever drains, and strand the parent **Delegating for the rest of the session** — `apply_activity_end` re-enters Delegating on every later tool end while `any_active_task` holds.

And the machinery is inert here regardless: hook-wins dedup needs a second transport (cursor is hook-only), and precise wait-resolution needs permission events this source never emits.

## What changed

- The false universal replaced with the true claim plus the hazard, so the next reader doesn't "fix" it.
- A test pinning the drop, with the reason in its failure message. **Negative-controlled**: wiring the id through — the exact fix the old text invited — turns it red.
- `transcript_path` now arrives on the wire (`null` at `sessionStart`, set from the first tool event), so a future watcher needn't reconstruct the path.

Note the drift watch covers cursor's event *names* (`CURSOR_EVENTS`) but not field shapes, which is why this went unnoticed.

## Type of change

- [x] Docs only (+ one test)

## How I tested it

`just preflight` 0 · `just doc-check` 0. Payloads captured from a live `cursor-agent` run, not from the published docs.

## Checklist

## AI assistance


IvanWng97:
Superseded by #901 — same single commit, rebased onto main.

#898 was branched off #897, which was **squash**-merged, so its history no longer matches main and updating it in place would need a force-push. Reopened on a clean base instead of rewriting the branch. No content change; the diff is byte-identical.

PR #899, opened 2026-08-10
fix(paths): read PATH-valued env vars as bytes, and gate the class shut

IvanWng97:
## Summary

`env::var` returns `Err(NotUnicode)` for a value that is not UTF-8 — which a filesystem path is never required to be on Unix — so **every path override we read with it was silently dropped**, and the resolver fell through to a different directory. Same failure shape as #880/#343/#342/#195 (override ignored, empty office, no error), reached through the *encoding* rather than the precedence.

Found while auditing the shim for #897, not from a user report.

## Proven before fixing

```
env::var("HOME")    -> Err(NotUnicode("/tmp/caf\xFF"))   <- a legal path, dropped
env::var_os("HOME") -> Some("/tmp/caf\xFF")
```

and at repo level `user_home_opt()` returned `None` for that `HOME`. **The first commit pins that bug against the unfixed code**; the second flips the assertion — so this is a real before/after pair, not a claim.

## Three layers, because two of them are one-time and one is forever

**The read.** `platform::path_env` / `path_env_trimmed` are the one authority (the shim keeps a deliberate twin — it cannot depend on core — pinned equal by `socket_path_parity.rs`). 22 path reads across 3 crates go through them. Blank-filtering and hermes's `.strip()` mirror moved to that boundary.

**The type.** `String` cannot hold a path, so the readers are only half of it: `user_home`/`user_home_opt`/`resolve_user_config_dir`/`OmpEnv`/hermes's and reasonix's cores/`expand_tilde` and ~30 call sites move to `PathBuf`/`&Path`. Breaking for `pixtuoid-core`, which is **free**: `v0.16.0` is the last release and `0.17.0` is already open and unreleased. Only NAME-valued vars (`PI_PROFILE`, `OMP_PROFILE`, `PI_CONFIG_DIR`) stay `String` — they are not paths.

The compiler then found two more instances on its own: `crash.rs` and `logging.rs` built paths by string concatenation (`format!("{state}/pixtuoid/log")`), which also breaks on Windows separators.

**The gate** — the half that makes this a class fix rather than a bigger instance fix. `just env-paths` fails any `env::var` on a path-named var, with the remedy in the message. Wired into `just lint` **and the CI hygiene job, which enumerates recipes and does not run `just lint`** — a check only in `lint` would have been inert in CI. Negative-controlled twice: `--selftest` proves the matcher fires on 5 forms and stays silent on 5 legitimate ones, and a planted `env::var("CODEX_HOME")` in production code turns the recipe red.

## Where reviewers should be most suspicious

Moving blank-filtering to the read boundary made ~8 pure-core assertions fail, **and I deleted them**. I believe every deletion is correct — those cases now belong to `path_env`'s own tests, which this PR adds. But "the refactor made tests fail and the author deleted them" is indistinguishable from a bad refactor when viewed from outside. **Please check each deleted assertion against the new `path_env` tests.** The affected tests are in `platform.rs`, `hermes.rs`, `opencode.rs`, `reasonix.rs`.

One deliberate exception, marked in place: doctor's home-split advisory renders these for a human, so it reads bytes and converts lossily at the **display** boundary.

## Type of change

- [x] Bug fix (+ a new gate)

## How I tested it

`just preflight` 0 (2984 tests) · `just doc-check` 0 · `just env-paths` 0 · `cargo check` clean for `x86_64-pc-windows-msvc` (workspace) and `x86_64-unknown-linux-gnu` (core + hook; the workspace's linux cross needs `alsa-sys`, unavailable on this host).

## Checklist

- [ ] **Two-lens review not yet run** — requested below.

## AI assistance


IvanWng97:
## Disposition sweep — at the final merge head `83cddbbb`

Two differentiated lenses ran on this PR; both returned **REQUEST-CHANGES**. Every finding is now at a terminal state, and the sweep is at the head being merged (not an earlier one).

### FIXED

| finding | source | what it was |
|---|---|---|
| `api-surface-check` RED | both lenses | The public-API golden still carried the pre-PR `Option<String>` signatures. CI-only gate — **I had claimed "all gates green" without running it.** Regenerated. |
| `doctor.rs:787` — a surviving path read | both lenses | `env::var(v)` where `v` is the registry's `home_env`. Not a literal, so my own gate couldn't see it — **and I'd asserted "0 remaining" on that gate's authority.** Also a real defect: with a non-UTF-8 override the resolver honoured it while doctor read it as unset, dropping `via $VAR` and silencing the ⚠ in exactly the case this PR makes diagnosable. |
| Gate blind over 44% of the tree | lens 2 | A one-way `#[cfg(test)]` flag tripped on a mid-file `#[cfg(test)] const` and on the string inside a `//!` doc comment — including `install/openclaw.rs`, a config-PATH resolver, 98% invisible. Now skips only a `#[cfg(test)] mod` body by brace depth. Coverage 55.5% → 64.6%; verified by planting a read at `openclaw.rs:130` that the old walker missed. |
| Selftest controlled the wrong half | lens 2 | It exercised the regex against string literals and never called `offenders()` — **the gate could have scanned zero lines and still printed ✓.** Now plants a three-file fixture and asserts on the walk. |
| hermes's `.strip()` mirror pinned by nothing | lens 1 | The deleted assertions moved the trim to `path_env_trimmed`, but nothing bound hermes to the trimming twin — swapping it left the suite green. Pinned, and mutation-verified as the only test that catches the swap. |
| Windows tests asserted a path's STRING form | CI | `#[cfg(windows)]` tests compared `default_socket_path()` (now `PathBuf`) against `r"\\.\pipe\..."`. **My cross-check ran `cargo check --target windows` without `--all-targets`, so test targets were never built.** Structural comparison now; re-verified with `--all-targets` for the hook and the whole workspace. |

Plus the stale prose both lenses caught: reasonix's doc (false in both clauses), a codex test name still claiming "empty", omp's assertion message describing the wrong reason, and `resolve_user_config_dir` — `pub`, and silently no longer filtering — now states its precondition.

### REFUTED / verified, not findings

Per-deletion audit of every removed assertion (lens 1): all absorbed by the new `path_env` tests except the hermes one above. Reader correctness on non-UTF-8 (replacement chars are never whitespace, so a bad value can't be blank-filtered; the value kept is the original `OsString`). `expand_tilde`'s `&str` → `&Path` rewrite regresses no input, and fixes `~//claw` which used to absolute-join-clobber. Shim/daemon parity intact across all branches. Resolution POLICY unchanged for well-formed UTF-8 across all nine resolvers — the filter's move from body to boundary is a no-op because every caller is `path_env`-fed. `OmpEnv`'s path/name split is correct. Breaking-API argument sound: crates.io newest is 0.16.0, workspace is 0.17.0, `cargo semver-checks` passes.

### Gates at this head

`preflight` 0 · `doc-check` 0 · `env-paths` 0 · `api-surface-check` 0 · `semver` 0 · `gen-contract` 0 (no diff) · `gen-wasm` regenerated and committed · CI 48/48 pass including all three Windows jobs · bot **Findings: 0** at `83cddbbb`.

Issue #907, opened 2026-08-11
comment-lint: the fn-body arm is walk-past-able, and the obvious gate was reverted for cause


IvanWng97:
## The problem, stated honestly

`scripts/comment-lint.py`'s fn-body arm (3+ consecutive `//` or `#` lines inside
a function body, via ast-grep) prints findings and exits 0. On PR #904 it printed
**18** findings, the author read them, reported the run as passing, and then
added a 13th violation an hour after clearing twelve. An advisory the author can
read and walk past did not hold.

The obvious remedy was tried in that PR and **reverted**, with two reproductions:

1. **It fails merged code.** Run the blocking gate against merged commit
   `5ffb7aaa` on main: **exit 1**, on `crates/pixtuoid-hook/src/main.rs:95` — one
   sentence, one idea, about the ordering constraint behind architecture
   invariant #5, flagged only for wrapping past two lines at 100 columns. The
   tree carries ~1,400 such runs across ~180 files, all grandfathered, and code
   MOTION re-adds them: a reindent, or a function moved verbatim into another
   file, both trip it. With no suppression mechanism and `--no-verify` forbidden
   here, an author's only escape is deleting a comment the repo wants kept.
2. **Its selftest could not see the regression it caused.** The first attempt
   inferred the requirement from `shutil.which("ast-grep")`, which broke the
   protected `comment-gate` CI job on every PR (that job carries no npm by
   design). Putting that bug back verbatim and running `--selftest`: **passes**.
   The selftest exercises `gate_fails`, a pure predicate; the breakage lived in
   `main()`'s wiring, which it never invokes.

The disqualifying property, which should have been priced first: a
3-consecutive-lines rule is a LENGTH cap, and `gate_fails`' own docstring already
argues at length that a length cap contradicts the semantic rule it serves —
which is precisely why the doc-run arm reports rather than blocks.

## So what would work

Not a mechanical length cap. Candidates, none costed yet:

- **A semantic unit.** The rule is about IDEAS, not lines (CLAUDE.md comment test
  #2: delete each sentence after the first; if nothing is lost, cut). Nothing
  mechanical can count ideas, which is why the repo assigns this to the comment
  lens of the two-lens review.
- **An escape hatch**, so a legitimate long WHY can be kept: a baseline file, or
  an explicit marker. That makes blocking survivable, at the cost of a second
  thing to keep honest.
- **Motion-blindness.** Most of the false positives are pure code motion. A rule
[...]

Issue #902, opened 2026-08-11
bug: creatures walk onto the wall through an 8px channel 2px west of the elevator


IvanWng97:
Reported by the owner against **0.16**: "有时小猫会走到墙上… 好像有一个垂直的区域在电梯左边"
(sometimes the cat walks onto the wall — there seems to be a vertical region to the
left of the elevator).

Reproduced and root-caused. The code is byte-identical in `v0.16.0` and on `main`.

## Root cause

`crates/pixtuoid-scene/src/layout/mask.rs`, the door cut:

```rust
const DOOR_CUT_W: u16 = 8;                        // ELEVATOR_W is 16
let cut_x = d.x.saturating_sub(2);                // starts 2 px WEST of the elevator
let cut_h = top_margin.saturating_add(OBSTACLE_PAD_PX);
mask.mark_walkable(cut_x, 0, DOOR_CUT_W, cut_h);  // from y = 0
```

Dumping the mask at 192x160 (`door.x = 174`, elevator spans `174..190`) gives a
walkable channel at **x 172..179, from y = 0 down through the whole top band**:

```
  y=  0 ##########################........############
  ...                             ^^^^^^^^
  y= 32 ##########################........############
        x=146                     x=172  x=179
```

Three separate problems in those three lines:

1. **`cut_x = d.x - 2` puts columns `d.x-2` and `d.x-1` outside the elevator
   sprite.** They are painted bare wall and marked walkable — this is the
   "vertical region left of the elevator" the report names.
2. **The cut starts at `y = 0`,** the very top of the buffer, not at the wall
   band's base. The channel runs the full height of the band.
3. **`DOOR_CUT_W = 8` against `ELEVATOR_W = 16`,** so the cut covers only the
   elevator's left half and is offset off-centre; the right half is blocked.

`creatures::walkable_target` picks a wander destination uniformly from walkable
cells, so a pet or the gateway mascot lands in the channel and walks up it. The
sprites are centre-anchored, so the creature also renders `frame_h / 2` rows
above its own point.

## Why nothing catches it

`SHARP-EDGES.md`'s spawn-threshold entry already documents this exact blind spot
for a different subject: *"move the spawn 8 px north and it stays `is_walkable`
at 264/264 swept layouts (the wall band's blocked rows END at `wall_band_h()`),
while `find_path` and `ReachSet::from_mask` both SNAP a displaced seed back into
the component — so every connectivity/routability guard above stays green on a
[...]

PR #910, opened 2026-08-12
fix(scene): four review-fallout defects — a shadow off its desk, a cat under the furniture, a mute refusal, a blind crop

### LINKED ISSUE #905: creatures: place_creature's recovery snap lacks the visual-clearance filter its twin gained
## What

`creatures::walkable_target`'s fallback snap now re-checks `is_visually_clear`
(#902's fix — a snap restores WALKABILITY only, and the cell under the fish
tank's glass is walkable by invariant #6). `place_creature` does the same shape
and did not get the filter:

```rust
fn place_creature(p: Point, extent: (u16, u16), layout: &Layout) -> Point {
    let clamped = clamp_sprite_inside(p, extent, layout);
    if layout.walkable.is_walkable(clamped.x, clamped.y) {
        return clamped;
    }
    // no is_visually_clear here
        .unwrap_or(clamped)
}
```

So the edge clamp can push a creature off a visually-clear cell, and the
recovery snap can put it back down on a walkable-but-decorated one — a pet under
a desk's overhang or the aquarium's glass.

## Why it is not urgent

The reachable set is narrow: the snap only fires when the clamp landed on a
NON-walkable cell, which is near the canvas border, and the rest sprites are 6 px
wide (`lo = 3`), so on the swept goldens column 3 is open floor wherever column 4
is blocked. Two reviewers tried and neither could construct a reachable case.

## Why it is filed anyway

#902's fix hardened ONE of the eight `snap_point_to_walkable` call sites in
`creatures.rs`, and the sharp edge it added now reads as if the class is closed.
It is not: this is the same predicate gap, one function over.

## Sketch

Give `place_creature`'s snap the same `.filter(|p| layout.is_visually_clear(*p))`
its twin has, and decide deliberately what the fallback is when nothing clears
(the twin falls through to `door_threshold`, which is open floor by
construction). Then sweep the remaining six snap sites and say, per site, whether
the filter applies — a per-site answer, not a blanket one, since a WALK endpoint
is explicitly exempt (walking through decor is accepted; `creatures.rs`'s own
comment says so).

Found by the correctness lens of PR #904's review (F7), which rated it LOW and
pre-existing.

### LINKED ISSUE #906: layout: the desk shadow centres on DESK_W/2, not on the painted desk's midline
## What

`pixel_painter::mod::desk_shadow_ellipse` centres the desk's shadow on
`desk.x + DESK_W / 2`, while the desk SPRITE's midline is
`desk.x + desk_furniture_def().visual.w / 2` — and `visual.w` is `DESK_W + 4`.
So the shadow sits 2 px west of the desk it belongs to.

## Why now

PR #904 fixed exactly this class for the SEAT: `DESK_WALK_X_OFF` was a second
formula that agreed with the painted desk by arithmetic rather than by
construction, and `seat_center_x` replaced it by deriving from
`desk_furniture_def().visual.w`. The shadow is the same defect two functions
away, and it was left standing because it is pre-existing and outside that PR's
scope.

## Sketch

Read the same authority the seat now reads. Note the shadow is a DESK property,
not a seat one, so it must NOT pick up `seat_center_x`'s per-desk jitter — the
right expression is the midline without the nudge. Worth checking whether any
other painter still carries a `DESK_W`-derived x that should be `visual.w`.

Art regeneration follows (`just gen` + `just gen-wasm`), and `gen-check`'s 0.0%
tolerance will show the 2 px move.

Found by the design lens of PR #904's review, rated pre-existing and out of scope
there.

### LINKED ISSUE #908: tui: a terminal under 31 rows renders a black office with no explanation
## What

At **80×24** — the classic terminal default — pixtuoid renders a BLACK office.
Footer only, no message, no hint. The user sees a broken program.

Bisected on both `df2924d9` and PR #904's head, identically: **30 rows fails, 31
rows succeeds**. `snapshot --cols 100 --rows 30` prints "(debug_walkable) layout
too small to compute" — so the layout REFUSES (correctly; `MIN_LAYOUT_H` is a
documented minimum), and the refusal simply never reaches the screen.

Pre-existing; PR #904 does not move the threshold. Found by that PR's
runtime/user-journey lens, which was the first review pass in a while to actually
run the binary at several terminal sizes.

## Why it is worth fixing

The failure mode is indistinguishable from a crash or a rendering bug. Every
other refusal in this codebase says what it wants — `doctor` names its probes,
the pack loader warns about missing animations. This one is silent, at the size a
first-time user is most likely to be at.

## Sketch

Where `render_floor` (or its caller) gets `None` from the layout, paint the same
kind of centred notice the empty-office state already has ("— office empty —"),
naming the actual requirement: the minimum rows/cols and the current size. The
constants are `compute::MIN_LAYOUT_W` / `MIN_LAYOUT_H`, so the message can be
derived rather than hardcoded.

Worth checking at the same time whether the floating window and the wasm hero
degrade the same way, since they share `render_floor`.

### LINKED ISSUE #909: snapshot: --crop-agent cannot frame any desk-seated agent, including every seated pose
## What

`snapshot --crop-agent <label>` fails with "agent … has no visual position" for
**any desk-seated agent**. Reproduced with a negative control: `--crop-agent
thinking` fails the same way, `--crop-agent couch-act` (a waypoint occupant)
succeeds.

Cause: `PoseHistory` records only waypoint and walking poses — `pose/mod.rs`'s
`_ => None` arm — so the seated poses never get a recorded position for the tool
to frame.

## Why it matters more than it looks

`.claude/skills/beautify-decoration/SKILL.md` makes visual verification a
requirement for sprite work, and `--crop-agent` is the tool for it. So the repo's
own instrument cannot frame the majority of its characters — every agent at a
desk, in every pose.

Concretely: PR #904's headline change was the WAITING state, whose whole point is
to pull the user's eye, and that PR's visual verification had to work around this
by cropping regions by hand. A reviewer following the skill would hit the same
wall.

Pre-existing — `StandingAtDesk` was equally unrecorded before that PR replaced it
— so this is not a regression, just a gap nobody had named.

## Sketch

Either record the seated poses in `PoseHistory` (they have a position:
`Seat::at_desk(desk, facing).render_anchor(w)`), or have `--crop-agent` fall back
to deriving the anchor the same way `anchors::character_anchor` does, which
already answers for every pose. The second is probably right — the history exists
for motion continuity, not for tooling, and widening it for a dev tool would put
a render concern in a sim store.

Found by PR #904's runtime/user-journey lens.

IvanWng97:
The four defects #904's review deferred rather than dropped. #907 — the
comment-lint arm — is not here: it changes CI policy, not the engine, and it
still has no remedy I'd defend. Each fix below was negative-controlled.

## #906 — the desk shadow fell 2 px west of its desk

`desk_shadow_ellipse` keyed its centre on `DESK_W / 2`. `DESK_W` is the desk's
SURFACE width; the side cabinets that actually touch the floor make the painted
and ground-contacting width `visual.w = DESK_W + 4`. The same function's `cy`
already read `visual.h` — one function, two authorities.

This is the class #904 fixed for the seat (`DESK_WALK_X_OFF` → `seat_center_x`),
sitting two functions away. Every axis now comes off the one furniture row, with
`half_w` expressed against it too: identical pixels, one source.

**The old position was pinned.** `assert_eq!(e.cx, desk.x + DESK_W / 2)` asserted
the defect — the same shape as the golden that certified #902's wall channel for
as long as it existed. Retargeted, plus a second pin requiring the shadow to stay
inside the ground contact it falls under.

## #905 — a resting creature could settle under furniture, and it was REACHABLE

Two reviewers rated this theoretical, because both looked at `place_creature`.
The reachable path is upstream: `walkable_target` judged the RAW draw, but
`snap_point_to_walkable` answers with the coarse cell's **centre**, so a clear
draw still comes back on a decorated cell.

Judging the snapped point fixes it — and because snapping is idempotent
(`cell_of(cell_center(c)) == c`), every downstream `snap(dest)` becomes an
identity, so the walk lands exactly where the rest is. By construction, not by
three separate filters, and the no-pop contract between leg end and rest pose is
preserved rather than patched.

Reverting it reds the new sweep at `192x160 seed 1 cycle 0` — a real cat under a
real sprite.

Per-site answer, which is what the issue actually asked for:

| site | verdict |
| --- | --- |
| `walkable_target` | judges the SNAPPED point now |
| `walk_between` endpoints ×2 | exempt — walking through decor is documented as fine |
| `mascot_elevator` | a walk terminus, same exemption |
| `mascot_home` | REST beat, and the corridor hosts appliances — filtered |
| `place_creature` | filtered for consistency, **not** a demonstrated trigger; removing it reds nothing, and its comment says so |
| `rest_pos`, `mascot_wander` rest | now identities, via the upstream fix |

## #908 — a terminal under the layout minimum rendered a black office

80×24 — the classic default — got a footer over nothing. Indistinguishable from
a crash, at the size a first-time user is most likely to be at. Bisected on the
base commit too: pre-existing, 30 rows fails and 31 succeeds.

Both refusal paths already funnelled through one helper, so the notice goes
there. It names what it needs and what it got, and **degrades to a short form**
on a terminal too narrow to hold the sentence — skipping the line would leave
exactly the blank screen this exists to prevent, which my first attempt did until
the 15×8 case caught it.

`pixtuoid_scene::layout::min_layout_size()` is a new public fn: the requirement
belongs to the layout, and only a painter can turn buffer units into rows.

## #909 — `--crop-agent` could not frame any desk-seated agent

Including #904's own headline `waiting` state, because `PoseHistory` records only
waypoint and walking poses. The repo's visual-verification skill names this tool
as the way to do the job.

Fixed at the tool, not the store: it falls back to the seat anchor the painter
would use, rather than widening a sim store for a dev concern. Four labels that
failed now frame; `couch-act`, which already worked, still does.

## Gates

`just preflight` 0 · **3,029 tests** · `gen-check`, `doc-check`,
`api-surface-check`, `gen-wasm-check` 0 · art and wasm regenerated, since both
the shadow and the rest positions move pixels.

Negative controls, each observed red:

- shadow `cx` back to `DESK_W / 2` → both shadow pins
- `walkable_target` back to judging the raw draw → the rest sweep, with the seed
- the too-small notice removed → the refusal pin
- (and the one that did NOT fire: removing `place_creature`'s filter changes
  nothing, which is why its comment calls it consistency rather than a fix)

One thing worth knowing about the diff: while adding the rest sweep I inserted a
doc comment between an existing `#[test]` and its fn, which silently unregistered
that test. `clippy` caught it as `never used` — recorded because a test that
stops running is invisible to a green suite.

Closes #905
Closes #906
Closes #908
Closes #909

https://claude.ai/code/session_018DvCSoZKuAokN1vMXRu5wP


IvanWng97:
## Review round — 1 HIGH, verified and fixed in `f257735e`

**Finding**: `mascot_home`'s new `.filter(is_visually_clear)` turned "the corridor centre is decorated" into `None`, and `mascot_position` takes it through `?` — so the whole gateway mascot stopped rendering, every frame, rather than resting a few px off.

**Verified before fixing** (the repo's rule — a reviewer finding is a hypothesis until it reproduces). Swept every layout from `min_layout_size()` up (+220 × +160, five desk caps → 19,980 layouts with a corridor):

```
before:  316 of 19,980 lose the mascot entirely
filter removed:  0     <- causation: it was the filter, not a pre-existing snap failure
```

The finding was not theoretical. It was live at 316 sizes.

**Why this one mattered more than a normal off-by-a-pixel**: `mascot_home` is *deterministic per layout* — it is the leg-0 origin, the fixed point where the elevator walk-in ends and wander cycle 0 begins (that hand-off is why it exists at all; after cycle 0 destinations come from the whole floor via `walkable_target`). So `None` is not a dropped frame. It is that layout having no mascot, permanently. And the corridor is exactly where the appliances stand, so the fixed landing point is *structurally* liable to be decorated.

**Fix** follows `walkable_target`, the house pattern for this shape — prefer a clear cell, but never fail. It steps outward along the corridor band, falling back to the centre. Re-swept:

```
after:  none=0  clear=19,980  fell_back=0
```

So the fallback arm does not fire at any swept size. It stays because a `None` here *deletes* the mascot rather than moving it — not a tradeoff decor should be able to make. The sibling filter in `place_creature` already had exactly this shape (`.unwrap_or(clamped)`); `mascot_home` was the one site that could answer "nothing".

**Test** pins the path the reviewer flagged as uncovered, at three of the 316 sizes — and asserts the fixture still *decorates* that centre, so it cannot pass vacuously if layout drift makes those sizes clear. Controlled: restoring the old `.filter()` reds it.

`just preflight` green, 3030 tests.

IvanWng97:
## Correction — the fix above was the wrong shape; `d54bdef9` supersedes it

Owner review caught that my fix guarded a failure mode that should not exist. It was right, and the code says so plainly once you look at the leg-0 branch:

```rust
let prev = if cycle == 0 { home } else { walkable_target(layout, seed, cycle) };
```

`home` **was** `walkable_target(seed, 0)` with special logic bolted on — leg 0 was the one leg whose origin came from somewhere else. Everything downstream followed from that one special case: the corridor coupling, the `Option`, the `?` in `mascot_position`, the `is_visually_clear` filter, and then my 20-line outward search plus a never-firing fallback to undo the damage the filter did.

My justification for the filter was also **factually wrong**. I wrote *"A REST beat … unlike the elevator endpoint, which is a walk terminus."* `mascot_wander` never rests at `home` — it rests at `dest` (the `else` branch); `home` appears only as `prev`, a walk ORIGIN, and as the terminus of the enter lerp. It is exactly as much a walk terminus as the elevator endpoint, so by the file's own stated rule it should never have taken the filter.

**What landed instead:** delete the special case. `prev = walkable_target(seed, cycle)` for every cycle; the enter walk ends at draw 0. The hand-off stays pop-free for the same reason as before, without a second mechanism.

```
                        before      after
production lines        +20         -103 / +23  (net -79)
mascot_home             Option      deleted
mascot_position `?`     3           2
sweep: layouts losing the mascot     316 → 0
```

The class is now **unrepresentable rather than guarded**: `walkable_target` returns a `Point`, not an `Option`, so there is no longer a `?` for decor to trip.

**Free inheritance**: #905's "judge the SNAPPED point" fix lives inside `walkable_target`. The mascot now stops resting under appliances by taking the corrected shared path — which is what my filter was clumsily reaching for.

This also unifies the mascot with `pet_position`, which never had a special origin:

```rust
// pet_position                                  // mascot_wander (after)
dest = walkable_target(l, pet_seed, cycle_n);    dest = walkable_target(l, seed, cycle + 1);
prev = walkable_target(l, pet_seed, cycle_n-1);  prev = walkable_target(l, seed, cycle);
```

What still differs between them is lifecycle and tempo — the mascot enters from the elevator and leaves on `Down`; its cycle length carries daemon state now that destinations no longer do. Those are real domain differences, not duplication, so no shared helper was extracted.

The corridor keeps exactly one job: `mascot_elevator`'s door fallback.

Comment rot fixed in the same pass — `MASCOT_ENTER_STAGGER_MS` claimed "the seed reaches only the steady wander", which this change makes false (the seed now picks the walk-in terminus too, so two gateways share only the door cell, not the whole line).

**Gates**: `just preflight` green, 3029 tests. `just gen-check` green including `lobster.png` — the mascot's committed still is unchanged, since `home` only ever affected the enter window and cycle 0.

IvanWng97:
## Two-lens review — disposition sweep at `bdb9d2ad`

Both lenses landed on the same **HIGH**, and it was in `fe95ef89`'s consistency change — the one whose own comment called it a no-op.

### HIGH — the recovery snap's filter put creatures back inside furniture

`place_creature` reaches its snap **only** when `clamped` is not walkable. So `.filter(is_visually_clear).unwrap_or(clamped)` discarded the snap and returned a cell already known to be furniture.

Verified independently before fixing — through `mascot_position`, 3 sizes × 8 seeds × 300 samples:

```
with the filter     257 / 7086 frames anchored off the walkable mask
without it           12 / 7086
```

A lens measured it wider: **2176 of 2254** changes swap a walkable point for a non-walkable one, across **97 of ~100** swept sizes. First hit is a `lobster_walk` frame — visible motion clipping through furniture, not a rest.

My comment was false on both halves:

> *"for consistency rather than for a demonstrated trigger: no swept size reaches this snap with a decorated cell, so removing it reds nothing."*

97 sizes reach it, and "removing it reds nothing" was true only because nothing tested the branch — which is how it shipped. A false universal inside a don't-touch-this note is the worst place for one.

**FIXED** by deleting the filter, and the WHY that edit had displaced (why the second `clamp_sprite_inside` is there at all) is restored.

### The rest

| Finding | State |
|---|---|
| `walkable_target`'s exhausted-draws arm returned `door_threshold` **unsnapped**, so the "snapping is idempotent, every downstream `snap(dest)` is an identity" comment three lines up was false on that path | **FIXED** — snapped, so the claim holds for every return |
| `SHARP-EDGES.md` said the fallback re-checks `is_visually_clear`; it has not since the check moved into the draw loop | **FIXED** — now describes the ladder that exists |
| `docs/mutants-exclusions.md` justified an equivalent mutant by "the walk-in's endpoint IS `home`" — the mechanism this branch deleted | **FIXED** — conclusion still holds, for a better reason; the old wording was the note-rots-into-an-unre-derivable-survivor class |
| `d54bdef9` deleted `f257735e`'s regression test and replaced it with nothing; "the disappearing-mascot class is unrepresentable" **overstated** it — `mascot_position` still `?`s on `mascot_elevator` | **FIXED** — `every_narrow_layout_still_draws_its_mascot`, and the claim is restated as narrowed, not closed |
| `draw_footer_only_frame` has **three** callers, not two | **FIXED** |
| `MASCOT_ENTER_STAGGER_MS` claimed seeded walk-in lines always pull apart; **2% of seed pairs draw the same terminus** — including 0/450, the fixture its own test uses | **FIXED** |
| `need_rows` lacked the `.max(MIN_SCENE_HEIGHT)` its twin has; and nothing pinned that the size the notice NAMES actually lays out (`need_rows + 1` passed every test) | **FIXED** — both |
| `mascot_wander_cycle0…` doc's second sentence was deducible from the assertion below it; a test local was still named `home` | **FIXED** |
| The **walk-OUT** reconstruction on `Down`, `walk_between` being A\* (not a straight lerp), speed, determinism, the desk-shadow geometry, `min_layout_size()` on the semver surface, the no-extraction call, the regenerated art set | **REFUTED** — attacked and held. Mean walk-out pop actually *improved* 30%; the new enter leg is on average **slower** than the old one and sits inside the busy-leg envelope that already ships |
| `place_creature`'s second clamp can still push a snapped point off the mask (12/7086) | **ISSUE #912** — pre-existing; `place_creature` is byte-identical to its base-commit form, and fixing it means choosing which constraint yields at the edge |
| The too-small notice is covered by the onboarding modal at 80×24 — the exact first-run case #908 exists for | **ISSUE #913** |
| #909's fallback arm: a planted `panic!` passes all 28 tests | **ISSUE #914** |

### One process note

`d54bdef9` deleted `f257735e`, which was the fix for the bot's HIGH. To a thread reader that reads as a silent drop, so stating it plainly: the deletion **supersedes** the fix and closes the failure mode harder — `walkable_target` returns a `Point`, so the decor trigger cannot produce a `None` at all. What it did NOT close is the class, hence the replacement test above.

### Gates at `bdb9d2ad`

`just preflight` green, **3032 tests**. `just gen-check` green including `lobster.png`.

One thing no gate covers, flagged by the design lens and worth stating rather than burying: the only committed baseline that renders a mascot stamps `entered_at = now − 20s` *specifically* to skip the walk-in, and this PR changes only the walk-in and cycle 0. **The visual gates are structurally blind to it.** I have not watched the new entry animation, and no check will.

IvanWng97:
## The one gate nothing covers: I watched it

I flagged above that the visual gates are structurally blind to this change — the only committed baseline that renders a mascot stamps `entered_at = now − 20s` *specifically* to skip the walk-in, and the walk-in plus cycle 0 is the entire behavioural delta. Flagging it isn't doing it, so:

Instrumented `inject_openclaw_presence` behind a temp env var (reverted; tree clean, `git status` empty), rendered `--openclaw idle` at 192×80 across the enter window, and isolated the lobster by largest connected red component:

```
age=   0ms   NOT DRAWN                      <- its own stagger; correct
age= 600ms   blob 1472   at (1347, 455)
age= 950ms   blob 3456   at (1295, 551)
age=1400ms   blob 3456   at (1199, 655)
age=1900ms   blob 3392   at (1095, 751)
age=2400ms   blob  704   at (1035, 823)
age=2900ms   blob 3392   at ( 871, 847)
age=3300ms   blob  704   at ( 819, 759)     <- past the enter window, wandering
```

**Read as a strip, it looks right.** The path is continuous and monotone — no teleport, no pop at the enter→wander hand-off (which the numbers already predicted: `walk_between(draw0, draw1, 0.0) == draw0`). The sprite is drawn whole at every frame; nothing is clipped against a buffer edge, which was the specific worry, since draw 0 can now be anywhere on the floor rather than mid-corridor.

The blob shrinking to ~700 at 2400 ms and 3300 ms is **occlusion, not clipping** — the lobster passing behind a desk. That is architecture invariant #6 working as documented: the walkable mask is the ground footprint only, so sprites legitimately overhang and get painted over.

Not a substitute for a gate — nothing here is automated, and the next person changing walk-leg behaviour will have to do this again. It is what the repo's "render and WATCH it before the verdict — this fires even when no committed art changes" asks for, and it says the new walk-in is fine.

IvanWng97:
## Final round — the bot found a HIGH in #909's fallback, and the missing test was the cause

Head `93088436`. Both findings verified against the source before fixing.

**HIGH — the seated fallback read floor 0's geometry for every agent.** `desk_index` is a GLOBAL index, and `single_floor_local()`'s own doc says it holds only *"within a single-floor scene … For a multi-floor scene go through `SceneState::floor_local_desk`"*. The snapshot tool builds `SceneState::uniform`, so any run with more agents than `--max-desks` seats the tail on floor 1+. For those the fallback resolved the wrong desk index **and** built the layout from floor 0's capacity with the raw `--floor-seed`, where every production caller derives `floor_seed(floor_idx)`.

The symptom is the misleading half: `--crop-agent` reports the agent *"is neither placed nor at a home desk"* while it sits at one, on a floor this code never looked at.

Controlled — with the floor-0 assumptions restored, the new test fails with exactly that message:

```
floor 1 agent "couch-act" must crop: agent "couch-act" is neither placed nor at a home desk
```

**MEDIUM — `compute_crop_rect` had no test at all**, pre- or post-#909, which is why a planted `panic!` in that arm passed all 28. That is the same gap I had just filed as #914 — and it is not a coincidence that the untested arm is where the defect was. #914 is closed by the fix; its secondary note (the sixth near-duplicate `buf_w`/`buf_h`/`compute_with_seed` preamble) is untouched and stays a smaller cleanup.

**CI note:** three jobs went red on `93088436` from `extractions/setup-just` failing to reach the GitHub API (`GET /repos/casey/just/releases - 500`, `self-signed certificate`), with `required-lint-jobs` and `ci-gate` following as aggregators. Infrastructure, not the diff — re-run, now 49 SUCCESS / 2 SKIPPED.

### Where this PR ended up

| | |
|---|---|
| CI | 49 SUCCESS / 2 SKIPPED / 0 failures |
| Online bot at head | **Findings: 0** |
| `just preflight` | green, **3033 tests** |
| `just gen-check` | green |
| Local lenses | 2 differentiated, both with demonstrated findings, all dispositioned |
| Walk-in animation | rendered and watched — no visual regression |
| Deferred | #912 (pre-existing clamp/walkable at the buffer edge), #913 (notice covered by the onboarding modal) |

Four review rounds found four real defects in this branch, three of them mine and two inside the fixes for the previous round's findings. Worth stating plainly rather than burying: the branch is better for it, but the pattern — a fix introducing the next finding — is the signal that the *reviewing* was doing the work, not the writing.

PR #911, opened 2026-08-12
test(comment-lint): price the reverted gate, and give the selftest eyes on main()

IvanWng97:
## What this is

#907 asked for teeth on `comment-lint`'s fn-body arm. **The answer is that
mechanical teeth are not available**, and this PR is the evidence plus the two
things that *are* fixable.

Per the plan agreed on the issue, the negative control came first. It killed
every candidate #907 listed — including one the issue got factually wrong.

## The measurement

`just comment-lint-replay 20` replays today's rules against already-merged
commits, so each finding is a block a would-be gate **would have imposed on work
that shipped**:

```
fn-body arm (advisory)   11/20 merged commits, 179 lines
blocking arms (--gate)    0/20 merged commits          <- the instrument's own control
```

Which prices all four candidates:

| Candidate | Verdict | Evidence |
|---|---|---|
| Semantic unit | not mechanizable | by definition — it counts ideas, not lines |
| Threshold calibration | no separation point | 86 runs at N=20 (3–10) and 818 at N=60 (3–12), **continuous, no gap**; #904's own slop sat at 3 |
| Motion-blindness | **0.0–1.1%** | 0–2 of 179 lines, and both matches are the bare string `//`. **#907 claims "most of the false positives are pure code motion" — wrong by a wide margin**; the issue text is corrected. |
| Escape hatch | **NOT PRICED — verdict withdrawn** | My rejection rested on "the author walks past the advisory", which the one observable counterfactual contradicts: on #904, 14 of 18 findings were acted on and the restructure stuck. See the round-2 disposition. |

A 3-consecutive-lines rule is a LENGTH cap, and `gate_fails`' docstring already
argues at length that a length cap contradicts the semantic rule it serves.
The data says the same thing quantitatively.

## What lands

**1. `--selftest` drives the real `main()`.** #907's second reproduction was
that the selftest could not see the regression it caused: it pinned
`gate_fails`, a pure predicate, while the breakage lived in `main()`'s wiring.
It now runs the script as a subprocess in both PATH states — the ast-grep-free
one being exactly what the protected `comment-gate` CI job carries — and pins
the exit-code contract in both directions.

Three controls, each applied then reverted after confirming it reds the
selftest:

| Mutation | Reproduces | Result |
|---|---|---|
| `docs_only and gate → return 1` | #907's CI breaker (absence ⇒ block) | selftest FAILS ✓ |
| `blocking \|= bool(new_hits)` | the reverted length cap | selftest FAILS ✓ |
| `gate_fails → False` | blocking arms going silent | selftest FAILS ✓ |

**2. `just comment-lint-replay N`** — on-demand, `fuzz`-tier, not in CI. The
next proposal to make an arm block is priced in one command instead of
re-derived. `CONTRIBUTING.md` names it where it explains why the arm stays
advisory.

**3. The advisory's COUNT is a candidate list, not a status line**
(`pr-review.prompt.md`). This is the half of #904 that was never a detection
failure: the arm printed 18 findings, they were reported as a passing run, and
the diff then gained another. The disposition contract already covers
*findings*; what was missing is that this count IS the candidate list the
comment lens owns.

## The instrument was wrong three times

Recorded because the numbers above are only worth what the harness is:

1. The first replay didn't inject `scripts/gitenv.py`, so `import gitenv` raised
   on every commit predating it, `2>/dev/null` swallowed the traceback, and the
   empty output parsed as **0 findings** — under-reporting by half (6/20).
2. Injecting the scripts instead made the replay count **the tool's own
   comments** as the commit's added lines — over-reporting.
3. The hardened guard then immediately caught a valid verdict I had not
   enumerated (a commit touching no Rust/Python).

The committed recipe aborts on an unparseable verdict rather than counting zero.

## Gates

- `just lint` — 17/17 green, exit 0
- `just comment-lint-gate` — green; this diff adds **zero** new fn-body runs
- `just comment-lint-replay 20` — reproduces the table above
- No Rust touched, so clippy/test surface is unaffected

## Why #907 stays open

It closes the two halves that are settled — the selftest gap the issue names in
its second reproduction, and the pricing of three of its four candidates. The
fourth, the **escape hatch, is unpriced**: my rejection of it was withdrawn under
review because its stated reason ("the author walks past the advisory") is
contradicted by the one observable counterfactual.

Closing #907 here would retire an option I explicitly declined to price. #907
stays open, re-scoped to that one candidate.

Refs #907

---

## Review rounds

**Round 1 — online bot at `c0889587`: Findings: 2 (1 high, 1 medium). Both verified before fixing, both fixed in `4e83c393`.**

| # | Finding | Verified how | Disposition |
|---|---|---|---|
| HIGH | The replay aborted on a *healthy* commit: when only the doc-run / re-parent / prose arm fires, `main()` prints a per-finding line and returns early, matching neither verdict shape the recipe enumerated. | Reproduced on a throwaway repo — a diff adding one 14-line `///` run prints `comment-lint: src/x.rs:1 — new 14-line doc-comment run (max 10)` and exits 0, which the recipe called a broken replay. | FIXED — the classifier now discriminates on what separates a run from a crash (any `comment-lint:` line), checked against all five real verdict shapes. |
| MEDIUM | The `HEAD~2` assertion passed for the wrong reason: each prose fixture is its own commit, so `HEAD~2` was a prose fixture and the gate blocked on `over_prose`. The assertion would have survived the re-parent arm being wholly broken. | Read the fixture history; confirmed `prose_case` commits per fixture. | FIXED — each case now gets its own fixture commit so `HEAD~1...HEAD` isolates the arm it names. Controlled both ways: stubbing `reparented_doc_hits` to `[]` reds it, stubbing `prose_share` does not. |

**Did the HIGH finding contaminate the published numbers?** No — it was fail-*loud* (`exit 1` aborting the whole run), not a silent skip, so the 20-commit measurement either completed or errored. It completed, and re-running after the fix gives the same 11/20 and 179.

**Round 2 — online bot at `4e83c393`: Findings: 0.**

Local differentiated lenses (correctness/grounding · design & blast-radius · adversarial measurement-validity) were run in parallel; the third exists because this PR's entire argument is one measurement, and the author's own instrument was wrong twice before it held.

## Gates at `4e83c393`

- CI: 49 SUCCESS, 2 SKIPPED, 0 failures
- pre-push: 3025 tests passed
- `just lint` 17/17, `just comment-lint-gate` green (this diff adds zero new fn-body runs)

### DIFF
--- a/.github/prompts/pr-review.prompt.md
+++ b/.github/prompts/pr-review.prompt.md
@@ -407,7 +407,7 @@ history:
   relocate-to-the-declaration-or-a-CLAUDE.md-sharp-edge call is this lens's.
   Its COUNT is that candidate list, not a status line: report N items each
   carrying a disposition, never "the run passed" — on #904 eighteen findings
-  were reported as a pass, and the diff then gained another.
+  mid-PR were reported as a pass, and the diff then gained another.
   **Redundancy-within is the axis this lens kept missing**: the other four
   compare the comment to the CODE, and a restatement repeats no code, so
   nothing else can flag it. Apply CLAUDE.md's comment test #2 verbatim — do
--- a/.github/prompts/pr-review.prompt.md
+++ b/.github/prompts/pr-review.prompt.md
@@ -405,6 +405,9 @@ history:
   the diff's new lines only — the ~5k
   pre-existing legitimate WHY are grandfathered), but the keep / trim-to-≤2 /
   relocate-to-the-declaration-or-a-CLAUDE.md-sharp-edge call is this lens's.
+  Its COUNT is that candidate list, not a status line: report N items each
+  carrying a disposition, never "the run passed" — on #904 eighteen findings
+  were reported as a pass, and the diff then gained another.
   **Redundancy-within is the axis this lens kept missing**: the other four
   compare the comment to the CODE, and a restatement repeats no code, so
   nothing else can flag it. Apply CLAUDE.md's comment test #2 verbatim — do
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -273,9 +273,8 @@ path matching that posts the documented blast-radius escalations as a sticky PR 
 prose-only escalation can't be silently skipped (#198); and `just comment-lint`'s ast-grep
 arm, whose npm install keeps it in advisory `ci-supplemental` where a registry outage cannot
 become a required check. A proposal to make that arm BLOCK is priced by `just
-comment-lint-replay N`, which replays today's rules against already-merged commits, so each
-finding it reports is a block that proposal would have imposed on work that shipped (#907).
-Merged is not adjudicated — read the flagged lines before calling them false positives.
+comment-lint-replay N` (#907). Merged is not adjudicated — read the flagged lines before
+calling them false positives.
 
 Which `comment-lint` arms BLOCK is stated once, in `gate_fails`' docstring in
 `scripts/comment-lint.py` — pinned in both directions by that script's selftest.
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -273,8 +273,9 @@ path matching that posts the documented blast-radius escalations as a sticky PR 
 prose-only escalation can't be silently skipped (#198); and `just comment-lint`'s ast-grep
 arm, whose npm install keeps it in advisory `ci-supplemental` where a registry outage cannot
 become a required check. A proposal to make that arm BLOCK is priced by `just
-comment-lint-replay N`, which replays today's rules against already-merged commits, where
-every finding it reports is a false positive by construction (#907).
+comment-lint-replay N`, which replays today's rules against already-merged commits, so each
+finding it reports is a block that proposal would have imposed on work that shipped (#907).
+Merged is not adjudicated — read the flagged lines before calling them false positives.
 
 Which `comment-lint` arms BLOCK is stated once, in `gate_fails`' docstring in
 `scripts/comment-lint.py` — pinned in both directions by that script's selftest.
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -272,7 +272,9 @@ Advisory backstops that surface risk but NEVER gate: `scripts/check_upstream_dri
 path matching that posts the documented blast-radius escalations as a sticky PR comment so
 prose-only escalation can't be silently skipped (#198); and `just comment-lint`'s ast-grep
 arm, whose npm install keeps it in advisory `ci-supplemental` where a registry outage cannot
-become a required check.
+become a required check. A proposal to make that arm BLOCK is priced by `just
+comment-lint-replay N`, which replays today's rules against already-merged commits, where
+every finding it reports is a false positive by construction (#907).
 
 Which `comment-lint` arms BLOCK is stated once, in `gate_fails`' docstring in
 `scripts/comment-lint.py` — pinned in both directions by that script's selftest.
--- a/justfile
+++ b/justfile
@@ -1260,13 +1260,17 @@ comment-lint-replay n="20":
       cp -R "$root/.ast-grep" "$root/sgconfig.yml" "$tmp/wt/"
       rc=0
       out="$(cd "$tmp/wt" && python3 "$root/scripts/comment-lint.py" "$sha^" --gate 2>&1)" || rc=$?
-      # A crash prints no verdict line, and would otherwise count as zero.
-      case "$out" in
-        *"new comment-slop"*) n_hit="$(printf '%s' "$out" | sed -n 's/^comment-lint: \([0-9]*\) new comment-slop.*/\1/p')" ;;
-        *"no new 3+-comment runs"*|*"no added/changed Rust or Python lines"*) n_hit=0 ;;
-        *) echo "comment-lint-replay: no verdict from ${sha:0:8} — the replay is broken, not the commit" >&2
-           printf '%s\n' "$out" >&2; exit 1 ;;
-      esac
+      # Every path through the script prints a `comment-lint:` line; a crash
+      # prints none, and would otherwise be counted as zero findings.
+      if printf '%s\n' "$out" | grep -qE '^comment-lint: [0-9]+ new comment-slop'; then
+        n_hit="$(printf '%s\n' "$out" | sed -n 's/^comment-lint: \([0-9]*\) new comment-slop.*/\1/p')"
+      elif printf '%s\n' "$out" | grep -q '^comment-lint: '; then
+        n_hit=0
+      else
+        echo "comment-lint-replay: no verdict from ${sha:0:8} — the replay is broken, not the commit" >&2
+        printf '%s\n' "$out" >&2
+        exit 1
+      fi
       total=$((total + 1)) flagged=$((flagged + n_hit))
       if [ "$rc" -ne 0 ]; then
         gated=$((gated + 1))
--- a/justfile
+++ b/justfile
@@ -1287,6 +1287,9 @@ comment-lint-replay n="20":
         printf '  %s  %-4s %s\n' "${sha:0:8}" "$n_hit" "$(git log -1 --format=%s "$sha" | cut -c1-52)"
       fi
     done
+    # A bad N leaves the `for` word list empty without tripping `set -e`, and the
+    # summary then reports its most reassuring shape.
+    [ "$total" -gt 0 ] || { echo "comment-lint-replay: no commits walked — check N" >&2; exit 1; }
     echo "fn-body arm (advisory): $advisory/$total merged commits, $flagged lines"
     echo "blocking arms (--gate): $gated/$total merged commits — an arm reds the"
     echo "  commits that predate its own calibration, so read this against \`git log -S\`"
--- a/justfile
+++ b/justfile
@@ -1238,17 +1238,21 @@ comment-lint-gate:
     python3 scripts/comment-lint.py origin/main --gate
 
 # Prices any proposal to make an arm BLOCK: today's rules replayed against
-# already-merged commits, where every finding is by construction a false
-# positive. On-demand like `fuzz` — N checkouts, and it needs origin/main.
+# already-merged commits, so each finding is a block the proposal would have
+# imposed on work that shipped. On-demand like `fuzz` — N checkouts, origin/main.
 [group('meta')]
 [doc('Replay the comment-lint arms against the last N merged commits (default 20)')]
 comment-lint-replay n="20":
     #!/usr/bin/env bash
     set -euo pipefail
+    # Without it the fn-body count is 0 for every commit and the run still exits
+    # 0 — the fail-open `fuzz` guards for the same reason.
+    command -v ast-grep >/dev/null \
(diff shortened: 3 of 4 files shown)


IvanWng97:
## Independent check of the load-bearing claim

The PR's conclusion rests on **claim B**: that the 179 flagged lines are false positives *by construction* because the commits are merged. That is the one claim whose failure would invert the whole argument — if a real share were genuine slop, the arm would be finding problems, not noise. So it should not rest on "they merged, therefore they're fine."

Sampled deterministically (every 13th flagged line across the five largest contributing commits, 13 of 168), each read against the code it describes:

| Location | What the comment does |
|---|---|
| `sprite/format.rs:440` | why digits-only, when `u16::from_str` accepts `+` — two spellings of a lookup KEY = two files a renderer picks between arbitrarily |
| `cutaway/order.rs:104` | why the index tie-break exists — a topological order is not unique, and reshuffling equal-depth pieces flickers |
| `cutaway/paint.rs:696` | why the front face samples the sprite's own material instead of a theme role (pack owns the colour; custom `--pack-dir` gets a match for free) |
| `cutaway/paint.rs:858` | what the previous code got wrong — raw frame = twelve copies of one person; `mirror_vertical` = upside-down west-facing |
| `cutaway/paint.rs:964` | why saturating: density is bounded, the BASE is not |
| `examples/cutaway_snapshot.rs:182` | why the anchor is lifted a full line — `draw_text_at` takes a TOP y |
| `examples/render_bench.rs:184` | why the RGB→base64 arithmetic does not describe SIXEL |
| `validate.rs:42` | why those three name classes are pack input and get sanitised |
| `check_upstream_drift.py:1677` | why word-boundary and not the quoted form the other checks use |
| `floor/tests.rs:1014` | why the assertion is at the LAYER, so a new `DrawableKind` inherits it |
| `pixel_painter/mod.rs:588` | why the pre-floor group exists — the corridor runner stayed daylight-tan in a dimmed office |
| `comment-lint.py:103` | why the fixture lengths are LITERAL and not derived from the constant |
| `comment-lint.py:250` | why the reset per hunk header — an open `/*` charges another file's code as prose |

**13 of 13 are legitimate WHY** — each carries a constraint a reader cannot deduce from the line below it, which is CLAUDE.md's test #1. None is restated WHAT, none is narration. Claim B holds on this sample.

This matters beyond this PR: it means the fn-body arm's 179 hits are not a backlog of deferred cleanup. They are the repo's deliberate WHY density, wrapping past three lines at 100 columns — which is precisely why a line-count rule cannot gate it.

IvanWng97:
## Review round 2 — three differentiated lenses. Disposition sweep.

Every finding drives to one terminal state. Head is now `eed2bfb5`.

### The one that matters: I broke it in round 1

**MEDIUM — the round-1 verdict parser reads a MATCH as a miss.** Two lenses demonstrated this independently. Under `set -o pipefail`, `printf '%s\n' "$out" | grep -q …` has `grep -q` quit on the first hit, the writer take SIGPIPE, and `pipefail` turn the pipeline non-zero. Both the `if` and the `elif` fail identically, so the recipe aborts on a **healthy** commit — and only once the verdict passes the pipe buffer, i.e. on the loudest, most informative commits.

Reproduced standalone under bash (the recipe's shell — my first attempt used zsh and did NOT reproduce, which is worth recording):

```
pipe    n=100    -> MATCHED
pipe    n=200000 -> MISSED     <- false negative
here    n=200000 -> MATCHED
```

`comment-lint-replay 40` died on `8383b1de` (1134 findings) — a commit the *pre-fix* `case` counted correctly. **Round 1 traded one abort class for an earlier one.** Fixed with a herestring; `N=40` now completes (21/40, 1379 lines).

### Findings → terminal state

| Finding | Lens | State |
|---|---|---|
| SIGPIPE false negative in the classifier | 1 + 3 | **FIXED** |
| No ast-grep precondition — run completes and prints `0/20, 0 lines`, exit 0 | 1 + 2 | **FIXED** — `fuzz`'s guard shape adopted |
| Isolating the e2e cases cost `main()`'s prose coverage (`over_prose = False` survived) | 1 | **FIXED** — prose case added |
| Recipe parses for a `comment-lint:` prefix that nothing pinned | 2 | **FIXED** — clean-path fixture + assertion |
| `[ "$gated" -eq 0 ]` hard-fails on its own measurement | 1 + 2 + 3 | **FIXED** — assertion dropped, output reworded |
| "false positive by construction" | 2 + 3 | **FIXED** — withdrawn from recipe + CONTRIBUTING |
| `renamed.rs` is a must-NOT-fire and a must-fire fixture 100 lines apart | 2 | **FIXED** — own file |
| `advisory.rs` is fn-body-only by arithmetic, not construction | 1 | **FIXED** — `code_floor` |
| stderr captured and discarded | 1 + 2 | **FIXED** |
| `path_without` docstring overclaims "other tools intact" | 2 | **FIXED** |
| "eighteen findings" not reproducible from merged history | 2 | **FIXED** — "mid-PR" |
| trap installed after `git worktree add` | 2 | **FIXED** |
| The ast-grep-free row can't catch "the advisory arm became blocking" | 1 + 2 | **REFUTED-deliberate** — documented at `comment-lint.py`'s skip note; `ci-supplemental` runs the full selftest with ast-grep. But the round-0 commit message implies coverage the protected job doesn't have — stated correctly here instead. |

Mutant battery at `eed2bfb5`, all six red: re-parent unwired · prose unwired · #907's length cap · #907's CI breaker · blocking arms silent · **a renamed verdict line**. That last is new coverage, and only the *findingless* fixture kills it — the three finding-producing ones can't.

### Corrections to my own numbers

| Claim | Was | Is |
|---|---|---|
| Legitimate runs | 84 | **86** at N=20 (3–10); **818** at N=60 (**3–12, still continuous**) |
| Motion-blindness | 3/179 (1.7%) | **0–2/179 (0.0–1.1%)** — both matches are the bare string `//`; my `git grep -F` was a substring test that over-counts |
| Blocking arms on merged main | 0/20 | 0/20 holds, but **3/40** — and those reds are commits that predate `PROSE_MAX_PCT` (merged 2026-08-10 vs 2026-08-04), so they measure an arm's calibration reach, not a regression |

Both corrections move the threshold-calibration and motion-blindness verdicts in the same direction they already pointed. Neither changes the conclusion.

### One verdict I am withdrawing

**The escape hatch is now UNPRICED, not rejected.** My rejection read "rubber stamp by construction" — which inherited the same invalid warrant as claim B. The one case where the counterfactual is observable cuts the other way: on #904, **14 of 18 findings were acted on and the restructure survived** a later revert of the gate itself. And #845 shows 4,439 flagged lines deleted after they merged. "The author walks past the advisory" is not what the history shows.

I am not re-opening #907 on this: the arm still cannot gate mechanically (that rests on the threshold and motion numbers, both re-verified and strengthened). But the escape-hatch row in the PR body now says *withdrawn* rather than *rejected*, because I did not price it and my stated reason was wrong.

### What survived scrutiny

Claim A reproduced independently by two lenses with their own parsers (11/20, 179, per-commit identical). Claim B's *sampled* form survived 35 independent reads across two reviewers — 13 mine, 22 Lens 3's, spread across all run lengths including the length-3 bucket where #904's own slop sat — with no slop found. Shell robustness (`set -e` arithmetic, trap-on-SIGINT, `mktemp -d`, worktree registration, overlapping runs, `{{ n }}` interpolation vs 16 house-style siblings) verified sound. Comment discipline: all six added comments pass CLAUDE.md test #2, no measurement in any of them.

`just lint` 17/17, replay green at N=20 and N=40.

IvanWng97:
## Review round 3 — both lenses found the same hole from opposite sides

Head `3875164f`. Round 2 pinned that **some** `comment-lint:` line gets printed; the replay depends on **which**. Two lenses demonstrated that gap independently.

| | Demonstration | State |
|---|---|---|
| **MEDIUM** — the pin goes vacuous exactly where the protected CI job runs. `main()` prints the ast-grep-absent notice *before* any arm, so with ast-grep off PATH a crash downstream still leaves a prefixed line. Sabotaging the re-parent report loop was **green** in that state, with `rc != 0` satisfying the block expectation for the wrong reason. | Lens A, `raise IndexError` in the report loop | **FIXED** |
| **MEDIUM** — the COUNT format was unpinned. Dropping one word from `{n} new comment-slop finding(s)` kept every gate green while the replay tallied **0 for all 20 commits** — the same silent zero the ast-grep guard just closed, arriving through the other door. | Lens B, one-word edit + `comment-lint-replay 4` reporting 0/4 against a truth of 3/4 | **FIXED** |
| **LOW** — `path_without` prunes a whole DIRECTORY. On this box git survives (`/usr/bin/git` vs `/opt/homebrew/bin/ast-grep`), but on nix, Linuxbrew, or a distro shipping ast-grep in `/usr/bin` both share one and the child dies in `gitenv.clean_env()`. | Lens A, fakebin with both symlinked | **FIXED** — named failure, not a traceback under the wrong label |
| **LOW** — a bad `N` leaves the `for` word list empty without tripping `set -e`; the summary then printed `0/0` and exited 0. | Lens B, `just comment-lint-replay ""` | **FIXED** |
| **LOW** — CONTRIBUTING restated the recipe header's rationale verbatim. That file's own norm, two paragraphs down, is the anti-duplication rule with its own war story. | Lens B | **FIXED** — only the unique sentence kept |
| **NIT** — with no ast-grep at all the two PATH states are byte-identical, so 4 of 8 child runs were exact duplicates under a misleading label. | Lens A, measured | **FIXED** — second state dropped when identical |

Each case now names the line **its own arm** emits, selected per PATH state (the fn-body case has no line to emit without ast-grep).

### Mutant battery — ten, all red

```
re-parent unwired        prose unwired            #907's length cap
#907's CI breaker        blocking arms silent     clean verdict reworded
re-parent report loop    ast-grep wiring deleted  count line reworded
prose report block broken
```

The last four are new. The first six still red, so the sharper assertion did not trade away what round 2 bought — that check mattered, because sharpening an assertion is exactly how coverage silently narrows.

### One process note against myself

My mutant harness reported `#907 CI breaker` as a SURVIVOR. It was not — my shell passed `\n` literally, producing a syntax error, and my pass/fail check only looked for the string `FAILED`, which a traceback does not contain. Re-run with a file-based mutation and an `ast.parse` check: **red**. Third time this session that my own instrument was the thing that was broken, which is the argument for controlling the harness and not just the subject.

### Verified clean by the lenses (not findings)

Per-case isolation instrumented at `HEAD~1...HEAD`: advisory 3 prose/23 code → no arm; wedge → re-parent only; prose 9/20 → prose only; clean 0/20 → nothing. The `HEAD~2`/`over_prose` false-pass class is gone. Subprocess wiring correct (`sys.path[0]` resolves `gitenv` independently of `cwd`; ambient `GIT_*` scrubbed by `gitenv.git()`). Protected job: +1.15s, no new dependency. Windows: no CI exposure, and `os.path.normpath` does not normalize case there — which `:315`'s inert-control turns into a loud failure rather than a silent pass. Replay order-independence verified by isolated per-commit runs matching their in-sequence numbers. All ten added comments pass CLAUDE.md test #2; no measurement in any of them.

One reading caveat worth carrying forward: `flagged` counts **lines**, and a long run yields overlapping ast-grep windows, so 179 lines is materially fewer distinct comment runs (86). The output says "lines"; it only misleads if quoted as "179 comments".

`just lint` 17/17.

PR #915, opened 2026-08-12
refactor(scene): the sim never moves a position to fit a sprite

### LINKED ISSUE #912: creatures: place_creature's second clamp can push a snapped point back off the mask
## What

`place_creature` (`crates/pixtuoid-scene/src/creatures.rs`) recovers an unwalkable anchor by snapping, then clamps the snapped point back inside the buffer:

```rust
    .unwrap_or(clamped)
```

The second clamp exists because snapping is free to move the point back out past the edge the first clamp pulled it in from. But that clamp can itself push the point off a walkable cell — the two goals (keep the sprite fully in frame / stand on real floor) genuinely conflict near the buffer edge.

## Measured

Through `mascot_position`, three sizes × 8 seeds × 300 time samples = 7086 drawn frames:

| | frames anchored off the walkable mask |
|---|---|
| with the (now removed) clearance filter | 257 |
| without it — i.e. today, and the base commit | **12** |

So ~0.17% of drawn frames. First seen at `112x100`, seed 3, 1600 ms, `Point{105,46}`, on a `lobster_walk` frame — visible motion, not a rest.

## Why it is not fixed in #910

`place_creature` is byte-identical to its form at `e3928053`, so this predates that PR; #910 only removed a filter it had added. Fixing it means choosing which of the two constraints yields at the edge — clip the sprite, or stand in furniture — which is a design call, not a cleanup.

Pinned meanwhile: `a_placed_creature_takes_the_recovery_snap_it_found` asserts the recovery snap is taken whenever it survives clamping, and skips the cases described here rather than silently passing over them.

## Fix sketch

Three candidates, none costed:
- Prefer walkability: return the unclamped snapped point and let `blit_centered` clip the sprite at the edge.
- Prefer the frame: search inward from the clamped point for the nearest walkable cell that also satisfies the clamp.
- Accept it and say so in `SHARP-EDGES.md` — an edge-adjacent creature overlapping by a few px may simply not read as wrong at half-block scale, which a render would settle.

IvanWng97:
## The defect is a direction, not an off-by-one

`place_creature` clamped a creature's position so its sprite would fit the canvas, then tried to recover with a snap when the clamp landed off the walkable mask. **The recovery could not work**: the clamp only fires where the point is already at the edge, so the snap's answer got clamped straight back. That is #912.

The real problem is which way the dependency ran. "Does the whole frame fit on the canvas" is a **rendering** question, and it was being answered by moving a **sim** position — onto furniture. The painter already owns that question and answers it correctly (`keep_sprite_on_canvas`, applied to the pet and the mascot at enqueue), because moving a *drawn* position harms nothing.

So the sim stops answering it.

```
sim    walkable_target → a walkable point, never touched by sprite size
paint  keep_sprite_on_canvas → moves the DRAWN position, never the sim's
```

`place_creature` and `clamp_sprite_inside` are deleted, and **`mascot_position` no longer takes the pack** — it cannot see a sprite's size, so it cannot be tempted to use it. The rule is a signature now, not a comment.

## Grounded in the industry layering, checked not remembered

[Bevy's `Anchor`](https://docs.rs/bevy/latest/bevy/sprite/struct.Anchor.html) is *"Normalized (relative to its size) offset of a 2d renderable entity **from its Transform**"*, and **the Transform is not affected by sprite size**. Position and sprite geometry are separate components; geometry never edits position. That is exactly the seam this PR restores.

## The alternative was measured, not argued away

Making the east/west edges real walls in the mask also fixes #912 — the office would simply not extend to the buffer edge, symmetric with north/south, which are already fully blocked (row 0 and the last row have **zero** walkable cells).

```
side walls in the mask     132 of 797 tests fail   (the layout packs desks to the edge)
this change                  1 test relocated
```

Worth keeping for the cleanup, where a layout rework absorbs that cost once instead of paying it for nothing.

## The test was the violation in test form

`a_wandering_mascot_sprite_never_spills_past_the_buffer_edge` asserted, **in the sim**, that a sprite fits the canvas — it even recomputed `pos - fw/2` itself. Split along the same seam:

| layer | guarantee |
|---|---|
| sim | `a_wandering_mascot_always_stands_on_walkable_ground` |
| paint | `the_painter_keeps_every_mascot_frame_on_the_canvas` |

Writing the sim half **caught me over-claiming**: my first version asserted every frame stands on walkable ground, and it failed immediately. A walk leg rides a coarse-router polyline and crosses furniture between cell centres *by design* (`walkable_target`'s own comment: "Walking through one stays fine."). Only settled frames can carry that guarantee, and the test now says so.

Both halves controlled: stubbing `keep_sprite_on_canvas` to identity reds the paint half on a 1px spill (`14x12 sprite drawn at Point { x: 186, y: 54 } spills the 192x160 canvas`).

## Gates

| | |
|---|---|
| `just preflight` | green, **3034 tests** |
| `just gen-check` | **pixel-identical** |
| clippy | 0 warnings |
| net | **−204 / +116** |

`gen-check` being unchanged is the load-bearing check here: the drawn positions never moved, only who decides them.

## Known, and deliberately NOT in this PR

**Agents get no on-canvas guard at all.** Measured over 4096 aimless destinations per size, **2% (192×160) and 7% (112×100)** would shift or clip at the edge — a larger gap than the one this PR closes. The fix belongs at the same paint layer, but agents appear in the committed stills, so it moves art and wants its own change with a regen.

Also unchanged and worth naming: `pet_position` still takes the pack, for a *presence* check (`pack.animation(kind.walk_anim())?` — does this pack have a cat), not a geometry one.

## Why this does not retire #912

**Correction, found in review and confirmed by my own measurement.** This PR does not change where a creature is DRAWN — `gen-check` being pixel-identical is the proof, and I shipped that as a virtue without noticing it also means #912's visible symptom survives. The painter's nudge performs the same move the sim clamp used to, and does not re-check the mask:

```
settled mascot frames, at head        192x160: 0 of 1316 drawn on a non-clear cell
                                      112x100: 41 of 1316 (3.1%)
```

What this PR fixes is the direction: the SIM's answer is now always real floor, so nothing downstream — the hover box, the z-sort row, the next leg's origin — inherits the compromise. The residual is a genuine conflict between two goals within half a sprite of the edge, and it belongs to the painter.

`Refs #912`, which stays open for that residual.


IvanWng97:
## Two-lens review — disposition sweep at `5b5acc75`

Both lenses found things I had missed **about my own claims**. The first is the discipline I applied everywhere else in this arc and skipped on my own test.

### The one that matters: my test could not fail

`a_wandering_mascot_always_stands_on_walkable_ground` swept `192x160`, where the deleted clamp fires 123 times and **every one lands on clear floor**. A lens restored the exact pre-#912 `place_creature` and the whole suite stayed green — the regression the test's own doc names walked straight through it.

Reproduced, then fixed at `192x80` (the repo's own snapshot geometry), verified both directions:

```
pre-#912 code + 192x80   RED   port 18904 at +20640ms: a resting lobster_rest at
                               Point { x: 185, y: 66 } is not clear floor
this PR      + 192x80   GREEN
```

**A test only ever seen to pass is not known to work** — and I had run that control on every other change in this arc.

### #912 is not cured, and the PR said it was

`gen-check` being pixel-identical is the proof, and I shipped it as a *virtue* without noticing what it also means: the drawn position never moved, so the painter's nudge still seats **41 of 1316** settled mascot frames on furniture at `112x100` (0 of 1316 at `192x160`). Confirmed by my own measurement, not taken on the lens's word.

`Closes #912` is withdrawn. What this PR fixes is the **direction** — the sim's answer is now always real floor, so the hover box, the z-sort row and the next leg's origin stop inheriting the compromise. The residual now carries the `SHARP-EDGES.md` entry that #912's own fix sketch asked for and that I skipped while choosing its option.

### Findings → terminal state

| Finding | Lens | State |
|---|---|---|
| The sim test cannot fail on the reintroduced bug | correctness | **FIXED** — swept at 192x80, controlled both ways |
| The delete also removed a RECOVERY SNAP, described nowhere — it fired on any unwalkable point, quantized mid-walk samples to coarse-cell centres, and its removal changes 8.8% of walk frames (in the good direction: the old code froze and teleported along a leg the sim walked smoothly) | correctness | **FIXED** — recorded in `WHERE-TO-LOOK.md`. "Sprite size never moves a position" is true of the clamp and says nothing about the snap |
| `Closes #912` retires a defect that is bit-for-bit unchanged | design | **FIXED** — withdrawn, + the sharp edge |
| "Moving a DRAWN position harms nothing" is false, repeated in 3 places | design | **FIXED** |
| The new paint test is redundant and its name claims painter coverage it lacks | correctness | **FIXED — deleted.** Verified: dropping the painter's guard leaves mine green while the pre-existing `a_roaming_creature_is_never_sliced_by_the_canvas_edge` reds. That one drives the real enqueue path and covers the pet too |
| `just gen-wasm` skipped — the documented trap nothing catches, and the site hero renders a mascot | correctness | **FIXED** — regenerated, `gen-wasm-check` OK |
| The deleted rationale (why not inset the destination) has no home | design | **FIXED** — in the sharp edge's rejected-alternatives clause |
| The layering claim is mascot-only; `pet_position` still takes the pack | design | **FIXED** — scoped in the guide (it asks whether the anim EXISTS, not its size) |
| Agents get no on-canvas guard (2% / 7%) | both | **ISSUE #916** — including the anchor difference that makes it *not* the same arithmetic: agents are feet-anchored, creatures centre-anchored |

### Refuted — attacked and held

Every consumer reads the **clamped** position: `enqueue_pet` / `enqueue_gateway_mascots` rebind `pos` before constructing `PetFrame`/`MascotFrame` and both `anchor_y` z-keys, and the TUI hover reads those. No consumer sees an unclamped sim point — the regression I most feared does not exist. No render path bypasses the guard (`tui`, `floating`, `pixtuoid-web`, the snapshot example and `corpus_check` all reach the creatures only through `FloorSession` → `render_to_rgb_buffer`). The y clamp is **dead** — 0 of 72,000 samples hit it, because row 0 and the last row hold zero walkable cells. `mascot_position` losing `pack` is behaviour-neutral at the base commit. The `gen-check` posters were rebuilt byte-identical.

### Gates at `5b5acc75`

`just preflight` green, **3033 tests** · `just gen-check` green · `just gen-wasm-check` OK · clippy 0.

IvanWng97:
## Correction to my own disposition comment above

The number I posted an hour ago was wrong, and the lens that flagged it was right. I had quoted **one configuration's** measurement as the general one — the same mistake twice in one PR.

Censused instead: four buffer sizes × two desk caps, settled mascot frames, through the painter's own `keep_sprite_on_canvas`.

```
192x80    off_mask 41/984    not_clear 41/984    <- the real residual
112x100   off_mask 41/984    not_clear  0        <- off the MASK, but visually CLEAR
192x160   clean
152x140   clean
```

So: at `112x100` this PR **cures** the visible symptom rather than leaving it — the 41 frames there are a footprint-only artifact, not a creature sitting on furniture. The size where the drawn sprite genuinely still lands on furniture is **`192x80`**, which is also the size the fixed test now sweeps. The conclusion (`Closes #912` withdrawn, the residual belongs to the painter) is unchanged; the evidence for it moved to the right size.

Fixed in `e3242afd`, and the sharp edge now names the configurations rather than quoting a bare number, because a config-specific number read as general is exactly how the first version went wrong.

### Standing back

Every finding across the three review rounds on this PR has been about a **claim**, not the code — a test that could not fail, a wrong `Closes`, a false "harms nothing", a redundant test whose name overclaimed, a skipped `gen-wasm`, and this number. Two lenses and two bots have found **zero defects in the production change**; the layering holds, and the hover box, z-sort and every render path were independently traced clean.

That is worth stating plainly rather than burying: the code was right the first time, and I have been writing assertions faster than I verify them. No further prose is going into this PR — anything else found goes to an issue.

Issue #900, opened 2026-08-10
perf: the two wash_since passes scan the whole buffer — 15-17% of a frame


IvanWng97:
## What

`paint_frame` applies the hour's cast to the foreground with two `wash_since`
passes, each a **full-buffer scan** comparing the live buffer against a snapshot
(`crates/pixtuoid-scene/src/pixel_painter/mod.rs`). The scans are the dominant
avoidable cost in the paint pass.

## Measured (release, 200 frames per configuration, 12 agents)

| | 192x160 | 360x240 |
|---|---|---|
| whole frame (`render_floor`) | 473 us | 1331 us |
| 2 clones + 2 scans, no writes | 80 us (**17%**) | 200 us (**15%**) |
| 2 clones + 2 scans, every pixel writes | 433 us (**92%**) | 1290 us (**97%**) |
| the two clones ALONE | 3.3 us | 9.8 us |
| the same, reusing a buffer (`clone_from`) | 2.7 us | 8.8 us |

Real frames sit near the no-writes end, so **~15-17% of a frame** is the honest
figure; the all-writes row is the ceiling, not a claim about today.

## What is NOT the problem

A review round proposed hanging two scratch buffers off `FloorCtx` to kill the
per-frame `clone()` allocations. **Measured prize: 0.6-1.0 us, i.e. 0.08-0.12%
of a frame** — the allocation is nearly free, the memcpy dominates, and reusing
a buffer still pays the memcpy. Refuted; recorded here so it is not re-proposed.

## The actual lever

Both scans walk `w x h` to find the pixels that changed since the snapshot,
which are exactly the pixels the enclosed painters touched. Bound the scan to
what was painted instead of the whole buffer:

- the drawable loop already knows each drawable's anchor and sprite size, so
  pass 2 could scan the union of their bounding boxes;
- the floor-fixture group (pass 1) is a small fixed set with known rects.

`wash_since`'s diff-against-a-snapshot approach is itself load-bearing and must
stay — `SHARP-EDGES.md` records that a rectangular BAND was tried and seamed the
window glass at one hour while double-washing the floor between pieces. The
change is to the scan's EXTENT, not to the mechanism.

## Guardrails for whoever picks this up

- `just gen-check` has a 0.0% pixel tolerance, so any regression in the washed
  output reds it. That is the correctness gate for this work.
- Re-measure with the same harness rather than trusting these numbers: a
  temporary `#[ignore]`d test driving `render_floor` N times, timed against the
[...]

PR #917, opened 2026-08-13
feat(hooks): a closing keyword may only be a trailer, never prose

### LINKED ISSUE #912: creatures: place_creature's second clamp can push a snapped point back off the mask

### LINKED ISSUE #916: scene: agents get no on-canvas guard — 2-7% of aimless destinations clip or shift at the buffer edge
## What

`pixel_painter::keep_sprite_on_canvas` is applied to the pet and the gateway mascot at enqueue. **Agents get nothing.** An agent whose position lands within half a sprite of the east/west buffer edge is drawn shifted (`walking_anchor` uses `saturating_sub`, pinning it flush left) or clipped by `blit_frame` on the east side.

## Measured

`pick_aimless_dest`, 4096 seeds per size, against `CHARACTER_SPRITE_W = 8`:

```
192x160    110 / 4096   (2%)
112x100    306 / 4096   (7%)
```

For context, the mask's outermost column is substantially walkable — 98 of 160 rows at 192x160, 113 of 160 on the east side — so agents legally stand there. North and south are already fully blocked (row 0 and the last row hold zero walkable cells), so this is an x-axis-only gap.

## Why it was not fixed alongside #915

#915 establishes the direction — sprite size never moves a sim position; keeping the frame on the canvas is the painter's. Under that rule the agent fix belongs at the same paint layer. It was left out because **agents appear in the committed stills**, so it moves art and needs a `just gen` regen in the same change, whereas #915 is pixel-identical.

## Not the same arithmetic

Worth knowing before writing it: agents are **feet-anchored** (`walking_anchor` subtracts `WALKING_Y_OFF`; `SHARP-EDGES.md` — *"Every character sprite is FEET-anchored"*), creatures are **centre-anchored**. So this is not "call `keep_sprite_on_canvas` for characters too" — the guard's bounds differ by anchor, and `enqueue_characters` works in top-left space where the creature paths work in centre space.

## Fix sketch

- One guard at the paint layer that takes the anchor convention as a parameter, or two named guards that make the convention explicit at each call site.
- `enqueue_characters` (`pixel_painter/mod.rs`) is the centre-anchored-sibling exception with nothing marking it deliberate — the N-1-of-N shape the review prompt calls the most recurrent escape. Whatever lands should make that site non-optional rather than another call each enqueue path must remember.
- Regen `just gen` and commit the moved stills.

IvanWng97:
## Three incidents in one day, the third inside this PR

GitHub reads the **keyword**, not the sentence around it. Two issues were retired today by prose that denied it — a heading asking why a PR does *not* retire 907, and a clause naming the **rejected** alternative that would also have resolved 912. Both were mine, both in this repo's own "don't do this" class, and I had quoted that rule earlier the same day.

The third instance happened while writing this: the first draft of the guard's own commit message quoted both incidents verbatim, and the guard rejected it. That draft would have re-retired both issues on merge.

## Checked against the docs, not my memory of them

I had been asserting the keyword rules from memory all day. [GitHub's docs](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue):

- nine keywords — `close/closes/closed`, `fix/fixes/fixed`, `resolve/resolves/resolved`
- the colon form counts (`Closes: #10`), case does not matter
- **no line-position requirement** — which is precisely why both incidents fired mid-sentence
- **commit messages AND pull-request descriptions** are scanned

## Why a keyword-presence grep would be wrong

`Closes: #N` is the *correct*, daily-used form — #910 retired four issues that way. A guard that fires on every closing keyword is noise people learn to skip, which is the exact failure mode this repo already documented for a length-based comment gate.

The discriminator is **position**, and that is what makes it mechanical instead of a judgment call:

```
Closes: #912                                  trailer, the whole line  → intentional
(the alternative that also fixes #912)        mid-sentence             → accident
```

`Refs #N` remains available for a mention that must not retire anything.

## Wired per this repo's own architecture rule

> "the justfile is the single source of truth for every check — CI and the git hooks call the same recipes (no local-vs-CI drift)"

One implementation, three call sites: `commit-msg` (where the message is still editable), `just lint`, and CI's hygiene job (to catch a bypass). Logic living inside the hook would have been exactly the drift that sentence forbids.

`just lint` is 18 arms now, and the wiring is controlled: breaking the script reds `isslink`.

## Bounded on both sides

The selftest carries both real incidents verbatim — which is *why* they live there and not in the commit message. Twelve cases, both directions:

| must fire | must stay silent |
|---|---|
| the 912 incident, verbatim | `Closes: #912` / `Closes #912` / indented |
| the 907 incident, verbatim | `Refs #912` |
| `we Fixed #916 while we were in there` | `the postfix #912 notation` |
| `this Resolves: #900 only partially` | `the auto-fix #912 pass` |
| | `closes the door on #912` — a keyword NOT touching the number |
| | a `#`-led git comment line |

## Is this best practice? Partly, and here is the honest split

- **The convention** (issue links only as trailers) — standard. Git trailers are documented practice; GitHub's own docs show the colon form; Conventional Commits puts refs in a footer.
- **The enforcement point** (`commit-msg`) — standard. That hook exists for commit-message policy; commitlint/husky occupy the same slot.
- **The rule itself** — **bespoke**. I searched: commitlint has no such rule, you would write a custom one. So this is not reimplementing a wheel, but it is not adopting an established one either.
- **GitHub's repository setting** is the wrong instrument. It is scoped to "auto-close issues with merged linked pull requests", the docs do not say it covers commit keywords, and disabling it would kill the legitimate usage this repo relies on.

## NOT covered, stated rather than implied

A pull-request **description** is authored on GitHub and passes through no hook. **One of today's two incidents came from exactly there.** Covering it needs a workflow that reads `pull_request.body`; that is a separate decision, and this PR does not pretend to have made it.

## Gates

`just lint` 18/18 · `just issue-links` green · selftest 12 cases both directions · the wiring's negative control reds `isslink`.

### DIFF
--- a/.githooks/commit-msg
+++ b/.githooks/commit-msg
@@ -1,0 +1,12 @@
+#!/usr/bin/env bash
+# Commit-msg hook: a GitHub closing keyword may only appear as a trailer.
+# The rule and its WHY live in scripts/issue-links.py; this is the local arm of
+# the same check CI runs over the whole branch, so the two cannot drift.
+#
+# Set up once per clone with:
+#   git config core.hooksPath .githooks
+set -euo pipefail
+
+[[ "${SKIP_PREFLIGHT:-0}" == "1" ]] && exit 0
+cd "$(git rev-parse --show-toplevel)"
+python3 scripts/issue-links.py --stdin <"$1"
--- a/.github/workflows/ci-lint.yml
+++ b/.github/workflows/ci-lint.yml
@@ -133,6 +133,7 @@ jobs:
       - run: just gen-guides-check
       - run: just env-paths
       - run: just gitenv-selftest
+      - run: just issue-links
 
   msrv:
     name: msrv
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -67,6 +67,14 @@ Cross-file report/upload semantics that actionlint cannot express are pinned by
 the yq + Conftest/OPA policy and real action/workflow behavior tests under
 `policy/ci-observability/`; `just ci-observability` runs them inside both
 `just lint` and the CI hygiene job.
+An issue link closes its issue from a merged commit message or a PR description, and
+GitHub reads the KEYWORD, not the sentence — `Why this does NOT close #907` and
+`(the alternative that also fixes #912)` each closed an issue they were denying. So a
+closing keyword belongs on its own trailer line (`Closes: #N`) and nowhere else;
+`just issue-links` enforces that over commit messages, from the `commit-msg` hook where
+it can still be fixed and from `lint`/hygiene to catch a bypass. A PR DESCRIPTION is
+authored on GitHub and passes through no hook — that half stays on the author.
+
 `just zizmor` adds the upstream workflow/action/Dependabot security analyzer:
 the repository deliberately requires a symbolic ref or SHA (not SHA-only),
 every checkout drops persisted credentials, and accepted analyzer findings use
--- a/justfile
+++ b/justfile
@@ -318,6 +318,7 @@ lint:
     run guides  just gen-guides-check     & pids+=($!)
     run prose   just comment-lint-gate    & pids+=($!)
     run gitenv  just gitenv-selftest      & pids+=($!)
+    run isslink just issue-links        & pids+=($!)
     for p in "${pids[@]}"; do wait "$p" || fail=1; done
     [[ $fail -eq 0 ]]
 
@@ -1293,6 +1294,16 @@ comment-lint-replay n="20":
     echo "fn-body arm (advisory): $advisory/$total merged commits, $flagged lines"
     echo "blocking arms (--gate): $gated/$total merged commits — an arm reds the"
     echo "  commits that predate its own calibration, so read this against \`git log -S\`"
+
+# A closing keyword GitHub reads but the sentence does not mean. Wired in three
+# places that must stay in step: the `commit-msg` hook, `just lint`, and CI's
+# hygiene job — a commit message is authored locally, so the hook is where it can
+# still be fixed, and the other two catch a bypass.
+[group('meta')]
+[doc('Check commit messages for inline GitHub closing keywords')]
+issue-links range="origin/main..HEAD":
+    python3 scripts/issue-links.py --selftest
+    python3 scripts/issue-links.py {{ range }}
 
 # The seam's WHY lives in scripts/gitenv.py's docstring. This pins the scrub AND
 # sweeps scripts/ for anything spawning git outside `gitenv.git()` — the recurrence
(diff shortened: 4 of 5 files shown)


IvanWng97:
Closing unmerged — this should not exist.

The premise was wrong. Other repos do not hit this because they do not write commit messages and PR bodies that argue *about* issue numbers mid-sentence. A normal message names an issue once, deliberately, in a trailer. Both of today's accidents came from my own prose — "why this does NOT close [907]" and "the alternative that also fixes [912]" — so the root cause is a writing habit, not a missing guard in the repository.

The proportionate fix is a convention with no code: **when mentioning an issue you do not intend to retire, write the number without the hash** (`issue 912`, or just `912`). GitHub only acts on the hash form. That covers commit messages AND pull-request descriptions — the second of which no commit hook can see, which this branch admitted but could not solve.

Against that, this branch was 130 lines of Python plus a hook, a recipe, a lint arm and a CI step, with a real false-positive surface (it rejected a legitimate quotation of the incidents). Machinery for a habit.

Recorded rather than deleted silently, because the lesson is the one worth keeping: two review lenses were about to examine a thing that should not have been built, and no amount of reviewing finds that.

PR #918, opened 2026-08-13
perf(scene): render frame −45/−57% byte-identical + criterion/CodSpeed bench pipeline

### LINKED ISSUE #900: perf: the two wash_since passes scan the whole buffer — 15-17% of a frame

IvanWng97:
## What

Three byte-identical paint-pass optimizations (profile-first, samply), plus a durable benchmark pipeline. Closes #900.

**Render optimizations** (each pinned by an exactness test vs the naive reference; `just gen-check` 0.0% pixel tolerance green, re-rendered from source):

- **`wash_since`** scans in 64-px chunks with a branch-free XOR-OR reduction — the clean-chunk compare vectorizes (LD3 on aarch64 handles the 3-byte stride), the scalar walk runs only inside differing chunks, and the wash goes through a lazily-tabulated `RgbLut`. The issue's isolated 2-clones+2-scans measure: 66.5→11.8µs @192×160, 180.5→25.5µs @360×240 (was 13.7%/14.5% of a frame — #900's 15-17% reproduced first).
- **`blend_floor_band`** tabulates its per-pass constant tint+strength into a 3×256 `RgbLut` (28% profile self-time → 11%). Byte-equality holds by construction: the LUT is built by evaluating the same `blend` per channel.
- **`BaseFillCache`** memoizes the carpet-noise + wall-band fill on `FloorCtx` (a paint-local store beside `FrameCache`), keyed on every input the fill reads. The weather-tint leg of its pinning test is the stale-key negative control — verified red with carpet dropped from the key.

**Whole frame** (same `#[ignore]`d `perf_measure_900` harness, release, M-series, 200 frames, 12 agents):

| scene | 192×160 | 360×240 |
|---|---|---|
| idle before → after | 483.9 → 263.9µs (−45%) | 1246.3 → 541.9µs (−57%) |
| busy after | 183.7µs | 528.2µs |

**Bench pipeline** (researched against primary sources — criterion FAQ, ratatui's deleted bencher.dev experiment, ruff/uv's non-gating CodSpeed jobs):

- `pixtuoid-scene/benches/render_frame.rs` (criterion via `codspeed-criterion-compat` 5, lean features, passthrough locally) + `just bench`.
- `bench.yml` — on-demand advisory (mutants.yml twin), artifacts carry the criterion report.
- `codspeed.yml` — per-PR instruction-count simulation trend (advisory, OIDC, no secret). **Activation needs the CodSpeed GitHub App installed on the repo.**

## Verification

- `just preflight` green (3036 tests) · `just gen-check` exit 0 (55 comparisons pixel-identical) · `just ci-observability` green (111 policy tests, both new workflows in scope) · `just risk-radar-test` green
- Negative control run: cache key deliberately broken → pinning test red; restored → green.


IvanWng97:
## Two-lens review — disposition sweep @ 6cd4c8e1

Three local lenses (correctness/grounding · design/blast-radius · behavior-preserving-refactor skeptic, per the escalation trigger) + the online bot. Every finding driven to a terminal state:

**Lens 1 (correctness) — APPROVE-WITH-NITS, 7 findings**
- gen-wasm skipped → **FIXED**: `site/public/wasm/` regenerated + committed (the committed wasm still ran the un-optimized painter).
- `wash_since` doc claimed SIMD codegen; emitted assembly is scalar → **FIXED**: comment now states the true mechanism (branchless per-chunk reduction).
- inline `~28%` measurement in a doc comment → **FIXED**: removed (lives in the perf commit message).
- `paint_lightning_flash` unconverted sibling → **FIXED**: converted to the same `RgbLut` + its own exactness pin.
- cache key negative controls covered only carpet/wall → **FIXED**: single-axis width-only and height-only resize legs added.
- `pub(crate)` widening + duplicate `#[ignore]`d harness → **FIXED**: harness deleted (criterion `idle12`/`busy12` cover both scenes), `wash_since` back to private.
- codspeed-app-absent red / bench 60s time-wrap → **RESOLVED** (app installed, job green) / **REFUTED-as-benign**: one bounded backward step per 600 iterations; core's documented `duration_since` folds absorb clock regression.

**Lens 2 (design) — REQUEST-CHANGES, 12 findings**
- F1 bench read the operator's XDG pack → **FIXED**: `#[doc(hidden)] embedded_default_pack()` (mechanism-not-contract escape hatch; api golden unchanged, `just api-surface-check` exit 0).
- F2 three perf harnesses → **FIXED**: criterion gains `idle12` (the slower scene), `perf_measure_900` deleted; `examples/render_bench.rs` is cross-referenced, not retired — it measures buffer-size SCALING through the floating offscreen renderer for the 2.5D design gate, a different instrument.
- F3 no path filter → **FIXED** (raycast/site/pages precedent).
- F4 workflow-scoped `id-token` → **FIXED**: job-scoped per ci-tests.yml's declared-not-inherited rule. Fork sub-point verified against CodSpeed docs: public repos support tokenless uploads from forks; OIDC kept per their recommendation — worst case is a red advisory check on a fork PR, switch to tokenless if that materializes.
- F5 LUT before guard → **FIXED**. F6+F7 panic contract + bare tuple → **FIXED**: `BaseFillKey` named struct + `blit_into` owning the copy with a length `debug_assert` (a transposed u16 pair no longer compiles).
- F8 step name overclaimed → **FIXED** ("criterion measurements"; verified JSON-only under the lean feature set). F9 taiki-e note → **FIXED** (self-expiring form). F10 → **FIXED** (private). F11 "mid-bucket" → **FIXED** (states offset 400).
- F12 `benches/` in the published crate → **REFUTED**: `site_badge_colors` is excluded because it reads `../../site` sources absent from the tarball (would fail a consumer's `cargo test`); the bench is self-contained, consumers never resolve dev-deps, and excluding an explicit `[[bench]]` adds a publish-time warning for zero consumer benefit. Shipping benches in-package is the ecosystem norm (ratatui).
- Comment audit: `wash_since` S2 → **FIXED**, using Lens 1's assembly evidence over the proposed SIMD wording (the lenses conflicted; the measurement wins).

**Lens 3 (equivalence skeptic) — APPROVE-WITH-NITS**: all three refactors byte-identical over ~2,240 differential cases vs the verbatim base-commit implementations, negative-controlled. F1 co-varying resize leg → **FIXED** (single-axis legs). N1 tabulate-before-guard → **FIXED**. N2 unpinned clamp arms → **FIXED** (degenerate-bounds pin). N3 visibility → **FIXED**.

**Gates at 6cd4c8e1 (local)**: preflight 3039 green · scene lib 802 green · gen-check exit 0 (byte-identical, 55 comparisons) · `just api-surface-check` exit 0 · `just ci-observability` 111 policy tests green · `just risk-radar-test` green.

Version adjudication (Lens 2, on record): no public-surface move, contributor-infra only — no bump in this PR; rides the next release.

plan-miss recorded in the review commit: the gen-wasm rule for scene changes was missed until review.

IvanWng97:
Deferral note for the record: the red `audit · format · lint · types · build` check is GHSA-jmr9-qjv8-65gv (extract-zip, no patched release) arriving via the site's lighthouse dev-chain — pre-existing upstream state surfaced because this PR's gen-wasm regen touched `site/public/wasm/`, not caused by this diff. It is not a required check (all required checks green). Owner decision: fixed-and-verified override (`@puppeteer/browsers ^3.2.0`, audit → 0) deliberately NOT shipped here — deferred to the upcoming site-governance pass.

PR #921, opened 2026-08-13
fix(site): clear GHSA-jmr9-qjv8-65gv by overriding @puppeteer/browsers to 3.x

IvanWng97:
## Summary

`npm run audit` has been red on **every** site-path CI run ([example](https://github.com/IvanWng97/pixtuoid/actions/runs/31661563962/job/94327261262)) — 6 high-severity findings, all one advisory: [GHSA-jmr9-qjv8-65gv](https://github.com/advisories/GHSA-jmr9-qjv8-65gv), extract-zip unvalidated symlink path traversal, reached through LHCI. One line of `site/package.json` overrides `@puppeteer/browsers` to `^3.2.0`, which drops the vulnerable subtree entirely.

## Related issue

n/a — no tracking issue. This closes Dependabot alert **#41** (same GHSA, manifest `site/package-lock.json`), which resolves on its own once the package leaves the lockfile.

## Why not just upgrade

That was the first thing checked, and it is **not possible** — nothing above `extract-zip` can float past it:

```
@lhci/cli 0.15.1          ← already latest (published 2025-06-25; `next` tag == `latest`)
  └ pins  lighthouse 12.6.1        ← EXACT, not a range (13.4.x exists but is unreachable)
      └ takes puppeteer-core ^24   ← a caret — but all 77 published 24.x releases pin an
          │                          EXACT @puppeteer/browsers (26 distinct values,
          │                          2.7.0 … 2.13.2), every one a vulnerable 2.x, so the
          │                          caret has nowhere clean to float
          └ pins  @puppeteer/browsers 2.13.2   ← exact
              └ extract-zip <= 2.0.1           ← first_patched_version: null
```

The bottom of the chain has no fix at all — confirmed against the GitHub advisory API rather than npm audit's rendering:

```
$ gh api /advisories/GHSA-jmr9-qjv8-65gv --jq '.vulnerabilities[]|{pkg:.package.name,patched:.first_patched_version}'
{"pkg":"extract-zip","patched":null}
```

2.0.1 is simultaneously the latest published version *and* vulnerable.

Proven empirically in a scratch dir (per the protocol in `site/CLAUDE.md` — never the working tree, whose lockfile is the artifact under test): with the override removed **and** `@lhci/cli` pinned to `latest`, npm still resolves `extract-zip 2.0.1` and `npm audit` exits 1.

`npm audit fix --force` proposes downgrading to `@lhci/cli@0.12.0` (2023) — a breaking change, and backwards.

## Why the override is safe

`@puppeteer/browsers` 3.x is the first major without extract-zip, so the whole subtree leaves: **37 lockfile packages out, 2 in** (−479/+39 lines).

<sub>(Precisely: `3.0.2` dropped extract-zip while still using `tar-fs`; `modern-tar` arrived two patches later in `3.0.4`, replacing `tar-fs`. Those are two separate upstream changes — an earlier draft of this PR fused them into one claim.)</sub>

That is a major-version swap underneath `puppeteer-core` 24's exact pin, so it did not get waved through on "LHCI never calls the download path anyway." Two independent checks:

1. **Full binding sweep.** Every binding puppeteer-core 24.43.1 takes from the package was extracted *mechanically* across both builds — `lib/esm` **and** `lib/cjs` — rather than from a hand-written list: **14** names (`Browser`, `CDP_WEBSOCKET_ENDPOINT_REGEX`, `ChromeReleaseChannel`, `TimeoutError`, `WEBDRIVER_BIDI_WEBSOCKET_ENDPOINT_REGEX`, `computeExecutablePath`, `computeSystemExecutablePath`, `createProfile`, `detectBrowserPlatform`, `getInstalledBrowsers`, `launch`, `resolveBuildId`, `resolveDefaultUserDataDir`, `uninstall`). All 14 exist in the installed 3.2.0.

   Two of those are reachable *only* by methods a static-import grep misses, which is why the sweep had to be mechanical: the CJS build uses `require(...)` + `browsers_1.<name>` member access, and `resolveDefaultUserDataDir` appears solely in a **dynamic** `await import(...)` inside `BrowserConnector` (esm:76 / cjs:113). ESM link-time checking cannot see either, so it is not on its own sufficient evidence.
2. **Both real entry points loaded**, exactly as their consumers do: ESM as `lighthouse` requires it, CJS as `@lhci/cli` does. The static-ESM link half was additionally negative-controlled — importing a deliberately bogus export from the same module *does* throw, so that half is not vacuously green.

3. **The module-system crossing**, which the first draft missed entirely: `2.13.2` was a *dual* package (`exports` carrying both `import` and `require`); `3.2.0` is `type: module` with no exports map at all — ESM-only. puppeteer-core's CJS build `require()`s it from five modules, so this only works via Node's `require(esm)`, unflagged from **22.12** — under this project's `engines.node >= 26` floor (`engine-strict=true`, both workflows on node 26). Both real entry points were loaded to confirm: ESM as `lighthouse` does, CJS as `@lhci/cli` does.

Independently: LHCI reaches Chrome through `chrome-launcher` (already overridden here), so puppeteer's browser-download path is not on the hot path to begin with.

**The shipped site is provably unaffected.** `dist/` built at base and at HEAD — with both non-deterministic inputs pinned identically (`GH_STARS_OVERRIDE`, and the GitHub PR-feed fetch stubbed) — is **byte-identical**: 76/76 files, same aggregate sha256 `de66018f…`. A set-difference over the lockfiles' `packages` keys shows **0 in-place version changes**, so no Astro/Vite/esbuild/TypeScript package moved. `puppeteer-core` has exactly one path in the tree (`@lhci/cli → lighthouse → puppeteer-core`), and `rehype-mermaid` renders via `mermaid-isomorphic`/Playwright, not puppeteer.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] New theme / sprite pack
- [ ] Docs only
- [ ] Refactor / chore

## How I tested it

Run in `site.yml`'s real order, not just the static tier. This matters: the previously-drafted version of this fix was only ever validated with `just site-check`, which **never launches a browser** — and `npm run lighthouse` sits *ahead* of `audit` in the workflow, so it is the step an incompatible puppeteer would actually break.

| stage | result |
|---|---|
| `just preflight` (lint · clippy · hack · test) | **0** — 3039 passed, 2 skipped |
| `just site-check` (format · lint · astro check · knip · unit · build · check:docs · audit) | **0** — `found 0 vulnerabilities` |
| `npm run e2e` | **0** — 96 passed |
| `npm run lighthouse` | **0** — 18 runs across 6 URLs, all assertions pass |
| `npm install-scripts ls` | no unreviewed install scripts |
| `npm audit` **at base** (negative control) | **1** — GHSA-jmr9-qjv8-65gv, 6 high |

The last row is the point: the gate genuinely fails without this override, so `npm run audit` — last step of `verify`/`site-check`, and wired at `site.yml:69` + `pages.yml:58` — is the regression teeth. Observed, not reasoned.

<sub>One e2e spec (the ★ star chip) failed on the first pass because the local build lacked `GH_STARS_OVERRIDE=842`, which `site.yml` sets; rebuilt with it, 96/96 green. Unrelated to this change.</sub>

- [x] Not a visual change, or — screenshot/GIF attached below and it reads at half-block scale.

## Checklist

- [x] I read the relevant `CLAUDE.md` (root + the nested one for the crate I touched) — root + `site/CLAUDE.md`.
- [ ] New behavior has a test (this repo is TDD-first). — **n/a**: no new behavior. The regression gate is the existing `npm run audit` step, which was red before this commit and is green after; the override's compatibility is covered by the link-check + negative control described above.
- [x] No `unwrap()` in non-test code; errors propagate via `anyhow`/`thiserror`. — **n/a**, no Rust changed.
- [x] No new `println!`/`eprintln!` on a production path (use `tracing`). — **n/a**, no Rust changed.
- [x] Docs updated in the same commit if I changed module structure, architecture, or public API (`CLAUDE.md` / `README.md`) — see below.
- [ ] Checked against the [recurring pitfalls](https://github.com/IvanWng97/pixtuoid/blob/main/docs/CONTRIBUTING.md#recurring-pitfalls-this-codebases-review-history-distilled) — **n/a**: that list (char-safe slicing · parallel copies · decode-boundary sanitizing · negative branches) is about Rust decode paths; this diff touches no code.

## Docs

`site/CLAUDE.md` gains a paragraph beside the existing `chrome-launcher` override rationale: why no upgrade clears it, what the cross-major swap rests on, which gate proves it (`npm run lighthouse` — `site-check` never runs LHCI), and **the retirement condition**, routed through the scratch-dir protocol the very next paragraph already documents: drop the entry there and read the top of npm audit's vulnerable `lighthouse` range (`13.3.0` today, so a `lighthouse >= 13.4.0` reaching `@lhci/cli` clears it). That file already warns that *nothing gates a stale `overrides` entry*, so the exit condition has to be written down — and written as a live-checkable property rather than a version number that rots.

## AI assistance

### DIFF
--- a/site/CLAUDE.md
+++ b/site/CLAUDE.md
@@ -233,15 +233,20 @@ old CommonJS launcher line to 0.15.2 (removing its deprecated
 `rimraf → glob → inflight` chain); do not widen it to 1.x, which is ESM-only
 while LHCI 0.15.1 still calls `require('chrome-launcher')`.
 The unqualified `@puppeteer/browsers` → 3.x override is the only route past
-GHSA-jmr9-qjv8-65gv, because no bump anywhere clears it: `extract-zip` has no
-patched release (the advisory range is `*`) and every link above it is an
-upstream EXACT pin — latest `@lhci/cli` 0.15.1 pins `lighthouse 12.6.1`, which
-pins `puppeteer-core ^24`, which pins `@puppeteer/browsers 2.13.2`; 3.x is the
-first to swap extract-zip for `modern-tar`. Crossing that major under
-puppeteer-core 24 holds only because LHCI launches Chrome through
-`chrome-launcher` and never enters puppeteer's download path, so the gate that
-proves it is `npm run lighthouse` — `site-check` stops short of the browser.
-Retire the entry once `@lhci/cli` ships a `lighthouse >= 13.4.1`.
+GHSA-jmr9-qjv8-65gv, because no bump clears it in place: `extract-zip` has no
+patched release (npm audit prints its range as `*`) and nothing above it can
+float past one — `@lhci/cli` 0.15.1 pins `lighthouse 12.6.1` exactly, which
+takes `puppeteer-core ^24`, and every published 24.x pins an exact
+`@puppeteer/browsers` 2.x; 3.x is the first major without extract-zip. Crossing
+that major under puppeteer-core 24 rests on two things: LHCI launches Chrome
+through `chrome-launcher`, so puppeteer's download path never runs, and 3.x
+being ESM-only (2.x shipped dual) is survivable only because puppeteer-core's
+CJS `require` rides Node's `require(esm)` — available from 22.12, under this
+project's node-26 floor. The gate that proves it is `npm run lighthouse` —
+`site-check` never runs LHCI.
+Retire it once nothing requests `@puppeteer/browsers < 3`: drop the entry per
+the protocol below and read the top of npm audit's vulnerable `lighthouse`
+range — 13.3.0 today, so a `lighthouse >= 13.4.0` reaching `@lhci/cli` clears it.
 
 **Nothing gates a STALE `overrides` entry**, so retiring one is a manual
 audit: copy `package.json` to a scratch dir (never the working tree, whose
--- a/site/CLAUDE.md
+++ b/site/CLAUDE.md
@@ -232,6 +232,16 @@ The version-qualified `chrome-launcher@^0.13.4` override upgrades only LHCI's
 old CommonJS launcher line to 0.15.2 (removing its deprecated
 `rimraf → glob → inflight` chain); do not widen it to 1.x, which is ESM-only
 while LHCI 0.15.1 still calls `require('chrome-launcher')`.
+The unqualified `@puppeteer/browsers` → 3.x override is the only route past
+GHSA-jmr9-qjv8-65gv, because no bump anywhere clears it: `extract-zip` has no
+patched release (the advisory range is `*`) and every link above it is an
+upstream EXACT pin — latest `@lhci/cli` 0.15.1 pins `lighthouse 12.6.1`, which
+pins `puppeteer-core ^24`, which pins `@puppeteer/browsers 2.13.2`; 3.x is the
+first to swap extract-zip for `modern-tar`. Crossing that major under
+puppeteer-core 24 holds only because LHCI launches Chrome through
+`chrome-launcher` and never enters puppeteer's download path, so the gate that
+proves it is `npm run lighthouse` — `site-check` stops short of the browser.
+Retire the entry once `@lhci/cli` ships a `lighthouse >= 13.4.1`.
 
 **Nothing gates a STALE `overrides` entry**, so retiring one is a manual
 audit: copy `package.json` to a scratch dir (never the working tree, whose
--- a/site/package-lock.json
+++ b/site/package-lock.json
@@ -3223,174 +3223,6 @@
         "node": ">=20"
       }
     },
-    "node_modules/@puppeteer/browsers": {
-      "version": "2.13.2",
-      "resolved": "https://registry.npmjs.org/@puppeteer/browsers/-/browsers-2.13.2.tgz",
-      "integrity": "sha512-5EUZSUIc37H6aIXyWO0Z4y8NlF8NnjgmqeQgOGiswAU7pY0HOo16ho4+alIWmSfdZnjqBRawMsP3I5YqLSn6kw==",
-      "dev": true,
-      "license": "Apache-2.0",
-      "dependencies": {
-        "debug": "^4.4.3",
-        "extract-zip": "^2.0.1",
-        "progress": "^2.0.3",
-        "proxy-agent": "^6.5.0",
-        "semver": "^7.7.4",
-        "tar-fs": "^3.1.1",
-        "yargs": "^17.7.2"
-      },
-      "bin": {
-        "browsers": "lib/cjs/main-cli.js"
-      },
-      "engines": {
-        "node": ">=18"
-      }
-    },
-    "node_modules/@puppeteer/browsers/node_modules/ansi-regex": {
-      "version": "5.0.1",
-      "resolved": "https://registry.npmjs.org/ansi-regex/-/ansi-regex-5.0.1.tgz",
-      "integrity": "sha512-quJQXlTSUGL2LH9SUXo8VwsY4soanhgo6LNSm84E1LBcE8s3O0wpdiRzyR9z/ZZJMlMWv37qOOb9pdJlMUEKFQ==",
-      "dev": true,
-      "license": "MIT",
-      "engines": {
-        "node": ">=8"
-      }
-    },
-    "node_modules/@puppeteer/browsers/node_modules/ansi-styles": {
-      "version": "4.3.0",
-      "resolved": "https://registry.npmjs.org/ansi-styles/-/ansi-styles-4.3.0.tgz",
-      "integrity": "sha512-zbB9rCJAT1rbjiVDb2hqKFHNYLxgtk8NURxZ3IZwD3F6NtxbXZQCnnSi1Lkx+IDohdPlFp222wVALIheZJQSEg==",
-      "dev": true,
-      "license": "MIT",
-      "dependencies": {
-        "color-convert": "^2.0.1"
-      },
-      "engines": {
-        "node": ">=8"
-      },
-      "funding": {
-        "url": "https://github.com/chalk/ansi-styles?sponsor=1"
-      }
-    },
-    "node_modules/@puppeteer/browsers/node_modules/cliui": {
-      "version": "8.0.1",
-      "resolved": "https://registry.npmjs.org/cliui/-/cliui-8.0.1.tgz",
-      "integrity": "sha512-BSeNnyus75C4//NQ9gQt1/csTXyo/8Sb+afLAkzAptFuMsod9HFokGNudZpi/oQV73hnVK+sR+5PVRMd+Dr7YQ==",
-      "dev": true,
-      "license": "ISC",
-      "dependencies": {
-        "string-width": "^4.2.0",
-        "strip-ansi": "^6.0.1",
-        "wrap-ansi": "^7.0.0"
-      },
-      "engines": {
-        "node": ">=12"
-      }
-    },
-    "node_modules/@puppeteer/browsers/node_modules/color-convert": {
-      "version": "2.0.1",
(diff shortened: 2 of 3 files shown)


IvanWng97:

Both lenses returned **REQUEST-CHANGES**. Deduped to 7 findings; every one below reaches exactly one terminal state. Both lenses cleared the **code** half independently — all fixes are prose in `site/CLAUDE.md`, and `site/package.json` + the lockfile are untouched by the review round (`26dd9599`).

Lens 1 (correctness/grounding) had to be re-run: the first dispatch returned no report at all, and a dispatch is not a review.

| # | Finding | Raised by | State |
|---|---|---|---|
| 1 | "every link above it is an upstream **EXACT** pin" — false universal | L2 · L1 | **FIXED** |
| 2 | Retirement threshold `>= 13.4.1` is not the boundary | L2 · L1 | **FIXED** |
| 3 | "3.x swapped extract-zip for `modern-tar`" fuses two changes | L2 · L1 | **FIXED** |
| 4 | "`site-check` stops short of the browser" is false | L2 | **FIXED** |
| 5 | CJS→ESM crossing omitted next to an ESM-is-disqualifying edge | L1 | **FIXED** |
| 6 | No blank line before the inserted paragraph | L2 | **REFUTED** |
| 7 | Range-qualified override form would self-retire | L2 | **OPEN — owner call** |

### 1 · False "EXACT pin" universal — FIXED

`lighthouse@12.6.1` declares `puppeteer-core: "^24.10.0"` — a caret. The sentence asserted every link was exact and then contradicted itself two clauses later with its own `^24`. Load-bearing rather than cosmetic: that caret is the single link through which the chain could self-heal and leave this override **silently inert** — the exact stale-entry case the very next paragraph says nothing gates, concealed by the sentence written to explain it.

I did **not** adopt L2's proposed replacement, which asserts every published 24.x pins an exact `2.13.x`. That is also false — `24.37.0`–`24.37.3` pin `2.12.0`/`2.12.1`. Swapping one false universal for a narrower one is the same defect. Enumerated all 26 published 24.x instead: every one pins an *exact* version and all are `2.x` (`2.7.0` … `2.13.2`). That enumerated fact is what the doc now claims.

### 2 · Retirement threshold — FIXED

`13.4.0` is the first clean release (`→ puppeteer-core ^25.0.2 → @puppeteer/browsers 3.0.2`); `13.4.1` was merely *latest today*. An over-strict trigger silently keeps a dead override alive — the failure mode the paragraph exists to prevent.

Rewritten as a live-checkable property routed through the scratch-dir protocol the next paragraph already documents, with the number as today's instance rather than the rule. **Negative-controlled**: run in an override-dropped scratch dir, `npm audit` prints the boundary as the top of its vulnerable range — `13.3.0` — so the written instruction is executable, not merely plausible.

### 3 · `modern-tar` mechanism — FIXED

`3.0.2` dropped extract-zip while still using `tar-fs`; `modern-tar` arrived in `3.0.4`, replacing `tar-fs`. Two separate upstream changes. Now: "3.x is the first major without extract-zip."

### 4 · "`site-check` stops short of the browser" — FIXED

False. `astro check` renders the Mermaid diagram through Chromium — `site/CLAUDE.md:44-46` says so explicitly, and **#776** was an outage caused by installing Chromium after `check`. The clause invited precisely that inference. Now: "`site-check` never runs LHCI", which is the true and intended claim.

### 5 · CJS→ESM crossing — FIXED

Neither I nor L2 caught this. `2.13.2` was a **dual** package (`exports` carrying both `import` and `require`); `3.2.0` is `type: module` with **no** exports map — ESM-only. puppeteer-core's CJS build `require()`s it from five modules, so this rides Node's `require(esm)`, unflagged from **22.12**, under this project's `engines.node >= 26`.

It matters because it sits directly below a sharp edge that treats ESM-only as *disqualifying* (`do not widen chrome-launcher to 1.x, which is ESM-only`). Two adjacent overrides sharing that property with opposite verdicts and no stated distinction invites the wrong read of both. The mechanism is now stated.

Deliberately **not** asserted: whether `chrome-launcher` 1.x would also survive under node 26. L1 flagged that as unverified and outside this diff, and I agree — asserting a distinction I haven't tested would repeat finding #1's mistake in a new place. Flagged below.

### 6 · Missing blank line — REFUTED

The adjacent `chrome-launcher` paragraph (`:231`) runs on the same way. Adding a blank line would make this insert inconsistent with its own neighbour rather than fix a defect. L2 itself noted the style is pre-existing and not introduced here.

### 7 · Range-qualified override form — OPEN, owner call

L2 tested `"@puppeteer/browsers@<=2.13.2": "^3.2.0"` and got an identical tree and `0 vulnerabilities`; its virtue is that it **self-retires** — once `puppeteer-core` declares `>= 3` the key stops matching and the entry goes inert on its own, which is the one property `site/CLAUDE.md:246` says nothing gates. L2 recommended it but would not block.

L1 argues the opposite and is also grounded: `@puppeteer/browsers` has exactly one dependent, so qualification is noise today; `npm ci` is lockfile-driven and `.github/dependabot.yml` manages `/site` *dependencies*, not `overrides`, so `^3.2.0` cannot float unattended.

Genuine trade-off, not a defect either way, so I'm not deciding it unilaterally on a verified-green artifact. Shipping the unqualified form as reviewed. Say the word and I'll switch it — it's a one-token change plus a re-verify.

### Not actioned, with reasoning

**An automated staleness check** — both lenses independently concluded it is *not* warranted: it would be inconsistent with the neighbouring `chrome-launcher` entry (which gets no such gate), and the primary defense (`npm audit` exit 1) already covers the case that matters. Recorded here rather than dropped.

### Verification at the review head

`just lint` **0** (12/12) · `npm run audit` **0** · and from L1's independent run at the pre-fix head, unchanged by a prose-only commit: `just site-check` **0**, `npm run lighthouse` **0**, `just site-e2e` **0** (96 passed), `just ci-observability` **0**, and the negative control `npm audit` **at base → exit 1**, which is what proves the gate has teeth.

`dist/` is **byte-identical** base↔HEAD (76/76 files, same aggregate sha256, **0** in-place version changes in the lockfile) — the shipped site is provably unaffected.

IvanWng97:
### Addendum — lens 1's re-verification, and two body-only findings

Lens 1 re-reviewed the rewritten paragraph and **flipped to APPROVE**, with no remaining findings, after independently re-deriving every new claim in it (it pulled the full puppeteer-core packument: **77** published 24.x releases, **0** non-exact pins, **0** non-2.x — so "every published 24.x pins an exact \`@puppeteer/browsers\` 2.x" holds across the whole population, not a spot-check). It also verified the *new* "2.x shipped dual" claim I added, which no finding had asked for.

Two further findings, both against the **PR body only** — the committed doc is clean on both. Recording them here so they reach a terminal state rather than being dropped:

**8 · body overstated what the link-check proved — FIXED.** The body described the export check as a static-ESM link over five importing modules. That method cannot cover two of the bindings: the CJS build reaches them by `require(...)` + `browsers_1.<name>` member access, and `resolveDefaultUserDataDir` appears *only* in a dynamic `await import(...)` inside `BrowserConnector` (esm:76 / cjs:113) — ESM link-time failure is blind to both. The conclusion was never at risk (the final sweep was mechanical across esm ∪ cjs ∪ dynamic, all 14 names present, and both real entry points were loaded), but the body claimed stronger evidence than the method delivered. Body now describes the method actually used, and says plainly that ESM link-checking alone is not sufficient evidence.

**9 · body's "26 of them" was the wrong quantity — FIXED.** 26 is the count of *distinct pin values*; there are **77** published 24.x releases. Verified both numbers myself before rewriting. The committed doc makes no count claim at all, which is why it needed no change — it asserts the property, not a tally.

Both are the same class as findings 1–2: a number or a method description asserted at a confidence its evidence didn't reach. Worth noting that the doc survived both *because* it states a property rather than a count — the version that rots least is the one that doesn't carry a tally.

Head at review: `26dd9599`.

PR #919, opened 2026-08-13
docs(guides): the guide diet — fact + WHY + pointer, with a size gate

IvanWng97:
## What

The guide diet: the `SHARP-EDGES.md` / `WHERE-TO-LOOK.md` siblings go from **317,236 → ~66,000 chars**, `LAYOUT.md` is deleted outright, and a recurrence gate keeps entries from regrowing.

The files had become per-subsystem changelogs — every PR appended its adjudication to the same bullet (the JSONL-watcher entry layered 11 issue numbers at 17.5k chars; the multi-source WHERE-TO-LOOK entry hit 32.7k). Each entry is now the protective minimum: **the surprising fact + the WHY + the authority pointer** (pinning test / in-code comment / issue#). Measurements and adjudication transcripts live only in the cited issues/PRs.

- **SHARP-EDGES**: mega-entries split into component edges (JSONL watcher → 5, hooks-proof-of-life → 3, abrupt-exit → 4). Three real edges buried in WHERE-TO-LOOK prose promoted to entries (grok leader-mode #826, the cross-fire guard, the b1 background-spawn trap).
- **WHERE-TO-LOOK**: an entry survives only when landing on the obvious symbol does NOT answer the question — it names a seam, a split, or a chokepoint. Feature tours the code proves itself are deleted outright.
- **LAYOUT.md deleted** (Closes #920): the annotated tree was the one hand-maintained mirror with NO drift gate against reality, and every sampled annotation is a projection of the file's own `//!` header. The post-fence notes (burn/token/focus-jump plumbing, the `render_floor`/`FloorSession` seam) were the real value — they are seam entries by the survival rule and moved into WHERE-TO-LOOK, under the gate. gen-guides drops the whole layout builder in exchange.
- **Recurrence gate**: `gen-guides` FAILS generation when a sibling entry exceeds `MAX_ENTRY_CHARS = 2000`, with selftest negative controls in BOTH directions (oversized fires on both builders; just-under stays green). Post-diet the largest entry is ~1,700 chars vs a pre-diet population where 40+ entries exceeded 2,000 — a real separation gap (unlike the comment-lint threshold #907 priced and rejected).
- **Root `CLAUDE.md`** encodes both write rules, closing the accretion channel at the source.
[...]

PR #922, opened 2026-08-13
fix(scene): the canvas guard learns both anchor conventions, and characters get one

### LINKED ISSUE #916: scene: agents get no on-canvas guard — 2-7% of aimless destinations clip or shift at the buffer edge

IvanWng97:
`keep_sprite_on_canvas` guarded the pet and the gateway mascot at enqueue; agents got nothing, so one standing in the outermost walkable column lost whatever ran past the buffer to `blit_frame`'s silent clip.

> **Corrections after review** — three claims in this body's first version were false and are fixed below: the census's scope, the guard doc's justification, and which satellites the post-pass covers. Details in the disposition comment.

## East is the only rim that slices

#916 describes two failure modes: an east-side clip **and** a west-side shift from `walking_anchor`'s `saturating_sub`. Both were measured, over 24 sweep sizes × 22 seeds = 528 layouts, across four anchor sources — aimless destinations, waypoint stand points (× every desk-as-origin × step-aside ranks 0/1/2), desk seats, and **every walkable cell** through `walking_anchor`:

| arm | n | west floor | north | east overrun | south |
|---|---:|---:|---:|---:|---:|
| aimless | 147,264 | 0 | 0 | 5,251 (max **3 px**) | 0 |
| waypoint | 275,883 | 4,548 | 0 | 1,364 (max **3 px**) | 0 |
| seat | 6,136 | 0 | 0 | 0 | 0 |
| walk legs | 4,052,006 | 144,690 | 0 | 114,102 (max **3 px**) | 0 |

The west floor **does** engage — all 4,548 waypoint cases come from rank 2's `-STEP_ASIDE_DX`. But it changes nothing, and the reason is exact rather than statistical: **`(x - w/2).max(0)` in `i32` is bit-identical to `x.saturating_sub(w/2)` in `u16` for every `x >= 0`.** The lower bound is already applied; it is just spelled by the type. No sprite is ever sliced on the west or north edge.

South is not merely unobserved but structurally impossible: `WALKING_Y_OFF == 12` equals the tallest character frame, so `anchor.y + h == pos.y <= buf_h - 1`.

## Why not signed coordinates

The textbook answer is a signed destination + a clipping blit: [`image::overlay`](https://docs.rs/image/latest/image/imageops/fn.overlay.html) takes `x: i64, y: i64` and shifts the *source* origin by `-x`; [`embedded-graphics::Point`](https://docs.rs/embedded-graphics/latest/embedded_graphics/geometry/struct.Point.html) is `i32`; [SDL](https://wiki.libsdl.org/SDL2/SDL_BlitSurface) uses `int`.

Each of them addresses a coordinate space **larger than the target buffer** — [multi-monitor virtual screens have genuinely negative origins](https://learn.microsoft.com/en-us/windows/win32/gdi/the-virtual-screen), and a library must blit anywhere. Ours is not that: layout units *are* buffer pixels (`blit_centered` says so), the shape [`ratatui::Rect`](https://docs.rs/ratatui/latest/ratatui/layout/struct.Rect.html) answers with `u16` + clamp — our nearest neighbour, and what the TUI painter already sits on.

Given the bit-identity above, signed anchors would emit the same bytes at all 149,238 west/north sites. They would only add a mechanism for the half-drawn sprite #912 already declined. Retired, not deferred.

## The change

One guard, parameterized on the pre-existing `layout::Anchor` (`Center` / `TopLeft`), living with the anchor helpers. Applied over `resolve_characters`' finished placements — all four pose arms, plus the dust / sleep-z / **waiting** bubble satellites that ride `p.anchor` — and in `character_anchor`, so the badge and the tui hit box track the sprite.

Two things the post-pass deliberately does **not** move, recorded at the call site: the z-key (unreachable — the y bound would need a `>= 17 px` frame, outside the documented cell-aspect envelope) and the **chitchat** visitor anchor (4 in 49,794 pairs, on a bubble already breath-decoupled).

Taking `Anchor` is not a *barrier* against picking the wrong convention — passing `Center` for a top-left `pos` reaches for exactly those bounds. It is one authority instead of two, no more.

## Verification

- **Negative controls, both halves.** Deleting the `resolve_characters` guard reds `a_wandering_character_is_never_sliced_by_the_canvas_edge` (`running 3 px past the 112 px canvas`). Deleting the `character_anchor` clamp reds `a_character_badge_is_never_anchored_off_the_canvas` **while the sprite twin stays green** — which is why the label half needs its own pin. Both tests assert they actually reached the rim, so neither can pass vacuously.
- **preflight** exit 0, 3042 tests (+2). **`gen-check`** pixel-identical.
- **`just gen-wasm`** committed — required of any scene change, and nothing catches a skip (`gen-wasm-check` only verifies the committed set against its own manifest; it returned 0 on the stale bytes).
- **Downstream**: `cutaway_anchor` re-uses `c.anchor` unprojected against `scale.to_buffer(layout.buf_w)`, so the clamp lands flush there too; the web overlay and the tui hit-test ride `character_anchor`. No transposition in the two `(w, h)` → `Size` conversions — checked against the art (`lobster_*` 14x12, `cat_walk_0` 8x6).

Closes #916


IvanWng97:
## Two-lens review — disposition

Lenses run: **correctness**, **design / blast-radius**, **wrong-abstraction** (escalation: a consolidation shipped + a "behavior-preserving" claim), plus the online bot. A comment lens was dispatched, returned a stub, and was re-run — its report is still outstanding and is the one gap in this sweep; I'll fold it before merge rather than close over it.

Every finding below has exactly one terminal state. Premises were verified locally before any fix.

### FIXED

| # | Finding | Source | Fix |
|---|---|---|---|
| 1 | The new sharp edge cited `a_wandering_character_is_never_sliced_by_the_canvas_edge` as the badge clamp's authority, but that test only drives `sim_step`/`resolve_characters` and never calls `character_anchor` — the label half shipped untested | bot (HIGH), lens-1 F1 (92) | Confirmed by deleting the clamp: nothing red. Added `a_character_badge_is_never_anchored_off_the_canvas`, driven through the real routed store. Negative control now reds it with `running 3 px past the 112 px canvas` **while the sprite twin stays green** — which is exactly why it needs its own pin. `936447d4` |
| 2 | The guard's doc claimed parameterizing on `Anchor` means "no call site **can** reach for the other convention's bounds" | lens-2 #1 (85), lens-3 F3 (70) | False — passing `Anchor::Center` for a top-left `pos` reaches for precisely those bounds; the enum is the affordance, not a barrier. It also re-described the signature one line below. Paragraph cut; the PR body repeated it as the answer to #916's N-1-of-N ask and is corrected too. `936447d4` |
| 3 | The sharp edge quantified badge-vs-sprite divergence as "the same ±1px" | lens-2 #2 (93), lens-3 F2 (92), lens-1 F2 (90) | Wrong, and in a don't-fix-this note. Both anchors go through `waypoint_anchor` (one on the real width, one on the constant 8) inheriting `(W−8)/2`; at the rim one clamps to `buf_w−W` and the other to `buf_w−8`, so the clamp **doubles** it to `W−8`. Scoping was wrong too: the bundled pack's seated frames are 8x10 / 8x9 against the label path's fixed 12, so the y axis diverges with **no `--pack-dir` at all**. Rewritten without the number, per #919's rule that measurements live in the PR. `936447d4` |
| 4 | `pub(crate)` wider than needed — every call site is inside `pixel_painter`, and it was the lone `pub(crate)` among that file's `pub(super)` siblings | lens-2 #4 (96), lens-3 F4 (85), lens-1 F4 (85) | `pub(super)`. No lint sees this: `unreachable_pub` fires on bare `pub` only. `936447d4` |
| 5 | `CHARACTER_SPRITE_H_CELLS * 2` had reached four copies of an unnamed authority, and that const's own doc warns the pixel offsets are a separate concern from the cell count | lens-2 #6 (80) | `CHARACTER_SPRITE_H` names it once, beside `CHARACTER_SPRITE_W`. `936447d4` |
| 6 | `character_anchor` is `pub` cross-crate and its contract moved with an unchanged signature — semver-checks and api-surface are both blind | lens-2 #8 (68) | The clamp is now in its `///`, not only an inline `//`. `936447d4` |
| 7 | Mascot's `(14, 12)` fallback stayed an anonymous literal while the pet's twin 30 lines up is a named const with a WHY | lens-2 #9 (65) | `MASCOT_FALLBACK`. `936447d4` |
| 8 | Reworded #912 entry's headline still said "creature" after its body widened to sprites | lens-2 #7 (72) | Headline scoped to "sprite". `936447d4` |
| 9 | `site/public/wasm/` not regenerated — CI is structurally blind (`gen-wasm-check` only verifies the committed set against its OWN manifest; it returned 0 on the stale bytes) | lens-1 F3 (72) | `just gen-wasm` committed. Without it the site's live hero keeps running pre-#916 code — the sliced sprite stays visible on the most-seen surface. Precedent: b14e56bb shipped the same two files. `43d5bb49` |
| 10 | The census in the first commit message and PR body claimed west/north/south never fire | lens-1, item 1b | **My scope did not cover my claim.** It swept aimless dests, waypoint stand points and seats — but not the walk legs, and not the step-aside ranks. Over a strictly larger sweep the west floor engages 144,690 times (walk legs) and 4,548 (waypoint rank 2's `-STEP_ASIDE_DX`). Corrected in the PR body and `43d5bb49`. The conclusion survives with a better argument than the one I gave: `(x - w/2).max(0)` in i32 is **bit-identical** to `x.saturating_sub(w/2)` in u16 for every `x >= 0`, so signed coordinates emit the same bytes at all 149,238 sites. |

### REFUTED — with trace

| # | Finding | Source | Trace |
|---|---|---|---|
| 11 | The post-pass leaves `p.anchor_y` (the z-key) on pre-clamp geometry while the creature sites re-derive their z row from the clamped point | lens-2 #5 (78) | Unreachable, measured: max bottom overrun across the sweep is −4 px, and every bundled character frame is ≤ 12 px. Firing it needs a `--pack-dir` frame ≥ 17 px, outside the terminal-cell-aspect envelope `pixtuoid/SHARP-EDGES.md` documents. Even then the desync sits in the same class as the ±1px breath decoupling `sim.rs` already accepts by design. Left alone; the asymmetry is now recorded on the post-pass comment. |
| 12 | `chitchat::Visitor.anchor` is captured pre-clamp, so a rim-clamped sitter's speech bubble stays at the unclamped x | lens-2 #5, lens-3 F1 (80), lens-1 F5 (60) | Real but marginal: 4 occurrences in 49,794 desk × waypoint pairs, narrowed further by `supports_chitchat`, on a bubble whose anchor is *already* deliberately decoupled from the drawn one (breath-free). Recorded in the post-pass comment rather than fixed. The PR body's claim that the post-pass covers "the bubble satellites" was an overclaim — the **waiting** bubble rides `p.anchor`, the chitchat one does not — and is corrected. |
| 13 | The consolidation might be the wrong abstraction (revert case) | lens-3 Trigger A | Argued per call site adversarially; does not hold. The `Anchor` param carries GEOMETRY, not policy — the policy differences (creatures accept landing on furniture, characters carry a badge) live outside the guard. |
| 14 | The `Center` arm's "byte-for-byte" claim | lens-3 Trigger B | Holds. Both creature sites' `(w, h)` → `Size` packing checked against the art itself (`lobster_walk_0`/`lobster_rest` 14x12, `cat_walk_0` 8x6) — no transposition. Sizing at the new site resolves the frame through the *identical* `frame_at` expression the blit uses, including its out-of-range fallback, so the clamp measures the frame that will actually be painted. |
| 15 | Double frame resolution per character per frame (one extra `&str`-keyed lookup, landing right after b14e56bb's render-perf work) | lens-3 observation | Accepted, not a change request: one lookup per agent per frame, and carrying a resolved `Size` on `CharacterPlacement` buys little. Noted so it isn't a surprise on the next bench delta. CodSpeed on this PR reports no performance change. |

### Process note

The four lenses ran in the **same** worktree I was committing from. Two wrote to it — temp tests, then a real `// MUTATION-B` edit that turned the badge clamp into `let _ = keep_sprite_on_canvas(...); Some(anchor)`. Caught by `git diff` and reverted before any commit, but a mutation that silently disarms a guard is precisely this PR's subject, and it came within one `git add` of shipping. Review lenses belong on a read-only copy; that's on me, not them.

IvanWng97:
## Disposition — comment lens (the outstanding one)

The comment lens I flagged as missing in the sweep above has now reported, against `43d5bb49`. It found ten items, and the interesting part is that **three of them were self-inflicted by the fix round itself** — the change created a real badge-vs-sprite drift, and the round rewrote the NEW comments without sweeping the OLD ones that asserted such drift was impossible. One `rg character_anchor` finds all three. That is the fix-the-class habit failing on my own edit.

Closed out in `da1fb3a2` and `65f095dc`.

### FIXED

| # | Finding | Fix |
|---|---|---|
| F1 | `anchors.rs` "label-vs-sprite drift is **structurally impossible**" — false at the rim, in the very function the diff changed | Scoped to UPSTREAM of the canvas clamps, which is what it actually pins |
| F2a | `sim.rs` "sprite and badge **can't drift**" — same class; the census records `waypoint east=1/2541`, so that arm does reach the rim | Same scoping |
| F3 | The sharp edge's "present on the BUNDLED pack's y axis too" | The premise is true (`seated.sprite` is 8x10 vs the label path's fixed 12) but the assertion of PRESENCE had nothing behind it, contradicted the same commit's census, and the new pin only asserts x. Clause deleted — an unmeasured universal in a don't-fix-this note is exactly the class the bot caught on this same entry last round |
| F4 | The declaration doc and the inline said the same two things fifteen lines apart | The previous round was asked to move rationale to the declaration and instead ADDED the doc while rewriting the inline — the review-fold bloat vector, verbatim. Inline deleted, its unique half folded up |
| F5 | "up to 3" in a test doc | A before-count belongs in the commit message (7eaae042 has it); the 3 is a function of the 112-column fixture |
| F6 | `CHARACTER_SPRITE_H`'s doc said "for the label path", excluding by wording its other reader — `resolve_characters`' missing-animation fallback | Reworded to cover both |
| F7 | The badge test's doc enumerated failure modes the assert cannot catch (a too-TIGHT clamp passes; `Size.h` is unasserted) | Narrowed to what it pins: "a clamp too loose to bind" |
| F8/F9/F10 | A clause restating the loop bound; a Center-only rationale labelling the `TopLeft` assert too (a lone `.min()` cannot panic); `anchors.rs`'s "±1px" vs the entry's "half-difference" — two numbers for one quantity in the same tree | All three corrected; the entry's characterization is the right one, since away from the rim both anchors are `dest.x - w/2` |

`just comment-lint`: **4 items → 0 new runs vs origin/main.** Two fn-body runs were trimmed to the house ≤2-line rule, one of which this review round had itself introduced.

### REFUTED — with trace

| # | Finding | Trace |
|---|---|---|
| F2b | `seat.rs`'s "the badge can **never** float above the sitter" is the same falsified-universal class, and sharper: `seated.sprite` is 8x10 against the badge's fixed 12, so a seat landing in that 2 px window would clamp the badge HIGHER than the sitter | Survives, because it is a Y claim and the y bound is unreachable: max seat bottom overrun across the sweep is `anchor.y + 12 - buf_h = -7`, so the badge clamp has 7 px of slack it never spends. Left alone deliberately. The lens correctly noted F2b and F3 cannot both stand — this is the resolution, and it is why F3's clause went rather than this comment |

Two lenses returned REQUEST-CHANGES on comments/docs only; no code change was required by either. Every finding from all four lenses plus the bot now has a terminal state.

PR #925, opened 2026-08-13
fix(site): clear GHSA-2v37-7h3g-55p8 with an in-range nanoid bump

IvanWng97:
Unblocks the site job that went red mid-afternoon on every branch: GHSA-2v37-7h3g-55p8 (nanoid custom generators loop on size 0) was published 2026-07-29 covering 4.x/5.x only; its 2026-08-13T15:43Z update added the `<3.3.18` range, so the unchanged lockfile started failing `npm audit` (chain: `eslint-plugin-astro → postcss → nanoid@3.3.17`, dev-only).

postcss wants `^3.3.x`, so this is a plain in-range lock refresh to 3.3.18 — no new `overrides` entry to carry (the ladder per site/CLAUDE.md stays: in-range bump → upstream bump → override as last resort). Diff is the 3-line nanoid lock entry; `npm audit` 0 vulnerabilities; `just site-check` exit 0.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01AQULEd97BBLcYgYwGM3VDi

PR #926, opened 2026-08-13
refactor(site): in-repo lighthouse-13 runner replaces the dormant @lhci/cli

IvanWng97:
Stacked on #925 — includes its nanoid lock bump; rebases to nothing once #925 merges.

## Why

`@lhci/cli`'s last publish (2025-06) left it pinning `lighthouse 12`, which dragged two load-bearing `overrides` (`@puppeteer/browsers` → 3.x for GHSA-jmr9-qjv8-65gv, and `chrome-launcher@^0.13.4` → 0.15.2). Lighthouse itself is actively maintained (13.4.1 shipped this week) — the dead weight was the wrapper.

## What

`scripts/lighthouse-runner.mjs` (~170 lines) drives lighthouse 13 through the **official programmatic API** (chrome-launcher + `lighthouse(url, {port})`, verbatim pattern from lighthouse `docs/readme.md`), with aggregation per lighthouse `docs/variability.md`: serial runs, assertions on aggregates (median-of-N). It reads the SAME `lighthouserc.json` — budgets unmigrated, `config/lighthouse-budget.test.mjs` untouched and still gating the file's shape. `npm run lighthouse` keeps its name and the `.lighthouseci/` report dir, so `site.yml` needs zero changes.

**One deliberate semantic divergence**: an assertion without `aggregationMethod` aggregates by MEDIAN (LHCI defaulted to optimistic). Stricter-or-equal; pinned by `config/lighthouse-runner.test.mjs`, which also pins that a renamed audit/category/`user-timings` mark FAILS the run rather than passing vacuously (gate-teeth, both directions).

Retired with the wrapper: both overrides above (lighthouse 13 takes `puppeteer-core ≥25`, natively on `@puppeteer/browsers` 3.x — the extract-zip chain is gone by construction). The lock now carries **zero** `extract-zip` and **zero** `@lhci/*` entries; net −1645 lines.

## Verification

- `npm run verify` exit 0 (format · lint · check · knip · test:unit · build · check:docs · audit).
- Real run: 6 urls × 3 runs, `lighthouse: 12 assertions passed on 6 urls`, 18 reports written.
- Negative control: an impossible FCP budget exited 1 with `FAIL … median 1878.774 violates maxNumericValue 1` — the replacement gate demonstrably fails.
- Unit suite: 63 pass (runner evaluator fixtures incl. median default, pessimistic direction per kind, user-timings mark-vs-measure, missing-value hard-fail).
- Docs current: site/CLAUDE.md override paragraphs rewritten; eslint header comment updated.

https://claude.ai/code/session_01AQULEd97BBLcYgYwGM3VDi

### DIFF
--- a/site/CLAUDE.md
+++ b/site/CLAUDE.md
@@ -193,7 +193,7 @@ failure in the footer, green. `lighthouserc.json` therefore asserts
 contract, so anything that must NEVER regress needs its own per-audit
 assertion, and a CONTRACT audit needs every run to clear it. The
 `aggregationMethod` is not decoration: the runner defaults to `median`
-(`scripts/lighthouse-runner.mjs`), and the median of three runs still greens a
+(`config/lighthouse-runner.mjs`), and the median of three runs still greens a
 binary 0/1 axe audit that failed ONE of them — hence `pessimistic` on the
 contract audits. Second surprise: Lighthouse does **not** pick a theme of its own.
 `Base.astro`'s init falls back to night off the 7/19 wall clock, so an
@@ -228,7 +228,7 @@ exact-version esbuild approval; `fsevents` is explicitly denied because npm's
 registry metadata flags an install script even though the installed manifest
 needs none. Use `npm install-scripts ls` after dependency changes: an
 unreviewed script must fail installation instead of becoming a warning.
-`npm run lighthouse` is an in-repo runner (`scripts/lighthouse-runner.mjs`)
+`npm run lighthouse` is an in-repo runner (`config/lighthouse-runner.mjs`)
 driving lighthouse 13 through the official programmatic API (chrome-launcher +
 `lighthouse(url, {port})`, per lighthouse `docs/readme.md`), reading the SAME
 `lighthouserc.json` budgets `@lhci/cli` did. LHCI was dropped after a year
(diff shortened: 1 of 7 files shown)


IvanWng97:
## Two-lens review — disposition sweep (local gate)

Lenses at `57a3df3f` (runner-semantics correctness · dep-hygiene/docs/CI-wiring), both APPROVE-WITH-NITS. Every finding driven to a terminal state; fixes folded into `8e65674f`.

**FIXED**
- L1-F1 — the default-→optimistic mutation survived the suite (empirically run): new fixture where one GOOD run of three must not rescue the assertion (median 0.5 fails, optimistic would pass). Mutant re-run at the fold head: dies.
- L1-F2 — the duration/startTime fallback-swap mutation survived: the measure fixture now fails via the DURATION read. Mutant re-run: dies.
- L1-F3 / L2-note — `chromeLauncher.launch` moved inside the try: a ChromeNotInstalledError tears the detached preview down instead of squatting port 4321 (the site-e2e sharp edge's port).
- L2-1 — the now-inert `tmp` override retired: zero `tmp` entries in the lock; its only pre-PR dependents (`@lhci/cli`, its inquirer chain's `external-editor`) left with the wrapper. Meets this PR's own retirement criterion.
- L2-2 — `site/CLAUDE.md`'s retire-protocol paragraph past-tensed its chrome-launcher example (this PR retired the pin it cited as live).
- L2-3 + L2-4 — resolved together by adopting the repo precedent: the runner moves to `config/` beside its test with an explicit `node:process` import (the established shape of every node-side site tool), which deletes the `scripts/` eslint carve-out entirely — including its browser-globals leak the lens probed.

**REFUTED / NO-ACTION (with trace)**
- L2 early-crash artifact note: a pre-collect crash leaves `.lighthouseci/` empty and the upload step errors on top of the real failure — the lens's own trace shows this is not a regression (an early LHCI crash produced the identical outcome, with no partial reports at all; the runner improves on it by writing per-run files).
- L2 uuid-override observation: still LIVE (mermaid depends on uuid in-range) — not this PR's defect; flagged for the next override audit per the documented scratch-dir protocol.

Re-verified at `8e65674f`: `npm run verify` exit 0, 63/63 unit tests, both mutants confirmed dying.

PR #927, opened 2026-08-14
perf(bench): CodSpeed gets occupancy, the wire path, memory mode — and a badge

IvanWng97:
## Summary

Turns the CodSpeed integration from one curve into actual coverage: adds the
README badge, two benchmark axes it couldn't see (occupancy, and the wire path),
and the memory instrument alongside simulation.

The starting question was "can we bench a bigger, real-monitor buffer?" —
measuring first killed that idea and pointed at better ones.

**Size is the saturated axis.** #900 measured 2.814x the time for 2.812x the
pixels; this branch reproduces it (192x160 -> 360x240 at 12 agents: 192.7 ->
547.4 us, 2.84x). And there is nowhere bigger to go: `office_scale` pins the
floating window's office buffer near `OFFICE_TARGET_H = 180`, so a *bigger*
display renders a *smaller* buffer.

| window (physical px) | scale | office buffer |
|---|---|---|
| 360x240 default | 1 | **360x240** |
| 1080p full | 6 | 320x180 |
| 5K 5120x2880 | 16 | 320x180 |

360x240 is `FLOATING_DEFAULT_W/H`, the scale-1 case, and the largest buffer any
painter ever produces. A third larger size would be a redundant point on a
straight line.

**Occupancy is what actually varies.** At a fixed 360x240, 12 -> 32 -> 64 agents
costs 547.4 -> 584.1 -> 719.4 us — **+31% at constant pixel count**, all of it
per-agent work (z-sort, routing, chitchat) the pixel curve cannot express.

**The wire had no coverage at all.** `cargo codspeed build` only ever built
`pixtuoid-scene`, so the decode->reduce half — per EVENT, not per FRAME — was
untracked. New `pixtuoid-core/benches/decode_reduce.rs` drives `harness::Drive`
(the one pipeline, so it cannot measure a decoder production doesn't use) over
both transports: 200 turns of JSONL 566 us, hooks 492 us.

**Memory instrument.** Simulation was already the CPU metric (instructions +
cache/memory access patterns), so heap allocation was the unmeasured axis — and
it is the one this repo argues about by hand. #900 refuted the "scratch buffers
on `FloorCtx`" proposal with a throwaway `#[ignore]`d test that no longer
exists, while the render path still clones the full buffer twice per frame.

## Related issue

n/a — follow-on tooling for the benchmarks #900 measured by hand.

## Type of change

- [x] Refactor / chore

## How I tested it

- `just preflight` — exit 0, 3060 tests passed (run twice; again by the pre-push hook).
- `just check-windows` — exit 0. Not in preflight, and this diff adds a **new bench target** that `--all-targets` builds on a platform local preflight is blind to.
- `just risk-radar-test` — exit 0, twice; this diff moves root `CLAUDE.md` prose, whose anchor gate is CI-only.
- `just lint` — exit 0 with all seven workflow-relevant arms confirmed green (zizmor, actions, ci-obs, composites, prose, links, guides).
- `node scripts/gen-readme.mjs --check` — exit 0, run *before* trusting the README edit (generated sections are install/features/tools; the badge row is hand-maintained).
- Benchmarks run locally for shape (1s warm-up / 3s measurement), numbers above.

Three things verified rather than reasoned about, because each could have
silently produced a meaningless benchmark:

1. **The 12-agent cases are unchanged.** The busy/idle mix is now proportional
   (n/2 active, n/6 waiting) instead of hardcoded index ranges. At n=12 that
   resolves to the same 6/2/4 split — asserted against the old rule before the
   change landed, so the existing four benchmarks keep their CodSpeed history
   instead of resetting it.
2. **The crowds actually seat.** A 360x240 floor lays out 80 desks; confirmed
   32/32 and 64/64 reaching `SimFrame.characters` before trusting the numbers.
3. **The new bench's guards have teeth.** `assert_clean` + registered/wire_events
   are the negative control. Corrupting one synthesized line makes it panic with
   "1 line(s) are not valid JSON" rather than quietly benchmarking the decoder's
   error path. Confirmed by doing exactly that, then reverting.

No production behavior changes, so there is no new-behavior test to add — the
checklist item below is left unticked rather than ticked vacuously.

## Visual verification

- [x] Not a visual change.

## Checklist

- [x] I read the relevant `CLAUDE.md` (root + `pixtuoid-core`, `pixtuoid-scene`, and the tests guide).
- [ ] New behavior has a test — **n/a, no production behavior**; the new bench carries its own negative-controlled guards (above).
- [x] No `unwrap()` in non-test code; the benches use `.expect()` on dev-only targets, matching the existing `render_frame.rs`.
- [x] Docs updated in the same commit — root `CLAUDE.md`, `docs/CONTRIBUTING.md`, and the `codspeed.yml`/`bench.yml` headers no longer describe the old single-mode, scene-only shape.
- [x] Checked against the recurring pitfalls — see the note below.

### Pitfall check — one live hit worth reviewer attention

Pitfall 2 (parallel-implementation drift): the bench's `CEILING = (360, 240)`
is a second copy of `pixtuoid`'s `FLOATING_DEFAULT_W/H`, with no bridge test
pinning them equal. It genuinely cannot be single-sourced — the DAG runs
`pixtuoid-scene <- pixtuoid`, so a scene bench cannot import the binary's
config — which is exactly the cross-crate case the magic-number rule says to
pin with a test or `debug_assert!`. I left it as a documented pairing in the
const's comment instead, on the grounds that a benchmark *parameter* is not a
contract and drift degrades representativeness rather than correctness. Flagging
it rather than burying it: if a reviewer wants the pin, it belongs in
`pixtuoid`'s test tree asserting against the literal.

Pitfalls 1, 3, 6 are n/a (no text slicing, no decode-boundary change, no
denylist). Pitfall 4 is addressed by the negative control above. Pitfall 5:
every addition is wired in the same diff — the core bench into both `just bench`
and `codspeed.yml`'s `-p` list, and `machete` confirms the new dev-dep is used.

### On the two upstream facts this leans on

`cargo codspeed build -p A -p B` and `mode: simulation,memory` are **not** in
CodSpeed's docs, which document a single flag and omit the rest. Both were
verified against source rather than assumed: `package: Vec<String>` with
`#[arg(short, long)]` and `--measurement-mode` as `ArgAction::Append` +
`value_delimiter = ','` in `cargo-codspeed`'s clap definitions, and `mode`
documented as comma-separated in the action's own `action.yml`.

The `timeout-minutes: 30 -> 45` bump is **a hedge, not a measurement**.
Measurement runs per mode, but the action forwards `--mode=` to CodSpeed's
distributed runner binary, so the exact multiplier isn't derivable from source
the way the flag semantics were. Worth tightening once the first run reports.

One note on the README badge: it is a **static logo chip**, not a metric.
CodSpeed serves one shared endpoint payload and has no per-repo `badge.json`
(verified — `esphome/esphome`'s 404s too, as a negative control), so unlike the
codecov and CI badges beside it, this one never changes value and never goes
red. Its value is the link: the dashboard is public, so it takes visitors
straight to the flamegraphs.

## AI assistance

https://claude.ai/code/session_01AQULEd97BBLcYgYwGM3VDi

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -16,7 +16,7 @@ permissions:
 
 jobs:
   bench:
-    name: render benchmarks (on-demand, advisory)
+    name: frame + wire benchmarks (on-demand, advisory)
     runs-on: ubuntu-latest
     timeout-minutes: 30
     steps:
@@ -28,7 +28,7 @@ jobs:
         with:
           prefix-key: v1-rust # match setup-cargo-just's default cache namespace
       - uses: extractions/setup-just@v4
-      - name: Run render benchmarks (advisory, non-blocking)
+      - name: Run frame + wire benchmarks (advisory, non-blocking)
         run: just bench
       # JSON measurements only — the lean criterion feature set builds no HTML.
       - name: Upload criterion measurements
--- a/.github/workflows/codspeed.yml
+++ b/.github/workflows/codspeed.yml
@@ -1,15 +1,20 @@
 name: codspeed
 
-# Instrumented (simulation-mode) benchmark trend on Rust-touching PRs + main
-# pushes — CodSpeed counts weighted instructions, not wall-clock, so
-# shared-runner noise doesn't apply (<1% variance; the same setup ruff/uv
-# run). ADVISORY by construction like bench.yml/mutants.yml: nothing `needs:`
-# it and it is not a required check — trend evidence, never a gate. Reports
-# need the CodSpeed GitHub App installed on the repo; auth is OIDC
-# (`id-token: write`, job-scoped per this repo's convention — see
-# ci-tests.yml's "Declared, not inherited"), no token secret. Steps follow
-# CodSpeed's documented criterion workflow; moonrepo/setup-rust installs
-# cargo-codspeed via binstall (revisit if taiki-e/install-action adds it).
+# Instrumented benchmark trend on Rust-touching PRs + main pushes, in TWO
+# modes: `simulation` counts weighted instructions (not wall-clock, so
+# shared-runner noise doesn't apply — <1% variance; the same setup ruff/uv
+# run) and `memory` counts heap allocations. ADVISORY by construction like
+# bench.yml/mutants.yml: nothing `needs:` it and it is not a required check —
+# trend evidence, never a gate. Reports need the CodSpeed GitHub App installed
+# on the repo; auth is OIDC (`id-token: write`, job-scoped per this repo's
+# convention — see ci-tests.yml's "Declared, not inherited"), no token secret.
+# Steps follow CodSpeed's documented criterion workflow; moonrepo/setup-rust
+# installs cargo-codspeed via binstall (revisit if taiki-e/install-action adds
+# it).
+#
+# Both modes are named on BOTH steps because build and run read the mode
+# independently (the run inherits it from the action via CODSPEED_RUNNER_MODE).
+# Measurement is per-mode, which is what the timeout headroom is for.
 on:
   push:
     branches: [main]
@@ -25,9 +30,9 @@ permissions:
 
 jobs:
   codspeed:
-    name: benchmarks (codspeed simulation, advisory)
+    name: benchmarks (codspeed simulation + memory, advisory)
     runs-on: ubuntu-latest
-    timeout-minutes: 30
+    timeout-minutes: 45
     permissions:
       contents: read
       id-token: write
@@ -41,8 +46,8 @@ jobs:
           cache-target: release
           bins: cargo-codspeed
       - name: Build instrumented benchmarks
-        run: cargo codspeed build -p pixtuoid-scene -p pixtuoid-core
+        run: cargo codspeed build -p pixtuoid-scene -p pixtuoid-core -m simulation,memory
       - uses: CodSpeedHQ/action@v5
         with:
-          mode: simulation
+          mode: simulation,memory
           run: cargo codspeed run
--- a/.github/workflows/codspeed.yml
+++ b/.github/workflows/codspeed.yml
@@ -12,9 +12,10 @@ name: codspeed
 # installs cargo-codspeed via binstall (revisit if taiki-e/install-action adds
 # it).
 #
-# Both modes are named on BOTH steps because build and run read the mode
-# independently (the run inherits it from the action via CODSPEED_RUNNER_MODE).
-# Measurement is per-mode, which is what the timeout headroom is for.
+# TWO action steps rather than one `mode: simulation,memory`: the runner uploads
+# only after EVERY mode's run part completes, so a single process drops an
+# already-good simulation result whenever the newer memory instrument fails.
+# The build takes no mode — both instruments consume the same build today.
 on:
   push:
     branches: [main]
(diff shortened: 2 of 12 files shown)


IvanWng97:
## Review disposition — 3 lenses, every finding to a terminal state

The comment-lens escalation trigger fired (this diff reshapes module headers + a workflow header), so the gate ran **three** differentiated lenses. Verdicts: **APPROVE-WITH-NITS** (correctness) / **REQUEST-CHANGES** (design) / **REQUEST-CHANGES** (comment). Each MEDIUM+ finding was re-verified against source before acceptance. Folded into 4d0c4c21.

### FIXED

| # | Finding | Fix |
|---|---|---|
| **F2** | `mode: simulation,memory` in ONE step **loses data** — `orchestrator.rs::execute` runs `run_executor(..).await?` inside the per-mode loop and `upload_and_poll` *after* it, so a memory failure discards the completed simulation result | Two action steps; memory `continue-on-error: true`. Same single build, no extra CI minutes |
| **F1/N1/C1** | `CEILING` asserted a **false universal** — all three lenses caught it independently | Renamed `FLOATING_DEFAULT`, typed as `Size`, doc scoped to what's true |
| **F4** | Wire-event guard one factor short of its own comment: a decoder arm that DECLINES (`Ok(vec![])`) halves the count while `assert_clean` stays green | Exact counts (`2*TURNS`, `4*TURNS+1`), negative-controlled |
| **F6** | Occupancy axis was busy-only while the doc attributed the delta to specific mechanisms | Added the idle twin of the largest crowd; dropped the unmeasured attribution |
| **F3** | `-m simulation,memory` on the build step inert, comment claimed it load-bearing | Both dropped |
| **F5** | Memory curve carries a harness component production doesn't pay | Documented as directional |
| **F8** | Bare tuple where `Size` is imported | `Size` end to end |
| **C2** | #900's ratio inline — this branch's own CONTRIBUTING edit says measurements go in commit messages | Moved |
| **C5** | "80 desks" was an unpinned premise | Runtime `assert!` |
| **C8** | `just bench 360` now selects 4 cases; recipe said "only the 360x240 case" | Corrected |
| **N2** | Cross-crate copy uncommented | Paired at the offscreen.rs row that already reds |
| **N3** | `Drive` call-site list missing the bench | Added |
| **C4/C6/C7/C9** | Narration, "full floor" vs 80 desks, dead clause, restated pointer | Trimmed |

### REFUTED — with trace

- **timeout 45 → 60** (F7, conf 70). Reverted to **30** instead. The lens reasoned from bench count but explicitly declined to invent a memory multiplier; the first real run measured **3m11s** for both modes vs ~2m10s single-mode on main — +1 min, not 2×. My own 45 was an unmeasured hedge and is gone too.
- **Published `.crate` ships a non-compiling bench.** Real, but pre-existing on main (`decoder_fuzz.rs` does the same), anchored by `pixtuoid-core/SHARP-EDGES.md`, and `cargo package` is exit 0 — verify builds the lib only.
- **F9** occupancy/layout confound — verified inert: `frame_layout` passes `max_desks: None`, so both crowds render the same 80-desk office.

### ISSUE-FILED
None — every finding was small enough to fix in-arc.

### Note on my own error
The ceiling claim was wrong in the PR body too. My evidence sampled five conventional 16:9 aspect ratios — a population that couldn't reach the claim I drew from it. `office_scale` divides by **height only**, so width is unbounded at scale 1 (1600×260 → a 1600×260 buffer) and the scale-1 band runs to h=**269**. The decision not to add a third size still stands, on the linearity leg alone.

---
`just preflight` exit 0 (3060 tests) · both benches green under `--test` · the four original bench IDs unchanged, so CodSpeed history survives.

⚠️ Both bots' `Findings: 0` were keyed to 661af1c6 and are now **stale** — this head needs a fresh verdict.

PR #929, opened 2026-08-15
feat(fixtures): record them off the wire, and gate what a fixture may claim

IvanWng97:
## Summary

Fixture bytes become **evidence**: recorded off the real CLI instead of composed, every scenario declaring its provenance under a test that gates it. A composed fixture pins whatever its author believed the wire looked like, and the decoder is then written to agree with it — so the pair is self-consistent and silent. Recording them found **six real decode bugs**, every one of which had a passing test sitting on top of it.

Also lands the live-source e2e tier from an unmerged local branch, without the release runbook that wrapped it.

## The seven bugs the old fixtures were hiding

| source | the bug | what it looked like to a user |
|---|---|---|
| **kimi** | spells the per-call id `tool_call_id`; the shared CC-shaped arms read `tool_use_id` | every id decoded `None` for the whole source's life, so hook-wins dedup and `active_tasks` were bypassed |
| **claude-code** | `PermissionRequest` was never registered as a hook | a session parked on a permission prompt kept rendering as *working*, indefinitely |
| **reasonix** | the AgentId was keyed on `cwd` | two sessions in one project merged into one sprite |
| **grok** | no `ModelInfo` arm (`update._meta.modelId`) | the model flame stayed dark |
| **cursor** | no `ModelInfo` arm — and the model CHANGES mid-session (`composer-2.5-fast` → `composer-2.5`), so it must be re-read on every event, not just at start | same |
| **hermes** | no `ModelInfo` arm (`extra.model`) | same |
| **hermes** | the approval gate `pre_approval_request` was never registered | a session parked on an approval prompt kept rendering as *working* — the claude-code bug, one CLI over (#930) |

None errored. The optional-field ones are indistinguishable from an absent value, and **the tests could not have caught any of them**, because each fixture was composed from the decoder's own belief and confirmed it back.

Two more findings were shape-level rather than bugs: modern codex's rollout carries no escalation marker on the `custom_tool_call` surface (the gate exists only on the hook wire there), and openclaw's teardown order was pinned backwards — `session_end` lands AFTER `gateway_stop`, because shutdown is what closes the session.

## Related issue

n/a — context: #901 (cursor's `tool_use_id` drift) found the same gap the expensive way.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] New theme / sprite pack
- [ ] Docs only
- [x] Refactor / chore

## What's in it

**The recorder** — `just capture-fixture <source> <scenario> <cmd…>`, driving `crates/pixtuoid-core/examples/capture_fixture.rs`. It records at the **shim's OUTPUT**, by pointing `PIXTUOID_SOCKET` at a listener of its own: the one seam that does not care how the payload reached the shim, and it leaves your installed hook config alone. Recording the shim's INPUT cannot cover every source — codewhale is env-mode, so a stdin tee captures a file of empty payloads. The CLI invocation is the caller's to pass; a table of ten CLIs' flags would drift silently, and a drifted row captures the wrong thing while still looking like evidence. It runs in a fixed generic workspace so the bytes ship unedited (every payload embeds its own cwd and transcript_path), harvests the transcript by BIRTH time rather than mtime, and never overwrites a committed capture.

It was a shell script first. That version accumulated nine defects that were all bash semantics rather than recording logic — an empty array under `set -u`, `pipefail` swallowing a lister's exit 3 *after* the turn was billed, `head -1` taking SIGPIPE once a corpus passed ~700 files, `stat -f %B` being macOS-only, three world-writable temp paths — and **none of them could fail in a selftest small enough to run in CI**. In Rust they are absent by construction, and the decisions are ordinary `#[test]`s that `just test` already runs. The tool is Unix-gated: it records at a `UnixListener`, and the Windows shim speaks a named pipe.

**The provenance gate.** Nothing IN a fixture separates a capture from a composition — the one edit a capture is allowed is redacting PII, so an invented `/Users/dev/proj` and a redacted one are the same bytes. Each scenario carries a `provenance.json`; `every_scenario_declares_its_provenance` enforces the schema, and `scripts/fixture-age.py --check-metadata` gates that the fields stay parseable. `unknown` is an admission, not a third way to leave a fixture unexplained.

**Backfill: 37 recorded, 4 composed, 1 unknown** (from 4 / 15 / 6 when this PR opened). `NO_WIRE_EVIDENCE_YET` is now **empty** — every hook-only source has real wire evidence. The five that are not `recorded` each carry a reason worth keeping:

- `claude-code/proof-session` — read at RENDER time by the site-media renderer; its timestamps ARE the clip's beat timings.
- `codex/permission-flow` — a legacy wire shape the current codex can no longer emit, so it cannot be re-recorded.
- `copilot/tool-run` — the only fixture carrying a real Windows `cwd` (spaces and parens), which a macOS host cannot reproduce; it is the Windows arm of `registry_cwd_extractor_matches_each_sources_real_head_shape`.
- `openclaw/gateway_mid_attach_failed_run`, `opencode/session-run` — kept beside recorded siblings that pin the same decode.

**The per-call id's wire name is a registry row**, `HookDecoding.tool_id_key`, exactly the shape `id_key: IdKey` already is. The first fix mirrored ~40 lines of the shared arms inside kimi for one field name, and the copy had already dropped `effort_info()`.

**Drift and freshness tooling.** A recorded fixture pins ONE version of one CLI's wire, and nothing else in the tree noticed when that CLI moved — codex switched its whole tool surface between 2026-07-25 and 2026-08-14 and the corpus said nothing. `scripts/fixture-age.py` reports version drift (advisory, exit 3) against each CLI's own probe, read from `corpus_check --roster` rather than a hand-copied table.

**Merged from a local branch:** `scripts/lib/tier-live-sources.sh` — launches each installed agent CLI and asserts ITS badge reaches a lifecycle state in one headless pixtuoid. The interactive 28-step release driver it was wrapped in is deleted; the tiers are the part with teeth.

## Six review rounds, and what they cost

Each round is its own commit. The two that mattered most were found by a lens reading **all 172 files** rather than the delta — a delta-scoped re-review structurally cannot see what round 1 missed:

- **The drift ledger would have excused this PR's own headline bug.** Registering `PermissionRequest` left it sitting in `check_upstream_drift.py`'s `CC_KNOWN_OMITTED` under the rationale "permission detail already covered by Notification" — the exact belief a capture had just refuted. Inert today; the day someone drops the registration, the watch stays silent. CC was the only member of that whole family contradicting its own `*_EVENTS`. Removed, plus a gate that intersects every ledger with what we register.
- **A test named for the anti-spoof rule never tested it.** `an_ordinary_tool_never_carries_the_task_detail` asserted `starts > tasks()` — a precondition on the FIXTURE. A decoder that tagged the bash call and left the real dispatch bare passed it. Replaced by one that reads the `task` call's own id out of the raw capture and binds the Task detail to it.

Also folded: a live-tier regex that spanned sibling agent entries (one source could PASS on another's state); that tier reporting its OWN timeout kill as "the CLI itself failed" and downgrading a real regression to *uncovered*; a settle window written as the difference of two unnamed literals, which made its restart-on-gate arm dead code; a `|`-arm parser with no control that could fail; and my own PII (an account email ×53, and a 2.3 KB personal instructions block) redacted out of a public repo whose `tests/` ships to crates.io.

## How I tested it

- `just preflight` — **rc=0**, 3107 tests passed, 2 skipped.
- `just check-windows` — **rc=0**. It was red: the new recorder opens a `UnixListener`, `--all-targets` builds examples, and three Windows jobs went down. preflight never cross-compiles, so four green preflights said nothing about it.
- `just drift-selftest`, `just tuidrive-selftest`, `scripts/fixture-age.py --check-metadata` — each rc=0, and **each new gate negative-controlled**: the ledger gate reds when `PermissionRequest` goes back in the ledger; the Task-binding test reds when opencode's `SUBAGENT_TOOLS` points at `read`, naming the exact call; the `|`-arm control reds on a first-alternative-only parser.
- Real captures, billed, across claude-code, codex, grok, kimi, omp, opencode, openclaw, copilot, cursor, hermes, reasonix, codewhale and antigravity.

## Visual verification

- [x] Not a visual change.

## Checklist

- [x] I read the relevant `CLAUDE.md` (root + `pixtuoid-core`, its `tests/`, and `pixtuoid-scene`).
- [x] Docs updated in the same commit.


IvanWng97:
## Review disposition — 4 lenses + a re-verification pass, at `92f4fded`

Re-reviewed the **whole PR** (185 files, +5637/−1027) rather than the delta, with
four differentiated lenses: **correctness**, **design**, **comment** (escalation
trigger: the diff reshapes a large block of comments), and **lifecycle**
(escalation trigger: it re-keys a source's AgentId and newly keys state
transitions on wire signals). Plus a re-verification of the earlier full-diff
lens's 19 findings.

Every finding below is in exactly one terminal state. Lens verdicts on arrival:
correctness APPROVE-WITH-NITS · comment APPROVE-WITH-NITS · design
REQUEST-CHANGES · lifecycle REQUEST-CHANGES. **Both blockers are fixed** — design's
was H2, lifecycle's was F1.

### FIXED — the two that mattered

| # | finding | fix |
|---|---|---|
| **lifecycle F1** (HIGH, 85) | Registering `PermissionRequest` **stranded the slot in Waiting**. CC fires the gate with no `tool_use_id`, then the idle `Notification`; both decode to `Waiting`. `apply_waiting` remembered the mid-flight tool on the first, then the second took the `else` arm and **deleted** it (a `Waiting` slot no longer matches `Active`). The approved tool's `PostToolUse` then failed `wait_resolved_by`, and CC has no `Stop` hook so the tuid-less turn-end escape never fires. Before this PR the `Notification` **was** the gate-setter. | The clear is now conditional — a re-notified `Waiting` slot is the same gate. Test written first, watched fail, walking the recorded capture's exact wire order. `0afaf345` |
| **design H2 / correctness F2** (97/96, found independently by two lenses) | **The recorder's 8 tests never ran.** An example is not a test target, so they compiled and were skipped — nextest's universe was 3089 and none were in it. The whole argument for the shell→Rust rewrite was "ordinary `#[test]`s that `just test` already runs, rather than a bespoke `--selftest`", which was false as written and *worse* than the shell, whose selftest had two gates. | `[[example]] test = true`, the wiring `pixtuoid`'s `snapshot` example already carried. 3087 → **3099**. `0afaf345` |

### FIXED — the rest

- **correctness F4** (95) — `CAPTURE_SEED` documented, implemented nowhere, so the re-record recipe for both gate fixtures was inert. Implemented + test.
- **design M2** (85) — the corpus walk existed **twice**, and the worse copy decides which bytes become a committed golden. Now `harness::transcripts_under`, one implementation, with the walk's pins moved beside it (1200-file scale shape, symlink refusal, plus a new one for the path filter — grok writes five `.jsonl` siblings and only `updates.jsonl` is the transcript).
- **design M3** (80) — `invocation_for` hand-copied twelve binary names, each byte-identical to `version_probe[0]` — the column this PR exposed so consumers would stop copying it. Now read from `--roster` field 5; verified all twelve resolve to what the table said, unknown id → empty.
- **design L2** (88) — **a gate assertion that could not fail.** `tuidrive.py`'s clamp check re-wrote the clamp instead of calling it, reducing to `ANSWERS[3] == ANSWERS[-1]` — true for any contents, inside a check `just lint` and ci-lint both run. Now `answer_for(n)`, asserted by calling it; mutating to `min(n, 2)` reds.
- **design L4** (78) — two codex captures shipped a `world_state` inventory: 57 installed skill names with paths + host timezone, **56 KB** across the two, in a repo whose `tests/` ships to crates.io. Dropped, and the golden snapshots verified **byte-identical** — which is what shows the redaction cost no evidence rather than merely looking safe.
- **comment C1** (92) — a WHY that was **backwards**: `session_end` was said to "leave the daemon Down"; it does the opposite, resurrecting a Down daemon to UP.
- **correctness F3 / design M1** (92) — `corpus_check --root` and `--list` were added for the shell recorder and orphaned by its replacement **in this same PR**. Deleted, like `--sources` before them.
- **design M4** (93) — `hook_command`'s doc still called the Unix command a bare name after this PR wrapped it in the `PIXTUOID_SOURCE=` prefix.
- **design L6** (90) — the `not(unix)` stub exited 0; every other "cannot do the job" path in that file exits non-zero. Exit 2.
- **design L5** — `gateway_mid_attach_failed_run` kept underscores while both siblings were hyphenated here. Renamed; `--unreferenced=reject` confirms no orphan snapshot.
- **design M5 / L5b** — two doc-drift lines: nothing said a capture can only be recorded on Unix (the reason `copilot/tool-run` stays `unknown`), and `tests/CLAUDE.md`'s "data scopes to the binary that reads it" no longer described the tree after `core_fixtures_root()` widened.
- **lifecycle nit** (90) — two hand-written kimi payloads still fed `tool_use_id`, the CC spelling this PR established kimi never sends.

### REFUTED — verified, not argued

- **correctness F5** (85) — "opencode carries a model id and got no `ModelInfo` arm." It has had one at `source/opencode.rs:127`, with tests, all along.
- **lifecycle nit, partial** — `kimi.rs:301`'s `tool_use_id` **stays**: that one is the deliberate negative control `the_cc_spelling_is_not_read_for_kimi`, whose whole job is to prove the CC name is *not* read for kimi.
- **design H1** — the Windows break. Real, and already fixed in `b1d447d6` before this lens reported; the lens was reading `3d0cd09f`.

### NOTED, no change

- **design L7** (95) — `Cargo.lock` patch bumps (`serde_json` 1.13.0→1.13.1, `similar` 3.1.1→3.1.2) rode in on the Windows fix commit. Harmless, `cargo-deny` green; naming them here rather than churning history.
- **comment lens count** — 36 advisory `comment-lint` candidates, each dispositioned by the lens. Net effect of this review round is **removal**: a 10-line run in `decode_hook_payload`, 8 lines in `install_target`, a 31-line module header, a 13-line one on `delegation/mod.rs`, and 3 lines on `claude_shim_ref` that were both redundant and false.

### Gates at `92f4fded`

Every exit code read in the foreground, never off a pipe or a backgrounded wrapper:

| gate | rc |
|---|---|
| `just preflight` | **0** — 3099 tests, 2 skipped |
| `just check-windows` | **0** |
| `cargo insta test --check --unreferenced=reject` | **0** — no orphan snapshots |
| `just drift-selftest` | **0** |
| `just tuidrive-selftest` | **0** |
| `scripts/fixture-age.py --check-metadata` | **0** |

Two process notes worth recording, both mine:

1. `just check-windows` went **green on the fix** while `just clippy` went **red** — the module is `cfg`'d out on Windows, so that arm never saw `unreachable_pub`. A cfg gate splits compilation in two and each check walks only one; both arms need running.
2. One intermediate check reported ✅ **vacuously** — a `find` returned two paths, the variable never resolved, nothing was modified, and the unchanged hash read as a pass. Redone with a script that fails loudly instead.

IvanWng97:
## Round-3 review disposition — 4 fresh lenses at `088234f5`

A second full-PR review with four **new** lenses (correctness, design, comment,
lifecycle), each briefed to find what the previous seven rounds MISSED rather
than re-derive them. Every finding below is in exactly one terminal state.

Arrival verdicts: correctness APPROVE-WITH-NITS · design APPROVE-WITH-NITS ·
lifecycle APPROVE-WITH-NITS ("nothing blocks") · comment REQUEST-CHANGES,
narrowed on re-read to two one-line edits, both now applied.

### The three that were real bugs, not review debt

| finding | what it was |
|---|---|
| **lifecycle F1** (MEDIUM 88) | A clean OpenClaw shutdown records `gateway_stop` → `session_end` **2 ms apart**, and `SessionEnded`'s "any event ⇒ up" arm resurrected the gateway that had just announced its own stop — `entered_at` re-anchored, so the mascot that had begun walking out walked back IN, for the 5-minute presence TTL. `apply_presence`'s own comment assumed the opposite order. Fixed at the creation-polarity rule: a session CLOSING is not proof the gateway continues. Pinned both ways, plus a sharp edge, because "harmonize the arms" is the obvious edit that would undo it. |
| **correctness F1** (MEDIUM 88) | The recurrence gate this PR added guarded the **inverse** direction: it intersected the ledgers with what we REGISTER (empty by construction) while the incident was a ledgered name our DECODER reads and we never registered. The lens reconstructed the pre-PR state and the gate stayed green. Now keyed on the decoder's arms, with `LEDGERED_BUT_DECODABLE` as a shrink-only ratchet — and two live members surfaced (`Stop`, `UserPromptSubmit`), each now carrying its reason. |
| **design F4** (85) | `live-sources` printed **PASS having covered nothing**: `jq` was unguarded (every sibling tier guards its tool), and without it all thirteen sources read "CLI not installed" — a full SKIP sweep exiting 0, identical to a host with no CLIs. `corpus-all` already drew this distinction with an exit-3 lane; this tier does now. |

### FIXED — the rest

- **correctness F2** — the recorder's anti-squat check used `fs::metadata`, which follows the very symlink the threat it names requires it not to follow. `symlink_metadata`, pinned by a discriminator test (a link we own pointing at a root-owned dir must read as ours).
- **correctness F3** — `--check-metadata` validated **nothing** when `origin` was absent or misspelled; one field switched the whole check off, and for the 8 provenance files outside the conformance tree this script is the only reader. Driven both directions.
- **correctness F6/F7** — two tests *I* added this round had no teeth: one asserted only `.is_some()` (a `min_by_key` mutant survived), one gated only its symlink CREATION on `cfg(unix)` so on windows-test it asserted that one file is one file.
- **correctness F4** — four `recorded` fixtures shipped redacted bytes with no `note`; the `/Users/dev` convention lived only inside individual notes. Stated once in the schema section, and the four annotated.
- **design F2 → #931** — `harness::transcripts_under` duplicated the watcher's admission rule, and it is the copy that picks which file becomes a committed golden. Now `source/admit.rs`: one ungated `classify`, each caller keeping its own traversal. `Entry` carries the skip REASON — my first cut folded both skips together and then re-asked `is_symlink()` at the call site, which is the same question asked twice, the exact smell the extraction exists to remove.
- **design F5** — `provenance.json` omitted `CAPTURE_PROMPT`/`CAPTURE_SEED`, the two inputs that produced the gate scenarios, while `command` records the UN-expanded argv (a literal `{prompt}`). Recorded now, written only when set.
- **design F6** — the `gated_before_waiting` sharp edge said FOUR clear sites; my own fix made a fifth. Counting them properly says **SEVEN** — `apply_activity_start` and `apply_activity_end` were never enumerated by anyone. Corrected against a grep, not against the previous sentence.
- **design F7** — the README told the recorder-is-Rust story twice, both claiming "nine defects" while enumerating seven. Duplicate gone, count gone; the enumeration was always the content.
- **comment A1–A6, B1–B2, C1–C4** — a 5-line doc block orphaned above `corpus_check::main` and truncated mid-sentence when its fn moved; a WHY that was backwards about `session_end`; two stale counts; a provenance note citing a flag that existed in **no commit**; and the redundancy cluster (the same WHY stated up to five times across module doc / fn doc / test comment).

### REFUTED — verified against the tree, not argued

- **correctness F5** — "opencode carries a model id and got no `ModelInfo` arm." It has had one at `source/opencode.rs:127`, with tests, all along.
- **the online bot's #1** — it asked `claude_shim_ref` to delegate wholesale to `shell_shim_ref`. Diagnosis right, prescription wrong: CC's Windows entry is exec form, a bare absolute path with neither quoting nor `--source`, which that parser's last-whitespace-token arm truncates at `C:\Program Files\…`. Split by the form that actually exists instead.
- **`kimi.rs:301`'s `tool_use_id`** stays — it is the deliberate negative control proving the CC spelling is *not* read for kimi.

### The one worth reading twice

The comment lens's two remaining blockers were both **reported folded in a commit
message while neither file had been touched** — a batch script died on the second
of four anchors and I read the first "fixed" line. It refused to clear on the
report rather than the tree, which is how a false claim in my own commit body got
caught. Both are applied now, and its cross-lens nit became a real gate: the
hermes arm asserts upstream's `_BLOCKING_EVENTS` is `{pre_tool_call}` alone —
the entire reason registering an approval gate is safe — and nothing watched it.
`check_upstream_drift.py` does now, driven through all three outcomes (BREAKING
when the gate joins the set, silent when it does not, probe-health *blind* rather
than falsely-green if the constant is renamed).

### Gates at `088234f5`, each exit code read in the foreground

| gate | rc |
|---|---|
| `just preflight` | **0** — 3107 tests, 2 skipped |
| `just check-windows` | **0** |
| `cargo insta test --check --unreferenced=reject` | **0** |
| `just drift-selftest` | **0** |
| `just tuidrive-selftest` | **0** |
| `scripts/fixture-age.py --check-metadata` | **0** |

IvanWng97:
## Rounds 4–5 disposition — the adversarial and replay lenses

Two rounds aimed at the review rather than the diff, on the theory that twelve
rounds of reviewers correlate: they all read the diff, the guides, the sharp
edges. These read something else.

### Round 4 — attack the review's own claims

Verdict REQUEST-CHANGES, cleared. Three findings, all failures of MY verification
rather than of the code under review:

| finding | what it was |
|---|---|
| **the recurrence gate was vacuous for 6 of 13 sources** | It scraped decoder arms with `[A-Za-z]+` from `decoder.rs` alone, so it saw the 11 CamelCase CC-shaped arms and none of the snake_case per-source ones. Control: put `pre_approval_request` — demonstrably decoded — into a ledger; the gate said "all checks passed". Reproduced before believing it. Now `[A-Za-z][A-Za-z0-9_.]*` over every `source/*.rs`, with a 30-arm floor and a shape assertion per naming convention, because the old `len > 5` floor passed on the CamelCase half alone. **This gate had already failed once**, guarding the inverse direction; twice unable to catch the class it was written for. |
| **eight `recorded` fixtures were edited and said nothing** | My own sweep the round before keyed on `/Users/dev` alone and missed the ones redacted in the owner column of a captured `ls -la` — the case `fixtures/README.md` names as ITS example. Four sentinels found eight. Now a gate, not a sweep I run and get wrong. |
| **a vestigial doc** | `transcripts_under` still instructed the next agent to hand-maintain a copy that commit `9a947245` deleted, citing the issue that removed it. |

Plus the root cause under the first two: **the provenance gate could not falsify
anything it stored.** Rewriting a record to `cli: "codex", version:
"0.0.0-A-LIE", captured: "1999-01-01"` passed all 84 tests and the metadata gate.
Two axes ARE answerable from the bytes and now are — the payloads' own
`_pixtuoid_source`, and `captured` against the `_shim_ts_ms` stamps. The second
immediately found three records that had been wrong since the day they were
written: the shell recorder dated in LOCAL time while the shim stamped UTC.

From the fresh-reader lens the same round: `fixture-age` reported "none stale"
about 11 fixtures it never compared (now a third lane, and it surfaces 5 real
drifts the old summary folded away); `verified_version` had not moved for any of
the twelve CLIs this PR recorded, leaving `doctor`'s drift category structurally
silent for 8 of 13 sources; a re-record writes `*.new` that NO gate reads, so a
billed turn could change nothing while the suite stayed green.

### Round 5 — replay, not review

One question, asked of nine gates: restore the tree to the state BEFORE the fix,
and does the gate actually red? **8 PASS, 2 VACUOUS**, and of the eight, seven
also catch a neighbouring instance rather than only the bug they were born from.
The two vacuous ones are structural, not defects (one quantifies over a const, so
shrinking the domain silences it — that is #932; the other cannot distinguish
pre- from post-extraction because the refactor is behaviour-preserving).

**The finding: the diagnosis recurred at HEAD, in the same commit that named it.**
`a_recorded_capture_anchors_its_sources_verified_version` asserted
`!= "unknown"`, so `verified_version: "0.0.0-A-LIE"` against captures pinning
2.1.233 passed everything — and three of the ten anchors that commit moved were
STALE rather than unknown, so they were never in its range. Written in the commit
whose message reads *"a record nothing can contradict is a claim, not evidence"*.
The anchor must now APPEAR in one of the source's recorded versions; controlled
on both classes the old form missed.

Also: `install/hermes.rs` claimed registered `==` decodable "enforced by" a test
that enforces one direction — and the other direction is exactly what shipped
`pre_approval_request` unregistered.

### The pattern worth keeping

Both rounds found the same shape, and it is the PR's own thesis turned on its
author: **a fix and its negative control written from one mental model confirm
each other when the model is wrong** — structurally identical to a composed
fixture and the decoder written to agree with it. The method that breaks it is
what round 5 did: derive the control from the ORIGINAL INCIDENT, replayed, not
from a description of the class.

### Filed rather than fixed here

- **#932** — the missing-registration drift arm exists for 5 of 13 sources.
  Hermes was one of the 8 without it, which is why #930 was found by hand. Five
  upstream probes, two of them behind closed artifacts; its own change.

### Gates at this head, each exit code read in the foreground

`just preflight` **0** (3109) · `just check-windows` **0** ·
`cargo insta test --unreferenced=reject` **0** · `just drift-selftest` **0** ·
`scripts/fixture-age.py --check-metadata` **0**

Issue #940, opened 2026-08-17
drift: no decoder-read ratchet at all — 130 field reads across 13 sources are neither watched nor breadcrumbed


IvanWng97:
Surfaced by PR #938's review rounds and deliberately NOT fixed there — it predates that branch, and closing it inside a deletion PR would have meant adding three more hand-maintained mirrors.

## The gap

`test_no_decoder_read_is_unaccounted_for` (`scripts/check_upstream_drift_selftest.py`) is the ratchet that stops a decoder gaining a wire read nobody watches. Its docstring names the failure it exists to prevent — `sessionId` (reasonix's primary AgentId key) and `extra` (hermes's Waiting reason AND model flame) both sat unwatched.

But it derives its population from the `*_PAYLOAD_FIELDS` **suffix**, so it covers only the sources that happen to have such a set: reasonix, opencode, copilot, openclaw, hermes — **5 of 13**. `acp.rs`, `grok.rs` and `codex.rs` escape it entirely.

Concretely uncovered: `decode_session_update` reads `sessionUpdate`, `toolCallId`, `title`, `rawInput`, `status` and the values `completed`/`failed`; `decode_grok_line` also requires `params.update`. A renamed `toolCallId` would un-pair every ACP start/end with no alarm.

## Why it is not just "add three more sets"

`*_PAYLOAD_FIELDS` is a hand-maintained mirror. #938 deleted 221 such entries because a list mirroring someone else's vocabulary rots by construction; the review adjudicated that a set mirroring OUR OWN reads differs in kind, because it is derivable from in-tree source.

So the shape worth exploring is the one #938 used for `read_codex_rollout_outers` / `read_acp_decoded_tags`: **derive** the read set from the decoder's own `.get("…")` calls instead of declaring it. The census already scrapes those calls to find unaccounted reads — the same scrape could BE the population.

The caveat that killed the naive version in review: a derivation parser returning a silent SUBSET fails OPEN — the check passes while watching less. #938 handled it by flooring each derived reader in the selftest (`>= N` plus a shape regex) and gating the N-of-N census so a row without a floor case cannot exist. Any fix needs both guards.

## Acceptance

- every decoder's top-level reads are watched or ledgered, for all 13 sources
- the population is derived, not declared — or the declaration is census-pinned
- a reader returning a subset reddens `just lint` rather than passing quietly

IvanWng97:
[...]

PR #942, opened 2026-08-18
refactor(drift): watch by transport and by whether the decoder can speak

IvanWng97:
## Summary

**`+2,863 / −2,624`.** One rule, in two clauses: **watch by TRANSPORT, and by
whether the decoder can speak.**

- A **hook-registered** CLI always needs an upstream watch. Registration is
  name-keyed, so a rename leaves our entry inert, the CLI fires nothing, the
  decoder is never reached, and no runtime breadcrumb can cover it. Silence, not
  a signal.
- A **transcript-bearing** CLI needs one **only if** its decoder stays silent on
  an unrecognised line. `omp` and `codex` both end `_ => vec![]`, so their
  transcript vocabularies are watched; the decoders that breadcrumb are not.

That rule holds for all 13 sources, verified source by source. It replaces
#938's premise — "closed CLIs are covered at runtime" — which is false for
exactly the hook-registered ones.

The watch that existed answered "did upstream rename something" with a scraper
whose vanish direction had **zero true positives**:

| issue | verdict |
| --- | --- |
| #406 | **false** — upstream restructured, our URLs went stale |
| #728 | **false** — our own WHY comment leaked into the parser (closed by the owner as "False positive from the watcher") |
| #793 | **false** — stale pin, upstream moved the file |
| #710 | **real** — but the APPEARING direction, which #938 deleted |

Roughly half that file existed to suppress its own noise: the anchor gate is
#793, `strip_rust_comments` is #728, the believability floors are #929. Those
suppressors are **kept** — they are what makes a source-tree watch honest.

Meanwhile the thing it exists to catch went by unseen — `fixture-age.py`'s own
header records that *"codex switched its whole tool surface between 2026-07-25
and 2026-08-14 and the corpus said nothing — that switch was found by accident."*

## What replaces it

Nothing new. The defenses were already there; this writes the ladder down at the
seam it governs (`source/drift.rs`), ordered by **whether it can lie to you**,
and adds a second axis: **whether it can be silent.** A defense that never
speaks never lies.

1. **Don't key on the name.** `subagent_type` presence, the payload a renamed
   type still carries. Cannot lie, and drift stops *mattering* rather than being
   detected.
2. **Breadcrumbs + recorded fixtures.** Cannot lie either — both meet the real
   wire, on the machine running that CLI, at the moment it happens. Verified end
   to end this session: `logging.rs` writes the warn floor →
   `doctor::scan_log_for_source` counts per source with samples →
   `drifted_sources()` → the live TUI footer (`tui/ui_state.rs:75`). But it is
   *silent* for the two cases above, which is why (3) survives.
3. **`check_upstream_drift.py`, against whatever DECLARES those names** — source
   tree included (18 of the 25 documents), because that is where a hook enum
   lives. It can lie, so every sweep is gated: an ANCHOR proving the document
   still owns the names (#793), a believability floor before either direction
   (#929), and a disposition split so a stale pin reads as **probe health**
   rather than drift.
4. `every_registered_*_event_decodes` asserts our own belief, so it can never
   see a rename — it binds registration to decoding, nothing more.

And the rule an alarm must obey: **state a fact about US** ("this decoder met X
and dropped it"), **never a guess about THEM** ("upstream renamed X"). Every
false alarm this watch produced was the second kind.

## Also: stop parsing our own source

All 16 of the watcher's readers regex-parsed **our** Rust — `CODEX_EVENTS in
install/codex.rs`, the `match kind` arms in `source/copilot.rs`. They existed
because the watcher is a separate program that cannot call us, and a scraped arm
goes stale silently: rename a const and the parser returns a *smaller* set, so
the watch narrows and nothing fails.

Each crate now emits its own half as data, from a test inside the crate that
owns the names, so a rename is a test failure **there**. Two fragments because
each crate's names are private to it, and because registration and decoding
answer different questions — the `*_EVENTS` sharp edge in
`crates/pixtuoid/SHARP-EDGES.md` is exactly that split.

## Single source of truth — measured, not asserted

```
dependency names: 142, all from the fragments; hardcoded duplicates in checks: 0
emitted keys: 24 (core 14 + bin 10) consumed rows: 24   symmetric difference: EMPTY
blind lines when reading the fragments: 0
```

Pinned by tests that are each negative-controlled — verified to redden when
broken, with the harness itself controlled by a null mutant first: the decode
arms vs their exported set, every `DECODED_*` export being a row or naming its
pin, the consts vs the committed fragment, byte-stability under `preserve_order`
feature unification, **emit ↔ consume in both directions**, the TS plugin vs the
Rust decoder across a language boundary, per-key isolation, and every swept URL
being anchored or structurally parsed (#793's gate).

**Every one of the 24 `add_breaking` arms fires when the thing it watches
changes, and stays silent otherwise** — a vanish for the 23 presence arms, an
*appearance* for the hermes blocking-events one, which is why it has its own
two-direction test rather than a row in the vanish census. All offline, driven by documents built from the set we actually depend on
in each upstream's own spelling — and shaped like the real documents in the
dimensions the parsers defend (struct variants, so `_enum_body` must brace-balance
rather than stop at the first `}`; grok's `hook_events!` macro table, which is the
branch its real document actually takes). The case list binds to `SURFACE_ROWS` by
set equality, so a new surface row without a case fails rather than passing quietly.

Two mutation censuses, each behind a null-mutant harness control: **24/24**
`add_breaking` arms killed, and **5/5** parser mutants. Before the fixtures were
reshaped three parser mutants survived — one a first-brace `_enum_body` regression
that files nine false ⛔ against the real documents while the selftest stays green.

The **registration** half is bound too: `every_install_events_const_reaches_the_fragment`
derives from the `*_EVENTS` consts in `install/`, so a new hook-registered CLI
cannot reach the roster with no watch. No exemption list — opencode registers
through its TS plugin and declares no such const, so it is absent by construction
rather than by exception.

## Numbers

|  | before | after |
| --- | --- | --- |
| `check_upstream_drift.py` | 2,150 | **1,566** |
| selftest | 1,572 | **993** |
| `add_breaking` call sites | 48 | **24** |
| upstream documents | 30 | **25** (22 anchored + 3 JSON Schemas parsed structurally) |
| covered by its own case | — | **24 / 24 surface rows** |

Both columns of every row use the same ruler: the `add_breaking` row counts call
sites on both sides, and the documents row counts distinct URLs by AST on both.

## What this branch had deleted, and what came back

Review found this branch had narrowed coverage, and the pattern took five
findings from three different reviewers to see: **`SURFACE_ROWS` carries name
SETS, so every wire fact that is a single VALUE fell outside the machinery.**
All five are restored and now ride the surface:

| value | rename it upstream and… |
| --- | --- |
| grok `_x.ai/session/update` + its 5 xAI tags | the whole xAI block stops decoding |
| hermes `_BLOCKING_EVENTS` membership | an always-exit-0 shim ANSWERS an approval (invariant #5) |
| ACP `completed` / `failed` | every ACP tool call stays Active forever |
| OpenClaw `DEFAULT_GATEWAY_PORT` | two live gateways collapse onto one mascot |
| omp `session_exit` | no omp session ever ends cleanly |
| omp `assistant` / `toolResult` / `toolCall` / `ask` | the turn decodes to nothing, or an `ask` strands its Waiting gate |

Then a seventh, and it is the one worth reading: **omp's `toolCall` and `ask`**
(plus the two roles beside them). My own census for #943 had missed them, because
it split production code at `src.split("#[cfg(test)]")` — and this branch gates
its exported consts with that *attribute*, so the split stopped at the first one
and hid 541 lines of `grok.rs`, 577 of `cc_probe.rs`, 182 of `omp.rs`. Cutting at
the test MODULE instead, the two published counts were both wrong: **130**
unaccounted field reads, not 93, and **20** equality-compared values, not ~10.
Both issues carry the correction and the reason.

Three of the seven were not review findings at all — ACP fell out of a new
depth-independent bare-literal ban, OpenClaw came from the online bot, and omp's
two came from diffing the swept-document set against base. #943 tracks the census
that makes an eighth impossible; its remaining population is **pre-existing**,
verified unwatched at `267c52b4` too.

Two axes are adjudicated rather than restored, and `source/drift.rs` states both
limits where the rule lives: payload FIELD reads (**130 across the 13 decoders,
neither watched nor breadcrumbed** — #940, restated after its original body
described a test this branch deleted) and equality-compared values (**20**, #943).

## Verification

A golden-record parity harness over 30 cached real upstream documents — 2,707
mutation cells reaching 47 of the 48 check sites — proved the *emit* refactor
behaviour-preserving before any deletion (`PARITY HOLDS`). It earned its cost
immediately: the first run showed 10 cells differing, because my arm extraction
used a line-range cutoff and missed copilot's `session.task_complete` and
`session.shutdown`. Those two watches would have been dropped silently.

Four flaws in the harness itself had to be burned down first — a suffix rename
cannot fire a prefix check; a prefix rename cannot fire a substring check (a
rename must REPLACE, not decorate); a single-document rename cannot fire a check
whose upstream set is unioned across documents; and our spelling is not
upstream's (`event_msg` vs `EventMsg`).

Two designs were built and reverted rather than shipped, both after the owner
asked the obvious question: a version-gap alert (unbounded noise — Reasonix ships
~16 releases/week) and a 90-day fixture-age timer (a calendar wearing a monitor's
clothes). `4f0bf211`+`51750540` is that round trip, kept in history rather than
squashed.

Gates at this head: `just preflight` exit 0, 3,154 tests; drift selftest exit 0
with **0 network calls**; live upstream baseline `breaking=0 review=0 blind=0
exit=0`.

Docs corrected in the same change, because the pivot made them lie: two
`SHARP-EDGES` entries still sent readers to "the VANISH direction", and
CONTRIBUTING's add-a-CLI step 11 still told you to add a scrape for any
open-source upstream. Step 11 is now **conditional** — it asks the two questions
above and points at `source/drift.rs` for the reason.


IvanWng97:
**Convergence-contract note (two-lens-review v1.2):** churn = 3,460 added vs the ~1,500 budget. Owner override recorded — this branch predates the contract (#944) and has been through six review rounds already; the budget clause is waived for this PR only. Every other clause applies in full: two-fix-round cap, orchestrator-confirmed-HIGH-only blocking, round-2 fix shapes (revert / deletion / red-test), no new gates in a fix round.

IvanWng97:
## Round-1 disposition (two-lens-review v1.2) — head `a68b0850`

Five real lens returns: comments (REQUEST-CHANGES, 24 items), design (APPROVE-WITH-NITS, 9), correctness×3 (APPROVE-WITH-NITS each; the full run + two focused halves after infra deaths). 47 findings, every one at a terminal state:

**FIXED in the fold (`a68b0850`)** — comments H1–H5, M1–M10, L9; design D1, D2, D4, D5, D6, D9(a–d); correctness F1, F2(acp half), F3(via `opencode.part_statuses` export + docstring), F6; harness N1(msg), N2; watches F1(docstring + the opencode escape), F2, F3.

**REFUTED-with-trace** — none survived verification falsely; all premises I checked held.

**SURFACED (pre-existing; owner decides, PR does not grow)**
- kimi `AgentSwarm` dispatch injects no `subagent_type` default → an omitting swarm call mints Generic and the Waiting-class stale window is not applied (bundle-verified).
- grok pager `notificationType` values (`permission_prompt`, `elicitation_dialog`, …) drive Waiting, are declared in a fetchable upstream doc, and are unwatched — the silent-DEGRADATION class the rule marks as owed.
- copilot `result.kind.starts_with("approved")` — the already-fetched schema declares the three `approved*` consts; a rename decodes every approval as a denial. Cheapest possible export if wanted.
- cc_doc_marker → `add_review` WIRING is unpinned (the fn is tested, the call is not).
- codex `EM_*`/`RI_*` groups sit outside the `DECODED_*` naming convention and are unguarded on the emission axis.
- Two Python ledgers (`CODEX_KNOWN_OMITTED`, `GROK_XAI_KNOWN_OMITTED`) have no stale direction; consolidation on shared identity proposed by design D7.
- Census exemptions key on file STEM (D8); collision class currently covered by the skip list only by coincidence.
- Core has no "every export reaches the fragment" census (design D3); a list-free mirror of the bin test exists as a design — **barred from this fold by the contract's no-new-gates clause; its own small PR if wanted.**

**DROPPED (taste, per the blocking bar)** — comments L1, L3–L8; design D7/D8 fixes beyond the surfaced lines.

Contract state: round 1 of 2 consumed. Round 2 = verify dispositions + delta-only review; any round-2 fix must be a revert, a deletion, or ship a red test.

IvanWng97:
## Round-2 disposition — head `db5e3615`. The two-round cap is now CONSUMED.

Delta lens (fold-only population, `d7a98895..a68b0850`): 9 findings, all against round 1's fold, all now terminal:

**FIXED in `db5e3615`** (every fix a revert, a deletion, or shipped with a demonstrated red test):
- F1 doc-block hijack on the part-status consts → moved back (revert).
- F2 tautological pin → deleted; `DECODED_PART_STATUSES` now routes through `assert_arms_match_export` (an un-exported new arm reds — `STATUS_QUEUED` probe verified both directions).
- F3 SKILL.md "fail nothing on their own" → refuted by my own gates, reverted to "every one test-gated".
- F4 `"error"` vanish probe masked by two `Omit<…>` sites → declaration-form needle (`Schema.Literal`, 1:1:1 on the real documents) + case render.
- F5–F9 text corrections (drift.rs counter-example, copilot retain attribution, pointer ambiguity, precedent citation, last spelling).

**Severity adjudication note:** the lens labeled F1/F2 HIGH; graded MEDIUM here by the review's own precedent (identical missing-teeth-not-wrong-behavior patterns were MEDIUM in round 1's correctness reports; the contract makes the orchestrator's grading binding). Both were fixed regardless — the grading mattered only for the revert-the-fold trigger, which requires a confirmed HIGH.

**Contract state:** two fix rounds consumed. `db5e3615` is the final behavioral commit; anything further found at this head is disposition-only (fix-forward is out; revert or merge). Gates at this head: preflight exit 0 / 3,156 tests; arm census 27/27 null-controlled; live baseline clean. Awaiting CI + bot at `db5e3615`, then owner merge.

IvanWng97:
## Bot HIGH at `db5e3615` — adjudicated: **REFUTED-with-trace**

**The finding's facts are right; its risk model describes a mechanism this branch closed structurally.**

The calibrated `PARSE_FLOORS = {"hermes": 30}` existed as a band-aid for #929's ACTUAL mechanism: a comment inside `VALID_HOOKS` containing a stray `}` truncated the parsed set SILENTLY, and a bare non-emptiness check believed the remnant. That mechanism cannot recur at this head, on three independently-verified grounds:

1. `python_set_literal` strips comments BEFORE the brace scan (`check_upstream_drift.py:1097`), pinned by the restored `test_every_block_reader_strips_comments_before_counting_braces` — and the parser-mutation census kills the `python_set_no_comment_strip` mutant (measured this round, null-controlled).
2. A DUPLICATED declaration (the other truncation-adjacent shape) now returns `None` → probe health, not a first-match read (`src.count(decl) != 1`, negative-controlled with a poisoned twin).
3. A fetch either completes or files transient/blind — there is no partial-fetch path to a shrunken set.

What the calibrated 30 would cost at this head: upstream `VALID_HOOKS` is 37 today; a LEGITIMATE upstream reduction to, say, 20 hooks that renames one of our 6 would file **blind** under floor-30 — suppressing exactly the true drift this watch exists for. The derived floor (`len(handled)` = 6) alarms correctly there, and the truncation scenario it "loosens" is the one items 1–3 make unreachable. A hand-measured 30 is also itself the hand-calibrated-mirror class this PR removes (#938).

Residual honestly stated: if a FUTURE parser regression reintroduced silent truncation, floor-6 catches less than floor-30 — but that regression class is now covered by the parser census (5/5 mutants killed), which is a stronger guard than a constant.

Per the convergence contract the two-fix-round cap is consumed; this disposition is the terminal state. The docstring on `parse_is_believable` could carry one sentence of this trace — noted for a follow-up doc pass, not this PR.

Gate summary at `db5e3615`: CI 50/50 SUCCESS · local lenses two rounds complete, all 56 findings terminal · `mergeStateStatus: CLEAN`. Ready for owner merge.

PR #945, opened 2026-08-19
tests(#936): one fixture line-reader for the sources binary

### LINKED ISSUE #936: tests: four copies of the fixture line-reader, already out of lockstep — consolidate into captures.rs
Four copies of "read a fixture's JSONL lines" live in the source test tree:

| file:line | signature |
| --- | --- |
| `tests/sources/cursor/mod.rs:15` | `fn lines() -> Vec<String>` |
| `tests/sources/delegation/mod.rs:12` | `fn lines(cli: &str, name: &str) -> Vec<String>` |
| `tests/sources/grok/mod.rs:21` | `fn lines(name: &str) -> Vec<String>` |
| `tests/sources/conformance.rs:41` | `fn read_lines(path: &Path) -> Vec<String>` |

Three were added by #929; `conformance.rs`'s predates it.

**They are already out of lockstep.** `conformance.rs`'s orders `.map` before
`.filter`; the three new ones order `.filter` before `.map`. Functionally identical
today, but nothing pins them together — which is the point. A change to line-reading
semantics (trimming a trailing `\r` for a Windows checkout, skipping a UTF-8 BOM)
applied to one copy and missed in the others would leave that source's tests parsing
fixture lines differently from its siblings, silently.

**Why this is worth recording rather than shrugging at:** #929's own headline
refactor consolidated the sibling helpers `fixtures_root` / `sorted_dirs` /
`transcripts_in` out of `conformance.rs` into `captures.rs`, with a comment naming
exactly this failure class — *"This file kept byte-identical copies whose `read_dir`
SWALLOWED errors where the other panics"*. The line-reader sat right next to them and
was left duplicated four ways in the same diff.

**Fix:** one helper in `captures.rs` beside the other consolidated tree helpers, and
the four call sites pointed at it.

**Deliberately not fixed in #929.** Test-only, behaviourally identical today, zero
user impact — and that PR's review loop had reached the point where its own fixes
were the main source of new defects (5 of 9 real defects across rounds 6-7 were
self-inflicted). Batching one more four-file refactor into it was the wrong trade.

Found by the automated PR reviewer at `00f9b88b`, verified by hand.

IvanWng97:
## Summary

Seven copies of "read a fixture's JSONL lines" → one `fixture_lines` in `captures.rs`, beside the tree helpers #929 already consolidated.

The issue named four; the class sweep found three more in the same binary (claude / codex / codewhale, inline ahead of their per-source decode) — and review found the final two hiding in the helper's own file, which the sweep's `grep -v captures.rs` had excluded. **Nine copies total, all routed through the one helper.** The map/filter order divergence the issue flagged has no semantic gap, so this is pure copy deletion, zero behavior.

Deliberately out of the class: `wire_to_pixels.rs` (other crate, cannot reach core's test-internal module), `proof_fixture_disjointness.rs` (its own test binary), the two examples. A shared test-util crate for one 8-line fn is the wrong trade.

Population census (the deletion-PR rule), re-measured at the fold head: the pattern has **5 hits** over `crates/` = the helper itself + the four adjudicated-out sites; base had 13. Wrappers stay (they name which fixture a test pins); paths derive from `captures::sources_root()`.

Gates: `just preflight` exit 0, 3,156 tests.

Closes #936

https://claude.ai/code/session_017ZsTZvLuSjSVJULz4VDPrS

### DIFF
--- a/crates/pixtuoid-core/tests/sources/captures.rs
+++ b/crates/pixtuoid-core/tests/sources/captures.rs
@@ -91,6 +91,19 @@ pub(crate) fn transcripts_in(dir: &Path) -> Vec<PathBuf> {
     out
 }
 
+/// A fixture's JSONL payload as its non-empty lines — THE line reader for every
+/// source test. One definition so a semantics change (a BOM skip, a `\r` trim
+/// for Windows checkouts) cannot reach one source's tests and miss its siblings;
+/// four diverging copies is how #936 happened.
+pub(crate) fn fixture_lines(path: &Path) -> Vec<String> {
+    std::fs::read_to_string(path)
+        .unwrap_or_else(|e| panic!("read {}: {e}", path.display()))
+        .lines()
+        .filter(|l| !l.trim().is_empty())
+        .map(str::to_string)
+        .collect()
+}
+
 pub(crate) fn sources_root() -> PathBuf {
     Path::new(env!("CARGO_MANIFEST_DIR")).join("tests/sources")
 }
--- a/crates/pixtuoid-core/tests/sources/captures.rs
+++ b/crates/pixtuoid-core/tests/sources/captures.rs
@@ -91,10 +91,9 @@ pub(crate) fn transcripts_in(dir: &Path) -> Vec<PathBuf> {
     out
 }
 
-/// A fixture's JSONL payload as its non-empty lines — THE line reader for every
-/// source test. One definition so a semantics change (a BOM skip, a `\r` trim
-/// for Windows checkouts) cannot reach one source's tests and miss its siblings;
-/// four diverging copies is how #936 happened.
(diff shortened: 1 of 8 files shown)


IvanWng97:
**Round-1 disposition** — head `bf6a84c8`. Correctness (APPROVE-WITH-NITS, 2): F1 in-file copies + census arithmetic → FIXED in the fold; F2 inert `\r` example → FIXED. Design (APPROVE-WITH-NITS, 6): 1 (same as F1) → FIXED; 2 doc count/superlative → FIXED; 3 sources_root reuse → FIXED; 4 exclusion phrasing → FIXED in the PR body; 5 naming → DROPPED (taste); 6 docs-currency touches → DROPPED (optional, taste). Zero HIGH anywhere. Awaiting CI + bot at this head.

PR #952, opened 2026-08-20
fix: restore the star history chart

FaintFlower:
The star history chart in the README is currently broken. Update its links so the project history renders again while preserving the existing chart options.

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,11 +163,11 @@ Inspired by [`pixel-agents`](https://github.com/pablodelucca/pixel-agents) (VS C
 ## Star History
 
 <p align="center">
-  <a href="https://www.star-history.com/?repos=IvanWng97%2Fpixtuoid&type=date&legend=top-left">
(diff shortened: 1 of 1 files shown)


IvanWng97:
Thanks for jumping on the broken chart — appreciated.

Digging into it, though, the root cause isn't something any hosted service can route around: GitHub restricted the stargazers API (and the `/stargazers` view) to a repo's own admins and collaborators — [changelog 2026-06-30](https://github.blog/changelog/2026-06-30-upcoming-access-restrictions-to-public-api-endpoints-and-ui-views/). That's why star-history.com blanked, and it applies equally to any mirror: the replacement host's chart for this repo already stops at Aug 02 (we're at 452 stars now), because it can't read the data either — it can only serve whatever it captured before the cut-over, going stale from there.

So we're going with #953 instead: the repo draws its own chart in CI with `github.token` (the one identity that can still read the list), publishes it to an orphan `star-history` branch, and the README embeds that. No third-party data path, updates on every star.

Closing in favor of #953 — thanks again for the PR.

PR #956, opened 2026-08-28
refactor: delete the guide siblings, their generator, and the comment gate (+ 13 files of extraction)

IvanWng97:
Cuts the repo's comment and guide-prose surface, and removes the machinery that
was policing it. Net **−3282 lines** across 95 files; 3172 tests pass.

## What went, and why

**The comment-lint gate (−1015).** `just lint`'s prose arm, ci-lint's blocking
`comment gate` job, four recipes, the ast-grep rules and their advisory
workflow. The gate direction was called off; comment volume is cut by hand
instead.

**The `SHARP-EDGES.md` / `WHERE-TO-LOOK.md` siblings (−800 with their four
generated index blocks).** No entry was migrated into code, and that is the
finding rather than a shortcut: every declaration read in FULL already carried
its own edge verbatim — `may_probe_version`, `per_floor_counts`, `OutcomeRow`,
`is_first_run`, `DriftScan`, `CURSOR_EVENTS`, `FloorCapacitySweep`,
`resolve_theme`, `PetEntry.kind`, `build_floor_scene`, `config::load`. The tui
migration (9f8c7ac8, already on this branch) measured the same ratio: 3 of 7
entries were a second copy.

The siblings existed because a reader could not see what the declaration
already said. Migrating them back re-runs that mechanism — this branch did
exactly that for 26 lines before reverting them, having judged "missing" off a
5-line grep window that cut the doc comment in half. Nothing is lost that git
does not hold: `git log -- <path>` retrieves any entry.

**`gen-guides.py` + `gitenv.py` (−939).** The projection half of `gen-guides`
dies with the siblings; the byte-BUDGETS half is dropped by decision — the
owner reviews every doc change directly, so a generator deciding what belongs
in a guide is the wrong shape. `gitenv` existed because a git hook's exported
`GIT_DIR`/`GIT_INDEX_FILE` outrank `git -C <dir>`, so a script driving a
throwaway repo silently drove the real one (#893, twice, while printing a clean
pass). `gen-guides`'s `git ls-files` was its last caller, so nothing routed through
`gitenv.git()` survives.

**Correction (review round 1):** an earlier draft of this said "nothing in
`scripts/` spawns git any more, verified by grep." That is false — the grep was
`--include='*.py'` while the claim was about `scripts/`.
`scripts/lib/tier-live-sources.sh:78-80` runs `git init` / `git -C add -A` /
`commit` on a throwaway repo with all three stderr-suppressed: the #893
signature exactly. It is NOT newly exposed by this PR — `bypass_sweep` was
Python-only by explicit decision and never globbed `scripts/lib/` — and
reaching it needs an ambient `GIT_DIR`/`GIT_INDEX_FILE`, which `preflight`
(lint/clippy/hack/test, no e2e tiers) and every workflow avoid; the live-source
tier is manual and billed. SURFACED for the owner, not fixed here.

## The comment cut itself

Nested `CLAUDE.md` now says what its crate IS and stops: core 9782 → 1021,
scene 9674 → 2856 (the screen-space compass stays — it IS a convention),
pixtuoid 4599 → 904, tui 2977 → 951 bytes.

In code, 21405 → 20002 comment lines (13.3% → 12.6% of source). The kept/cut
line is **"delete this and someone confidently makes a wrong change"**, not a
length. Kept: the counterfactual that was tried and failed, the upstream fact
the code answers but cannot observe, why this and not the obvious neighbour,
the coupling nothing enforces, and where the invariant is pinned. Cut: setup
phrases, the derivation of a conclusion the next sentence states, the history
of what a shape replaced, and manifesto.

The largest single cut is `source/drift.rs`, 68 lines → 26: a four-item
taxonomy of defenses ranked by whether they can lie, a "silent in three ways"
section, two aphorisms, and per-issue status. What survives is the rule, the
counterfactual that produced it, and why the free-form values are made
display-safe at emission.

`hook_cmd`'s OS fork was restated above four three-line functions whose bodies
are a call to the authority — and the authority's own member list had already
rotted ("Codex and Reasonix" while cursor and kimi route through it too). All
four now name it.

## Coverage

Every `//`-block over 2 lines and every `///` run of 5+ was read: **1351 blocks
across 256 files**, plus three tree-wide mechanical scans over all 4815 comment
blocks (two rhetorical-shape passes, one 7-gram near-duplicate pass — the last
is what surfaced the `hook_cmd` restatements). ~1300 blocks were left alone
because cutting them would delete a billed capture's upstream fact, a failure
that was tried, or a coupling the compiler does not check.

## Notes for the reviewer

- Branch `refactor/sharp-edges-to-code` is superseded: its 14 non-gate commits
  are cherry-picked here, its 3 `fuck-ai-comments` commits deliberately are not.
- `main.rego`'s `ci_advisory_job_keys` still reserves `supplemental`; nothing
  calls it now that ci-supplemental.yml is gone, and a future advisory group
  would fail loudly rather than silently if the reservation were dropped.
- `just links` 98 unique / 0 errors; `just lint`, clippy, `cargo hack` and the
  full suite green.


IvanWng97:
## Correction to the description above — this is NOT a comment-only PR

Self-review before the gate found the body mischaracterizes the change. **13 `.rs` files carry 1265 lines of non-comment change** — function extractions that hoist inline blocks into named fns so a long rationale can live on a declaration:

| lines | file |
|---|---|
| 220 | `crates/pixtuoid/src/tui/mod.rs` |
| 184 | `crates/pixtuoid-core/src/source/decoder.rs` |
| 166 | `crates/pixtuoid/src/doctor.rs` |
| 140 | `crates/pixtuoid-core/src/state/reducer/mod.rs` |
| 124 | `crates/pixtuoid/src/install/mod.rs` |
| 118 | `crates/pixtuoid-core/src/source/claude_code.rs` |
| 114 | `crates/pixtuoid-scene/src/cutaway/paint.rs` |
| 97 | `crates/pixtuoid-scene/src/layout/placement_sweep.rs` |
| 51 | `crates/pixtuoid-scene/src/layout/compute.rs` |
| 21 | `crates/pixtuoid-core/tests/sources/captures.rs` |
| 11 | `crates/pixtuoid-scene/src/pixel_painter/tests.rs` |
| 10 | `crates/pixtuoid/src/audio/mod.rs` |
| 9 | `crates/pixtuoid-scene/src/layout/decor.rs` |

Example: on main the attribution/attachment logic is inline in `decode_cc_line`; on this branch it is `attribution_label()` + `attachment_effort()`.

The direction follows the repo's own convention (a rationale too long for a fn body MOVES onto a declaration — which requires a declaration to exist). But two facts matter for the gate:

1. These arrived by cherry-pick from **`refactor/sharp-edges-to-code`, which was never reviewed** — this PR is their first review.
2. The 256-file sweep this branch ran read **comments**, not code. It did not verify these extractions are behavior-preserving.

`two-lens-review`'s convergence contract says a change that both adds and deletes at scale is two PRs. Non-comment added lines are 976 — under the ~1500 churn budget — so the budget alone does not force a split, but the framing did need correcting before review.

Review lens 1 is verifying the 1265 lines for behavior equivalence; lenses 2 and 3 cover the tooling removal and the reshaped comments. Dispositions follow in this thread.

IvanWng97:
## Review round 3 — CodeQL's 25 HIGH alerts: REFUTED-with-trace

CI shows `CodeQL / 25 new alerts including 25 high severity security vulnerabilities`. Verified all 25, not a sample.

**Every one is `rust/cleartext-logging` on an `assert!`/`panic!` message inside `#[cfg(test)]` code.** 25 of 25 in test code, 0 on a production path. The traced "sensitive source" is the test's own hardcoded `serde_json::json!({...})` fixture — sensitive only because a field is spelled `session_id` — and the "log sink" is a panic message:

```rust
assert_eq!(evs.len(), 1, "expected exactly one event, got {evs:?}");
```

**Why they appeared now.** `main`'s full-tree scan carries 1 open alert of this rule (`cc_probe.rs`); this branch shows 25. The delta is not new code — it is the extraction refactor in `decoder.rs`. Inlined, CodeQL could not connect `hook_session_id(…)` through `decode_hook_payload` to the test's `{evs:?}`; split into `hook_source` / `decodes_to_nothing` / `source_owned_arms` / `shared_hook_arms`, the dataflow became traceable and a latent class surfaced. Three of the flagged files — including `tests/sources/decode/mod.rs` — this branch never touched.

**Not a real finding**: no production path, no user data, and the repo's actual defense for this class (`decoder::display_safe` sanitising untrusted wire values before `tracing`, because the non-TUI stream writes to raw stderr) is untouched and still covers every production sink.

**Not mechanically blocking**: `ci-gate` is the only context in branch protection (`policy/ci-observability/main.rego`), and CodeQL is not in its `needs`. But leaving a red CodeQL and 25 HIGHs is how the next reader learns to stop reading it.

**SURFACED, not fixed here.** The structural answer is a CodeQL config that does not treat a test assertion's format args as a logging sink — new config, which the convergence contract keeps out of a fix round and sends through the design gate as its own PR. Flagging rather than patching.

Round 3 also re-reviewed the delta since round 2 (`f7f5bba7`, two oversized module headers, doc-only): no findings. Round 1 and round 2 dispositions all still hold.

IvanWng97:
## Round 4 — the fallback lens, and its disposition sweep

Both bots are unusable at `75eb4077`. The review bot's own body reads *"Review incomplete: only 1 of 6 parallel scoped sub-reviews finished"*, and the security workflow posted its own absent-marker (*"total_cost_usd of 0 … a spent quota"*). Under the review skill that makes the gate unsatisfiable as written; the documented fallback is **one extra differentiated lens + owner merge, recorded here**.

One thing in the PR's favour that the check table hides: the security review at `f7f5bba7` **did complete, Findings: 0**. The only delta since is comment-only — verified, every changed line in `f7f5bba7..75eb4077` is a comment line. So the security lens covered this PR apart from 12 lines of prose.

**The lens run:** *"Substantial new/reshaped comments"* — read each added comment against the code it sits on. It is the escalation trigger this diff most obviously matches (1699 added comment lines / 108 files) and nothing had run it: round 1's local lens was behaviour-preservation, and the bot reached only the hook-shim seam.

### Findings and terminal states

| # | Where | State | Evidence |
|---|---|---|---|
| F1 | `daemon.rs:265` | **FIXED** | `"— see"` lost its object (`SHARP-EDGES.md`) and ran into `"Pinned by"`. Broken sentence. |
| F2 | `openclaw.rs:108` | **FIXED** | `Pinned by gateway_start_decodes_to_gateway_up_with_pid` — that test asserts the decode mapping only; nothing about announce-only, once-per-process, or the no-poll decision. Citation dropped; the upstream fact stands alone under the repo's own rule. |
| F3 | `reducer/mod.rs:586` | **FIXED** | Was `"No **in-tree JSONL** path emits Identity today … **not dead code**"`, trimmed to `"No path emits this today"`. That is **false** — `decoder.rs:429/445/460/469` each emit `hook_identity` — and it tells a reader the guard is dead. Scoping restored. |
| F4/F7 | `pinned_by_claims.rs` | **SURFACED as documented limits** | Collapse anchors at column 0 (an indented wrap is skipped; `install/mod.rs:144` is that shape), and `declared` is a raw text scan (18 tree-wide names are "declared" only by prose). Measured: **0 of 26 claims exploit either**. Mechanism untouched — the round cap forbids patching patches — but the module doc now states both. |
| F5 | `pinned_by_claims.rs` | **FIXED** | The doc claimed *"runs tree-wide"* and credited itself with a catch in a **markdown** file — outside its own population. It is `.rs`-under-`crates/`; `desk.sprite:17` and `add-source.prompt.md:19` carry unchecked claims (both resolve today). |
| F6 | `state/mod.rs:715` | **FIXED** | Called a green-tested deliberate behaviour a defect and prescribed `DeskPlacement`, a type that exists nowhere. `build_floor_scene_skips_agent_below_grown_offset` asserts *"skipped, not mapped to desk 0"*. Now cites it, and qualifies the two `pixtuoid_scene` symbols it names. |
| — | `wall.rs:49` | **FIXED (self-refuted)** | My own claim. *"Diverge at every load-bearing point"* is **false** — two of the five ladder alphas are identical (mullion 0.8, last 0.72). Replaced with the three real divergences: the H-only `GLASS_CAP_PX` cap, opposite run/anchor argument orders, three differing alphas. |
| — | `install/io.rs:329` | **FIXED (deleted)** | `sibling` at `io.rs:189` already carries the append-vs-`with_extension` fact, with the worked example (`config.local.toml` → `config.lock`) mine lacked. Mine was a weaker restatement two delegations away. |

`fresh_spend`, `attribution_label`, `decodes_to_nothing`, `source_owned_arms`, `late_child_start_refused`, `resolve_parent_link`, `enrich_existing_start` and the rest — **15 extractions checked adversarially-toward-revert; none moved semantics** (guard order, early-return scope, borrow width, short-circuit).

### Dangling-reference sweep

`rg --hidden` for all nine deleted artifacts (`gen-guides.py`, `gitenv.py`, `comment-lint.py`, `.ast-grep`, `sgconfig.yml`, `ci-supplemental.yml`, `SHARP-EDGES.md`, `WHERE-TO-LOOK.md`, `bypass_sweep`) across the whole tree returns **exactly one hit**: `main_test.rego:1004`, a hand-written fixture that models a hypothetical `ci.yml` and reads nothing from disk. **Inert.**

### SURFACED — owner's call, not changed here

**`main.rego:674` — `ci_advisory_job_keys := {"supplemental"}`.** Earlier notes in this thread called it an orphaned key; that understates it. It is a **live mechanism** — advisory-group exclusion — whose only user this PR deleted, and `main_test.rego`'s fixture depends on it (`test_ci_gate_covering_every_group_is_accepted` passes *because* the key excludes `supplemental`). Emptying it deletes the mechanism and rewrites its test — bigger than this PR. Residual risk: a future job named `supplemental` is silently excluded from the merge gate.

Also still open from earlier rounds: the CodeQL config (all 25 HIGH are `rust/cleartext-logging` on `#[cfg(test)]` assert messages — REFUTED-with-trace, and CodeQL is not in `ci-gate`'s `needs`), and `site/SINGLE-SOURCED.md`.

### Method defect worth recording

The `io.rs` deletion is not a slip. I added that line in round 3 claiming a census proved the fact was gone. **It was not gone** — it was on `sibling`, where it belongs. The census asked whether a *declaration* carried the fact; it should have asked whether the *tree* still carried it somewhere load-bearing. Same shape as the earlier `grep -B 5`-as-a-full-declaration error.

### Gate state

`ci-gate` **SUCCESS** (the only context in branch protection) · `mergeable: MERGEABLE` · preflight green, 3174 tests · `doc-check` exit 0 · `fmt` clean.

Zero open confirmed HIGH. **A human merges.**

IvanWng97:
### Follow-up: what the two deleted gates enforced, and whether anything still does

The disposition above covered the comments and the dangling-reference sweep. It did not answer the question a deletion PR most owes — **for each gate this PR removes, what did it enforce, and is that invariant now unenforced?** Answering it narrows one thing I had been overstating earlier in this PR's history.

**`comment-lint.py` / `.ast-grep/` / `ci-supplemental.yml`** — enforced the comment rules as a CI gate. Now unenforced, and that is the PR's explicit purpose, not a regression.

**`gitenv.py`** — enforced that every git spawn under `scripts/**/*.py` went through an env scrub, because a git hook's exported `GIT_DIR` / `GIT_INDEX_FILE` **outrank `git -C <dir>`**: a script driving a throwaway repo silently drives the real one (#893, twice, while printing a clean pass). Its `bypass_sweep()` was an AST walk that failed the build on unrouted git argv.

Traced end to end:

- **Git spawns left in `scripts/`: three**, all in `scripts/lib/tier-live-sources.sh:78-80`. Nothing else in `scripts/` spawns git.
- **Nothing in the tree scrubs** — no `git rev-parse --local-env-vars`, no `GIT_DIR`/`GIT_INDEX_FILE` handling anywhere.
- **`.githooks/pre-commit` → `just fmt-check`. `.githooks/pre-push` → `just preflight`.**
- **`preflight: lint clippy hack test`** — does not include `live-sources`.
- **`tier-live-sources.sh` has exactly one caller**: `just live-sources` (justfile:673), a manual, billed, deliberately-invoked recipe.

**So the invariant is not unenforced in the hook path — it is unreachable there.** No hook route reaches those three git calls, so the #893 mechanism has no live trigger today.

The residual is real but narrower than I had been implying: if `just live-sources` is ever run from *inside* a git-spawned process — a future hook that calls it, or `git rebase --exec 'just live-sources'` — those three calls target the real repo instead of `$WS`. `tier-live-sources.sh:79` is `git -C "$WS" add -A`, which is the destructive one.

That is a real hazard with no live trigger, so it is **SURFACED**, not a blocker for this PR. A fix is already drafted on branch `fix/e2e-env-scrub` (an `e2e_init_repo` helper in `scripts/lib/e2e-common.sh` that scrubs via git's own `rev-parse --local-env-vars` rather than a `GIT_*` prefix match — the prefix form would also strip `GIT_SSH_COMMAND` and `GIT_COMMITTER_*`), with a negative-control selftest that runs the naive form first to prove it corrupts. It is uncommitted and belongs in its own PR, not this one.

PR #957, opened 2026-08-28
docs: the three operational guides drop what the code already states

IvanWng97:
Follow-on to #956. That PR moved the per-crate sharp edges onto their declarations and deleted the `SHARP-EDGES.md` / `WHERE-TO-LOOK.md` siblings. The three guides that carry **operational** content — `tests/`, Raycast, site — kept theirs inline and were out of its scope. This is that scope.

**−97 / +58 across four guides**, plus two files that gained a header they never had.

## The method, and where it caught me

Every deleted clause was checked by **opening the declaration it claims to live on**, not by grepping for the symbol. That distinction matters: the same shortcut produced a wrong deletion in #956 (a `backup_once` doc that duplicated `sibling`'s, which I only found in round 4).

Three clauses turned out **not** to be duplicates. They are handled, not dropped:

| Clause | Why it survived |
|---|---|
| **The multi-file test binary rule** — `<area>/main.rs`, never `<area>.rs`, because a top-level `<area>.rs` is a crate root whose `mod foo;` resolves to a **sibling** `tests/foo.rs` | Nothing in the tree stated it, and no declaration can own a fact about **directory layout**. Kept in the guide, folded onto the line it qualifies. |
| **The Windows parity twins** | `transport/pipe.rs` and `shim_pipe.rs` each opened with a bare `#![cfg(windows)]` and **no header at all**. Each now names its twin. |
| **The outstanding Raycast republish** | Gone entirely. Restored, and split by what is checkable — see below. |

## The Raycast republish, split by verifiability

The old text asserted a store state. Verified from git: the wire split (`e21ec7f0`, 2026-07-02) **is** a descendant of the last publish marker (`b870d8ba`, 2026-06-19), so the version built for the store predates the split and still prefix-strips `failed: <msg>`.

Not verifiable from this repo: whether a republish has since cleared it. The guide now says so explicitly rather than asserting a state I cannot check — per `CLAUDE.md`'s "external-surface claims are fetched, not remembered".

## Verified as genuinely duplicated (destination read, not grepped)

- `assert-docs-rendered.mjs`'s header owns the silent-empty-render class
- `astro.config.mjs`'s `cspInlineHashes()` owns all three CSP mechanisms
- **`Docs.astro:296` carries the callout specificity arithmetic in more detail than the guide did** — the (0,3,1) vs (0,1,1) weights, the theme-prefixed exact tie that only source order breaks, and the (0,4,2) twin. The guide only gestured at it.
- `runPixtuoid`'s comment owns the `err.stdout` recovery
- `resolveBinary`'s doc owns the re-read-every-call rule
- `contract-outcome.ts` owns the PUBLISHED-wire rule

## Additions

- `Cargo.toml`'s `exclude` gains the clause explaining why the four tests stay **flat** (a grouped binary's submodule cannot be excluded on its own).
- The Raycast guide gains the `npm audit` policy — it has nowhere else to go, since `package.json` cannot carry a comment, and the guide now says that.

## Why a fresh branch

Branched off `main`, not rebased from `refactor/guides-to-code`. That branch predates #956's round-3 and round-4 fixes, so landing it would have **reverted seven of them** — including the trim that had turned *"No in-tree JSONL path emits Identity today … not dead code"* into the false *"No path emits this today"*.

## Gates

`preflight` green (3174) · `just links` 98 unique / **0 errors** · `rustfmt --check` clean.

One gap named rather than hidden: the two edited `.rs` files are `#![cfg(windows)]`, so **preflight on darwin never compiles them**. Their syntax was checked separately (`rustfmt --check` parses both; a minimal `//!`-before-`#![cfg]` fixture compiles under `rustc`). CI's `windows-check` is the real gate.

---

## Census completion (round-1 review)

The list above enumerated 9 clause-groups; the diff deletes 11. The two the
body never named — both verified to survive, so neither is a lost fact:

| Clause | Where it survives |
|---|---|
| `conformance.rs`'s one-`AgentId` bullet | `conformance.rs:1-13` |
| the insta-naming bullet | `fixtures/README.md:147-158`; the concrete name form is retained at `tests/CLAUDE.md:28` |

A reviewer should be able to check a census, not reconstruct the population.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -9,11 +9,11 @@ in nested guides, auto-loaded when you touch their tree:
 - [`crates/pixtuoid/CLAUDE.md`](crates/pixtuoid/CLAUDE.md) — the binary: install, runtime, cli, config, multi-floor. (+ [`src/tui/CLAUDE.md`](crates/pixtuoid/src/tui/CLAUDE.md): the terminal painter.)
 - [`integrations/raycast/CLAUDE.md`](integrations/raycast/CLAUDE.md), [`site/CLAUDE.md`](site/CLAUDE.md) — the non-Rust `--json` consumers; their gates are `tsc`/`eslint` / `just site-check`, not cargo.
 
-**What the guides hold.** A nested `CLAUDE.md` says what its crate IS, plus
-the cross-file couplings no declaration can own (site's build inputs, Raycast's
-`package.json` facts). Everything else a change needs — the constraint that looks
-like a bug, the WHY, the test that pins it — is on the declaration it
-constrains: read the item's doc comment before changing it.
+**What the guides hold.** A nested `CLAUDE.md` says what its crate IS, plus the
+facts no declaration can own (site's build inputs, Raycast's `package.json`
+rules, `tests/`'s directory-to-binary layout). Everything else a change needs —
+the constraint that looks like a bug, the WHY, the test that pins it — is on
+the declaration it constrains: read the item's doc comment before changing it.
 
 **A third consumer lives outside this repo**: homebrew-core's `pixtuoid`
 formula asserts exact CLI output (`test do`) and needs `pixtuoid man` /
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -10,8 +10,8 @@ in nested guides, auto-loaded when you touch their tree:
 - [`integrations/raycast/CLAUDE.md`](integrations/raycast/CLAUDE.md), [`site/CLAUDE.md`](site/CLAUDE.md) — the non-Rust `--json` consumers; their gates are `tsc`/`eslint` / `just site-check`, not cargo.
 
 **What the guides hold.** A nested `CLAUDE.md` says what its crate IS, plus the
-facts no declaration can own (site's build inputs, Raycast's `package.json`
-rules, `tests/`'s directory-to-binary layout). Everything else a change needs —
+facts no single declaration can own: cross-file workflows, comment-less
+manifests, and directory-shape rules. Everything else a change needs —
 the constraint that looks like a bug, the WHY, the test that pins it — is on
 the declaration it constrains: read the item's doc comment before changing it.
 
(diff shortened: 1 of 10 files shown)


IvanWng97:
## Two-lens review — round 1 disposition (head `695f28de`)

Three differentiated lenses: correctness/grounding · design/blast-radius · comment-accuracy (escalation lens — "substantial new/reshaped comments").

### HIGH-1 — a second, lossier copy of the `npm audit` bullet — **FIXED**

Independently found by all three lenses. Base had **one** bullet (`:98`); this PR added a second at `:70-76` and left the original at `:88-99`. The added copy dropped `GHSA-mh99-v99m-4gvg`, the backport-to-2.1.3 correction, the `@oclif/core → ejs → jake → filelist → minimatch` chain, "same as site's" — and the surviving bullet's own warning that the vague reading *"has already misled one reviewer"*, then restored exactly that vague phrasing. A reader hit the lossy copy first.

Deleted the new copy; section retitled `## Toolchain policy`. Base has one bullet, head has one bullet — net zero change on that fact.

### Three facts deleted with no surviving home — **FIXED**

The lenses' deletion-side sweep (open the destination, don't grep the symbol) found these; the PR body never listed them.

| Lost | Restored |
|---|---|
| tsconfig clause after the colon: bundled-tsc stamp, the TS 6 migration (`moduleResolution: "Bundler"` + explicit `types: ["node"]`), and that **`ignoreDeprecations: "6.0"` would have broken `ray build`** | to the guide — no declaration can own a warning about an *absent* key, and no `.json` in this tree carries comments |
| the Raycast store URL — gone repo-wide (only `package.json`'s `$schema` matched) | to the republish paragraph |
| the `ray publish` marker's ref name | restored, and the paragraph now says the tag is local-only (`git ls-remote --tags origin` has none) instead of implying a fresh clone can check |

The `ignoreDeprecations` one is a live trap: it's the obvious move during a TS upgrade, and `ray build` is **not in CI**, so it would surface at publish time.

The CSP `verified vs 7.0.5` anchor moved onto the claim it dates (`astro.config.mjs`) rather than being dropped — `site/package.json` is already on `^7.2.1`, two minors past it, which is exactly the staleness signal the anchor exists to give.

### MEDIUM — twin headers named a job, and pointed one way — **FIXED**

`run only by the windows-test job` was a restated job name *and* wrong: `windows-check` (`ci-builds.yml:45`) and `coverage-windows` (`ci-tests.yml:103`) compile these files too. The real gate is `#![cfg(windows)]`. Headers now name the platform.

The pointer was also one-sided — only on the two `#![cfg(windows)]` files, which a macOS contributor never opens, while the person who needs it is editing `socket.rs`. Both unix halves now name their twins, and the sentence is **imperative**, because the rosters have already drifted (socket 14 vs pipe 10, shim 7 vs shim_pipe 6) — it was asserting a parity that does not hold.

*Provenance, stated honestly:* the `windows-test` wording is verbatim from base `tests/CLAUDE.md:99`, so the inaccuracy is pre-existing, not introduced. Fixed anyway because this PR authors the line and relocating known-wrong text onto a declaration — where "name the authority, never restate its value" bites harder — makes it worse.

### MEDIUM — root map's parenthetical read as exhaustive — **FIXED**

It named three instances while both guides still hold much more; an agent applying it literally would delete two-thirds of each. Now names the category.

### Census — **FIXED** (PR body)

Body enumerated 9 clause-groups, diff deletes 11. The two unnamed (`conformance.rs`'s one-`AgentId` bullet, the insta-naming bullet) both survive; added to the body so a reviewer checks a census instead of reconstructing the population.

### Nits — **FIXED**
Blank line before `## Dev server`; `Cargo.toml` exclude comment's third restatement of one failure mode cut and its misleading "for the same reason" corrected; root guide paragraph rewrapped.

### REFUTED
`//!` before `#![cfg(windows)]` is legal — both are inner attributes. Proved, not reasoned: two fixtures through `rustc --emit=metadata`, both orders exit 0.

### Verified clean (not findings)
All six claimed relocation destinations genuinely own their facts; several state them in *more* detail than the guide did (`Docs.astro`'s specificity arithmetic). The republish ancestry claim holds (`merge-base --is-ancestor b870d8ba e21ec7f0` → 0), and the marker hasn't moved since (`e21ec7f0` is not an ancestor of the tag → exit 1). The multi-file test-binary rule is true and correctly kept in the guide. Fresh-branch-off-main was the right call — rebasing would have reverted seven of #956's later fixes.

**Gates:** `just lint` 0 (18/18) · `cargo fmt --check` 0 · both unix test bins compile · full `preflight` on push, 3174 passed.

PR #958, opened 2026-08-28
fix(e2e): scrub git's repo-relocating env before the throwaway workspace repo

IvanWng97:
Restores the guard #956 deleted, as a mechanism instead of a script.

## The hazard

`scripts/lib/tier-live-sources.sh` built its throwaway workspace with a bare `git init` / `git -C "$WS" add -A` / `commit`. A git hook exports `GIT_DIR` and `GIT_INDEX_FILE` into every child it spawns, and **those outrank `git -C <dir>`** — so run from a hook, that sequence does not touch `$WS` at all. It commits the developer's own tree into the developer's own repo. Exit 0, no output.

That is #893, which landed **twice** while printing a clean pass. `scripts/gitenv.py` held the guard until #956 deleted it along with the generator it served.

Measured, not reasoned — with `GIT_DIR` exported:

```
git init -q "$ws"      → exit 0, and $ws/.git is NOT created (it re-inits the real repo)
git -C "$ws" add -A    → exit 0, and stages the REAL repo's files
```

## No live trigger, and the PR says so

`.githooks/pre-push` runs `just preflight`; `preflight` is `lint clippy hack test`; `tier-live-sources.sh`'s only caller is the manual, billed `just live-sources` (justfile:673). **Nothing reaches those git calls from a hook today.** This keeps that true for the next caller rather than depending on it.

## The scrub

`e2e_init_repo` scrubs via `git rev-parse --local-env-vars` — git's own answer, 15 vars, fetched in-session rather than remembered. Deliberately **not** a `GIT_*` prefix match: that would also strip `GIT_SSH_COMMAND` and the `GIT_COMMITTER_*` identity, neither of which relocates a repo. The selftest pins that distinction.

## The selftest, and the hole mutation testing found in it

It runs the **naive form first and proves it corrupts**, so a scrub that stopped scrubbing cannot pass vacuously.

Two rounds of mutation testing (empty the scrub, the suite must red). The second round found a real hole in my own first version:

| | |
|---|---|
| **round 1** | keyed the safety assertion on the real repo's **index**. Under the mutation it **PASSED** — because the unscrubbed helper's third command is `commit`, which lands the stolen files as a real commit and leaves `diff --cached` clean again. The index is not the damage; the commit is. |
| **round 2** | assert on **HEAD**. The load-bearing check now reds under the mutation. The index check stays as belt-and-braces, with a comment saying it is *not* the one with teeth, and why. |

`GIT_WORK_TREE` is exported alongside the other two: without it an unscrubbed `add -A` resolves its work tree from the CWD, so what it stages depends on where the suite was invoked and the assertion passes or fails by accident.

Hermetic — one `mktemp -d`, no network, the real repo is never a target.

## Wiring

Both places, because one alone is a no-op:

- `just lint`'s fan-out (`✓ e2escrub`)
- **`ci-lint.yml`'s hygiene job** — that job enumerates each selftest as its own step, so the recipe alone would have run locally and never in CI.

## Gates

`preflight` green (3174) · `shellcheck -x` exit 0 · `shfmt-check` clean · `just lint` green · `actionlint` clean · **116 Rego CI-contract tests pass** on the workflow edit.

## Note

Supersedes the stale `fix/e2e-env-scrub` branch (uncommitted work in another worktree). Branched fresh off `main`.

### DIFF
--- a/.github/workflows/ci-lint.yml
+++ b/.github/workflows/ci-lint.yml
@@ -126,6 +126,7 @@ jobs:
       - run: just drift-selftest
       - run: just env-paths
       - run: just tuidrive-selftest
+      - run: just e2e-scrub-selftest
       - run: just star-history-selftest
       # `just fixture-metadata` is NOT here: it became Rust tests in #929, so it
       # already rides `just test` in windows-test, macos-test and coverage —
--- a/justfile
+++ b/justfile
@@ -316,6 +316,7 @@ lint:
     run links   just links               & pids+=($!)
     run drift   just drift-selftest       & pids+=($!)
     run tuidrive just tuidrive-selftest   & pids+=($!)
+    run e2escrub just e2e-scrub-selftest  & pids+=($!)
     run starhist just star-history-selftest & pids+=($!)
     run fixpii  just fixture-pii          & pids+=($!)
     run piiself just fixture-pii-selftest & pids+=($!)
@@ -1279,6 +1280,16 @@ drift-selftest:
 [doc("Self-test the TUI capture driver's pure logic")]
 tuidrive-selftest:
     python3 scripts/lib/tuidrive.py --selftest
+
+# `e2e_init_repo`'s env scrub. A git hook exports GIT_DIR/GIT_INDEX_FILE into
+# every child and those OUTRANK `git -C <dir>`, so an unscrubbed helper COMMITS
+# to the developer's real repo while printing nothing (#893, twice). The suite
+# runs the naive form first and proves it corrupts, so a scrub that stopped
+# scrubbing cannot pass. Hermetic: one mktemp -d, no network, no real repo.
+[group('meta')]
+[doc("Self-test the e2e helper's git env scrub")]
+e2e-scrub-selftest:
+    bash scripts/lib/e2e-common-selftest.sh
 
 # The README star chart's renderer — bitmap font, axes, paging. (Its copied
 # office colours are pinned from Rust: scene's `tests/readme_chart_palette.rs`.)
--- a/scripts/lib/e2e-common-selftest.sh
+++ b/scripts/lib/e2e-common-selftest.sh
@@ -6,6 +6,14 @@
 # Everything happens inside one `mktemp -d`; the real repo is never a target.
 set -uo pipefail
 
+# A LINKED worktree's pre-push exports GIT_DIR (a plain checkout's does not), and
+# `just lint` reaches this file from pre-push — so without this the fixture setup
+# below builds no repo of its own and commits into the developer's: #893, inside
+# the file that pins #893. It has to happen HERE, before anything builds a repo;
+# the suite's own GIT_* exports come later and are unaffected.
+# shellcheck disable=SC2046  # githooks(5)'s own idiom — the list must word-split
+unset $(git rev-parse --local-env-vars)
+
 here="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 # shellcheck source=/dev/null
 . "$here/e2e-common.sh"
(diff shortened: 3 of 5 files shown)


IvanWng97:
## Two-lens review — round 1 disposition (head `f39435ed`)

Three differentiated lenses: correctness/grounding · design/blast-radius · shell-robustness + CI-wiring (escalation lens, because the payload is entirely shell + CI where the Rust gates are blind).

**All three landed the same HIGH, and it was reproduced end-to-end with the shipped file and no mutation.**

### HIGH-1 — the selftest committed into the developer's repo on every push from a linked worktree — **FIXED**

Two independently-benign facts composed into the exact #893 corruption, inside the file that exists to pin #893:

1. The suite built its victim stand-in with the naive form (`git init` / `add` / `commit`) at lines 34-37, which run **before** the suite exports its own `GIT_DIR` at line 44. `e2e_init_repo` was scrubbed; its fixture was not.
2. The linchpin assertion `test "$(head_of_real)" = "$head_before"` compares empty to empty when `$real/.git` was never created — so it **passes** in exactly the state a leaked `GIT_DIR` produces.

The trigger was live, not theoretical. Measured both ways: a **linked worktree's** `pre-push` exports `GIT_DIR`; a plain checkout's does not. Chain: `.githooks/pre-push:24` → `just preflight` → `lint` → `justfile:319` → this file. This repo mandates worktrees for concurrent sessions, so the blast radius was every push from one — strictly wider than the hazard the PR fixes, whose only caller is the manual, billed `just live-sources` (`justfile:675`).

CI could never have caught it: the runner uses a fresh plain checkout and invokes the recipe directly, never from a hook. That is verbatim the blindness #893's own body documents.

This is not hypothetical — it happened. It left `core.bare = true` in this repo's shared config and a staged `tracked.txt` (blob `keep`, i.e. line 35's `printf`), both cleaned up before this review.

**Fix:** clear git's own list at the top of the suite, before anything builds a repo — `unset $(git rev-parse --local-env-vars)`, the idiom `githooks(5)` prescribes verbatim (fetched in-session, git 2.54.0) — plus an assertion that the fixture repo actually has a HEAD.

### NIT-1 — `e2e_init_repo` failed OPEN — **FIXED**

If `git rev-parse --local-env-vars` ever failed, `$scrub` was empty and every `env $scrub git …` ran fully unscrubbed, silently. Now returns 1 with a message. Verified with a stub `git` on PATH.

### Verified clean (not findings)

- The scrub list is git's own 15 vars, correctly narrow: `GIT_SSH_COMMAND` and `GIT_COMMITTER_*` survive, pinned by the suite.
- The HEAD-keyed assertion genuinely has teeth: emptying the scrub still reds 3 checks. A partial scrub (`-u GIT_DIR` only) also reds.
- Both wirings exist and run (`justfile:319`, `ci-lint.yml:129`, no `continue-on-error`).
- `tier-live-sources.sh` is the only caller.

**Post-fix verification, four ways:** clean env green · scrub emptied still reds 3 (tooth preserved) · leaked `GIT_DIR` at a throwaway victim leaves it byte-identical while the suite stays green on genuine checks · stub `git` makes the new guard return 1.

**Gates:** `shellcheck -x` 0 · `just shfmt-check` 0 · `just actionlint` 0 · `just ci-observability` 0 (116 Rego tests) · `just e2e-scrub-selftest` 0 · full `preflight` on push, 3174 passed.

IvanWng97:
## Round 2 disposition (head `f1111445`) — last behavior-changing commit

Round 2 exists to verify round 1's dispositions and catch defects the fold itself introduced. It caught one, in my own fix.

### The round-1 fail-closed guard failed open — **FIXED**

Round 1 gave `e2e_init_repo` a `[ -z "$scrub" ] && return 1`. That is only a guard if a caller reads the status, and none does: `tier-live-sources.sh:13` is `set -uo pipefail` with **no `-e`**, and `:78` calls `e2e_init_repo "$WS"` bare. So round 1 printed a warning and then ran the unscrubbed git calls anyway — a guard that does not guard, which is precisely the class this PR removes.

Now `exit 2`, reusing this file's existing authority (`e2e_require_bin`, `e2e-common.sh:16`) rather than inventing a second convention.

Measured with a shadowed `git`, against a caller shaped like the real one:

| `git rev-parse --local-env-vars` | round 1 | round 2 |
|---|---|---|
| exits 129 | returns 0, caller continues, **victim corrupted** | caller **exit 2**, aborted before the next line |
| exits 0 with empty stdout | returns 0, caller continues, **victim corrupted** | caller **exit 2**, aborted |

Victim HEAD unmoved in both round-2 modes.

### Round-1 findings re-verified at this head
Selftest 7/7 green · emptying the scrub still reds 3 checks (tooth preserved) · leaked `GIT_DIR` leaves a throwaway victim byte-identical · `shellcheck -x` 0 · `just shfmt-check` 0 · `just e2e-scrub-selftest` 0 · full `preflight` on push, 3174 passed.

Every round-2 change is a one-line hardening with its failure mode measured both before and after. Zero open confirmed HIGH. Within the two-round cap — no round 3.

PR #960, opened 2026-08-31
fix(hooks): clear git's repo-relocating env once, at the pre-push entry

IvanWng97:
Follow-on to #958. That PR scrubbed git's repo-relocating env inside `e2e_init_repo`; this moves the same scrub to the one place the pollution comes from.

## Why the helper is not enough

#958 makes the correct thing easy. It does not make the incorrect thing impossible — any future script that writes a bare `git init` bypasses the helper entirely.

That is not hypothetical. **#958's own selftest did exactly that**: it built its fixture with the naive form, committed into the developer's repo, and printed `OK`. The author knew the trap, was actively writing the fix for it, and still bypassed it in the same file. A mechanism that the author bypasses at their most alert moment is not a mechanism.

## Why the hook entry is the right place

The pollution has exactly **one** source: a git hook. Measured both ways on git 2.54.0:

```
push from a plain checkout   → GIT_DIR=<unset>
push from a LINKED worktree  → GIT_DIR=…/main/.git/worktrees/<name>
```

and `GIT_DIR` outranks `git -C <dir>`. Chain: `.githooks/pre-push` → `just preflight` → `lint` → every selftest and script it fans out to.

Clearing it at the entry means every downstream child — preflight, lint, every selftest, **every script not yet written** — is immune without its author knowing the trap exists.

This is the shape [pre-commit](https://github.com/pre-commit/pre-commit/blob/main/pre_commit/git.py) uses: its `no_git_env()` filters the environment in the *framework*, so third-party hooks cannot get this wrong. Its comment is blunt — *"Too many bugs dealing with environment variables and GIT"*.

`githooks(5)` prescribes the idiom verbatim (fetched in-session from `man githooks`, git 2.54.0). Deliberately **not** a `GIT_*` prefix match — that would strip `GIT_SSH_COMMAND` and the `GIT_COMMITTER_*` identity, neither of which relocates a repo.

## Measured

Real hook, a probe standing in for preflight, pushing from a linked worktree:

| hook | real repo `core.bare` | |
|---|---|---|
| before this change | `true` | polluted |
| after | `false` | clean |

Side effects checked, not assumed: after the `unset`, `--show-toplevel`, `--abbrev-ref HEAD`, `--git-dir` and `status` all still resolve the worktree correctly — git falls back to the working directory it sets for the hook, exactly as `githooks(5)` describes.

Over-reach checked: the only `GIT_CONFIG_COUNT` user in this repo is `star-history.yml`, which runs on a fresh runner checkout with no `core.hooksPath`, so it never reaches this hook.

## No selftest, deliberately

**CI structurally cannot execute a hook test** — the runner uses a fresh plain checkout and invokes recipes directly, never from a hook and never from a linked worktree. That blindness is exactly why #893 landed twice and why #958's selftest bug survived 47 green checks.

A hook selftest would therefore be green locally and never run in CI — the trap #958 had to fix in its own wiring (`just lint` alone would have been a no-op). The primary defense, the helper's scrub, keeps its CI-covered selftest. This is depth.

## Note

`core.hooksPath` is an absolute path to the main checkout, so this takes effect for **every** worktree the moment it lands — and correspondingly it does not self-verify on its own push, which is why the before/after control above was run explicitly.

### DIFF
--- a/.githooks/pre-push
+++ b/.githooks/pre-push
@@ -8,6 +8,16 @@
 # Bypass with `--no-verify` or `SKIP_PREFLIGHT=1 git push`.
 set -euo pipefail
 
+# git exports GIT_DIR into every child of a hook, and from a LINKED worktree it
+# outranks `git -C <dir>` — so anything preflight runs that builds a throwaway
+# repo writes into THIS one instead, exit 0 and silent (#893, twice; then #958's
+# own selftest). Clearing it HERE means no downstream script has to remember.
+# githooks(5) prescribes this idiom for git commands aimed at a foreign
+# repository; the working directory git sets for a hook still resolves the
+# worktree, so nothing below loses its bearings.
+# shellcheck disable=SC2046  # the var list must word-split
+unset $(git rev-parse --local-env-vars)
+
 if [[ "${SKIP_PREFLIGHT:-0}" == "1" ]]; then
     printf '\033[33m[pre-push] SKIP_PREFLIGHT=1 — skipping preflight\033[0m\n' >&2
     exit 0
--- a/.githooks/pre-push
+++ b/.githooks/pre-push
@@ -8,22 +8,30 @@
 # Bypass with `--no-verify` or `SKIP_PREFLIGHT=1 git push`.
 set -euo pipefail
 
-# git exports GIT_DIR into every child of a hook, and from a LINKED worktree it
-# outranks `git -C <dir>` — so anything preflight runs that builds a throwaway
-# repo writes into THIS one instead, exit 0 and silent (#893, twice; then #958's
-# own selftest). Clearing it HERE means no downstream script has to remember.
-# githooks(5) prescribes this idiom for git commands aimed at a foreign
-# repository; the working directory git sets for a hook still resolves the
-# worktree, so nothing below loses its bearings.
-# shellcheck disable=SC2046  # the var list must word-split
-unset $(git rev-parse --local-env-vars)
-
+# Ahead of the scrub, so that bypass still covers everything below it.
 if [[ "${SKIP_PREFLIGHT:-0}" == "1" ]]; then
     printf '\033[33m[pre-push] SKIP_PREFLIGHT=1 — skipping preflight\033[0m\n' >&2
     exit 0
 fi
 
-cd "$(git rev-parse --show-toplevel)"
+# Resolved before the scrub drops GIT_DIR: when the hook's CWD is outside the
+# worktree (`git --git-dir=… push` from elsewhere) that is the only locator.
+toplevel=$(git rev-parse --show-toplevel)
+
+# A LINKED worktree's hook exports GIT_DIR to its children, and it outranks
+# `git -C <dir>`: anything preflight runs that builds a throwaway repo then
+# silently writes into THIS one (#893) — githooks(5)'s idiom, applied here
+# rather than per-script so nothing downstream has to remember it.
+# A failed or empty list would leave the scrub silently not happening, and
+# `unset $(...)` discards the substitution's status so `set -e` never sees it.
+if ! scrub=$(git rev-parse --local-env-vars) || [ -z "$scrub" ]; then
+    printf '\033[31m[pre-push] git gave no local-env-vars — refusing to run unscrubbed\033[0m\n' >&2
+    exit 1
+fi
+# shellcheck disable=SC2086  # the var list must word-split
+unset $scrub
+
+cd "$toplevel"
 
 if ! command -v just &>/dev/null; then
     # shellcheck disable=SC2016  # backticks are literal text, not expansion
--- a/justfile
+++ b/justfile
@@ -1281,13 +1281,14 @@ drift-selftest:
 tuidrive-selftest:
     python3 scripts/lib/tuidrive.py --selftest
 
-# `e2e_init_repo`'s env scrub. A git hook exports GIT_DIR/GIT_INDEX_FILE into
-# every child and those OUTRANK `git -C <dir>`, so an unscrubbed helper COMMITS
-# to the developer's real repo while printing nothing (#893, twice). The suite
-# runs the naive form first and proves it corrupts, so a scrub that stopped
-# scrubbing cannot pass. Hermetic: one mktemp -d, no network, no real repo.
+# The git env scrub, both copies: `e2e_init_repo` and the pre-push hook. A git
+# hook exports GIT_DIR/GIT_INDEX_FILE into every child and those OUTRANK
+# `git -C <dir>`, so an unscrubbed helper COMMITS to the developer's real repo
+# while printing nothing (#893, twice). The suite runs the unscrubbed form first
+# and proves it leaks, so a scrub that stopped scrubbing cannot pass. Hermetic:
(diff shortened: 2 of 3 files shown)


IvanWng97:
## Two-lens review — round 1 disposition (head `3e4b0d55`)

Three differentiated lenses: correctness/grounding · design/blast-radius · **adversarial** (dispatched because this file runs on every contributor's every push — its brief was literally "try to brick `git push`").

A 10-line change. The review found a defect in the one line that mattered, two regressions it introduced, and a missing test I had argued was impossible. All three were my calls.

### HIGH — the guard did not guard — **FIXED**

`unset $(git rev-parse --local-env-vars)` discards the substitution's status; only `unset`'s own counts, and bare `unset` returns 0. So under `set -euo pipefail` a failing git left the scrub **silently not happening** and the push proceeded. Found independently by lens 1 and lens 3; reproduced three ways.

This is #958 round 1's `return 1` in a second disguise — a guard asserted rather than run. Now `if ! scrub=$(...) || [ -z "$scrub" ]` with `exit 1`.

### MEDIUM ×2 — regressions this change introduced — **FIXED**

| | |
|---|---|
| `cd "$(git rev-parse --show-toplevel)"` sat *after* the scrub, which removed the only locator when the hook's CWD is outside the worktree (`git --git-dir=… push` from elsewhere). And `cd "$(...)"` swallows the failure **by the same mechanic as `unset $(...)`** — nine lines apart, and I fixed one without seeing the other. | `toplevel` now resolved before the scrub. OLD exit 0 · NEW-before-fix exit 1 in `/tmp` · NEW-after 0 |
| `SKIP_PREFLIGHT=1` no longer bypassed the new failure, breaking the contract line 8 advertises | early exit now precedes the scrub |

### HIGH — "CI structurally cannot execute a hook test" was false — **FIXED**

I shipped no test on that reasoning. It conflated testing a *push* with testing the hook's **env contract**, which is observable at its `exec`: stub `just`, hand the hook the `GIT_DIR` a linked worktree exports, read what the child saw. **184 ms**, no push, no worktree, no toolchain.

The failure it guards is silent by construction — delete the line and nothing else reds.

Design details that earned their place:
- **Enumerates `.githooks/*`** rather than naming `pre-push`, so a hook added later fails *here* instead of inheriting the gap. `pre-commit` is exempt by name while it only runs `cargo fmt`.
- **The control is a literal two-line hook**, not one derived by `grep -v` from a real one — deriving it orphans the guard's `if` body, so the mutant exits before `just` and the control silently stops controlling.

### The test's own first version was vacuous — **FIXED** (caught by running the mutation)

Its fixture `git init` ran under the `GIT_DIR` this suite exports at its top, so `$s/repo/.git` was never created; the hook then died at `toplevel=` under `set -e`, never reached `exec`, `seen` never existed, and `test -z ""` passed **against the real hook while the control still fired**. The harness was bitten by the bug it tests for. Cleared the list around that init.

### MEDIUM — `e2e-common-selftest.sh`'s scrub rationale — **FIXED**

It named the pre-push chain as its cause, which this commit makes false (the hook now scrubs upstream). The line still earns its place, so the cause was replaced: direct `just e2e-scrub-selftest` runs, and `submodule foreach` — verified in-session, it exports `GIT_DIR=.git` to children with no hook involved.

### REFUTED — keep `exit 1`, keep `[ -z "$scrub" ]`

Lens 1 measured that **git prepends its exec-path to a hook's `PATH`**, so a hook invoked by `git push` cannot fail to find git; `--local-env-vars` has existed since v1.7.1 (2010). So the guard's population is a `git` shim that exits 0 printing nothing. Not dead — the trigger *can* fire — and the alternative is the silent corruption this PR exists to prevent. Blocking is proportionate.

Sharp-edge check on the sibling: `e2e-common.sh:47`'s `-z` test looks identical but is **not the same claim** — there the value comes from a pipeline whose status belongs to `tr`, so `-z` is the only guard.

### Comment

Owner feedback: too long. Trimmed 7 → 4 lines. Trimming also dropped **`LINKED`**, turning a true sentence false — a plain checkout's pre-push exports nothing. Restored, still 4 lines.

### Teeth, verified by mutation
- delete the hook's scrub block → **reds**
- keep the block, unset nothing (behavioural) → **reds**
- empty the helper's scrub → **reds**

**Gates:** `shellcheck -x` 0 · `just shfmt-check` 0 · `just actionlint` 0 · `just lint` 0 (19/19) · selftest 9 checks · full `preflight` on push, 3174 passed.

IvanWng97:
## Round 2 + verification disposition (final head `2b312e58`)

Round 2 was the last behavior-changing commit; an independent verification lens then re-derived everything at this head. **Verdict: APPROVE-WITH-NITS, zero open HIGH.**

### Round 2 — the bot's MEDIUM — **FIXED, with a green-before/red-after pair**

The online review bot found that the new hook test's stub recorded only `${GIT_DIR:-}`, so a scrub narrowed to `unset GIT_DIR` passed while `GIT_INDEX_FILE` still relocated the index — half of the #893 mechanism, and the harness was already injecting it without ever looking.

The stub now reports **every** name from `git rev-parse --local-env-vars` still set at the `exec`. Verified independently by the verification lens, both directions: the round-1 test is **green** against the bot's exact mutant (proving the finding real); the round-2 test **reds** it. A stronger mutant (`unset GIT_DIR GIT_INDEX_FILE`, leaving only `GIT_WORK_TREE`) also reds — the assertion is general over git's list, not over names anyone picked.

Contract shape: a change shipping a test that fails without it (+7/−3, one test file, production untouched).

### Verification at `2b312e58` (default REFUTE — attacks that came back clean)

`${!v:-}` indirection on bash 3.2.57 · the generating `printf`'s quoting (byte-exact stub) · `rev-parse --local-env-vars` with no repo / a nonexistent `GIT_DIR` · word-splitting, IFS poisoning, glob-named files in CWD · the trailing `exit 0` on every path · shellcheck 0 · shfmt 0. All seven mutation rows behaved. The round-1 disposition's claims were re-audited item by item against this head: no mismatches. The real repository was bookended before/after: untouched.

### SURFACED — one line for the owner (V-1, MEDIUM, test blind spot; the hook itself is unaffected)

A hook that dies **before `exec`** passes the per-hook check vacuously: `:150` discards the hook's exit status and `:165` asserts an empty string, indistinguishable from a `seen` file never written. Live instance: `SKIP_PREFLIGHT=1` ambient in a developer's shell → the hook takes its documented early exit → the check is green even against a hook with no scrub at all. Measured both ways (same mutant reds with the var unset). It falsifies **none** of the mutation evidence above — every reported red runs through a path that reaches `exec` — it only names a mutant class that would slip through.

Follow-up one-liner when the owner wants it: assert the `seen` **file exists** (catches every died-before-`exec` cause), plus `SKIP_PREFLIGHT=` in the injected env as defense-in-depth. Not folded here: the two-round cap is spent, and this is a test-coverage nit, not a defect in the change.

V-2 (LOW): the injected `GIT_WORK_TREE` duplicates a file-scope export — inert, self-documenting, left as is.

**Gates at this head:** CI CLEAN (47 checks, 0 failing) · claude-auto-review **0 findings** · claude-security-review **0 findings** · `just lint` 0 (19/19) · selftest 9 checks · preflight on push 3174 passed.

A human merges.

PR #961, opened 2026-08-31
fix(e2e): a hook that dies before exec no longer passes the scrub check

IvanWng97:
The one SURFACED item from #960's verification round (V-1), as its own PR.

## The blind spot

The per-hook check asserted an empty **string**, and a `seen` file never written reads identically to one written empty. So a hook that died before `exec` passed the clean check vacuously. Two measured instances, both green before this fix:

- an early `exit` anywhere before the `exec`
- an operator's exported `SKIP_PREFLIGHT=1` sending the hook down its documented bypass — which the harness's `env` did not clear — leaving the check green **against a hook with no scrub at all**

## The fix, per the verification lens's recommendation

- A missing `seen` is emitted as `hook-died-before-exec`, which the clean check treats as a leak. General: it catches every died-before-`exec` cause, including ones nobody has thought of yet.
- `SKIP_PREFLIGHT=` cleared in the injected env — defense-in-depth for the one known instance, not the fix.

The control keeps its own teeth: with the `just` stub removed entirely, the suite still reds through the per-hook check.

## Measured, all six classes

| mutation | before | after |
|---|---|---|
| normal | green | green |
| `exit 7` after `set` | **green (the hole)** | red |
| empty scrub + ambient `SKIP_PREFLIGHT=1` | **green (the hole)** | red |
| scrub narrowed to `unset GIT_DIR` | red | red |
| whole scrub block deleted | red | red |
| harness self-broken (stub removed) | red | red |

Falsifies none of #960's mutation evidence — every red reported there ran through a path that reached `exec`; this closes the class that did not.

`shellcheck -x` 0 · `just shfmt-check` 0 · preflight on push 3174 passed.

### DIFF
--- a/scripts/lib/e2e-common-selftest.sh
+++ b/scripts/lib/e2e-common-selftest.sh
@@ -136,7 +136,8 @@ hook_hands_just() {
     mkdir -p "$s/bin" "$s/repo"
     # Cleared for the fixture's OWN init: this suite exports GIT_DIR at the top,
     # so a bare `git init` here re-inits THAT repo, leaves $s/repo without a
-    # .git, and the hook then dies before `exec` — passing the check vacuously.
+    # .git — the hook then dies before `exec`, which the died-before-exec check
+    # below turns into a red rather than the silent pass it used to be.
     # shellcheck disable=SC2046  # the var list must word-split
     (unset $(git rev-parse --local-env-vars) && git init -q "$s/repo")
     # Reports EVERY local-env-var still set, not just GIT_DIR: a scrub narrowed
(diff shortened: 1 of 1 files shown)


IvanWng97:
## Two-lens review — round 1 disposition (head `7e934080`)

Two differentiated lenses (correctness · design/blast-radius) + the online bot. All three converged on the same defect from three angles.

### HIGH/MEDIUM — the sentinel satisfied the OTHER consumer — **FIXED**

`hook_hands_just` had two consumers with opposite polarity: the per-hook check (`test -z` — sentinel = red, intended) and the negative control (`test -n` — sentinel = **pass**). Folding died-before-exec into stdout satisfied the control for free. Sharpest evidence (lens 1): mutate the **control hook itself** to die before `exec` — base reds the control correctly; the sentinel version ran **fully green with a dead negative control**. A PR whose thesis is "no vacuous passes" gave another check exactly one.

Fix: two channels. stdout stays the leak list; `cat`'s own exit status carries reached-exec (the status the sentinel's `||` was discarding); `hook_scrubs_clean` reads both. `local out` declared apart from its assignment — `local out=$(...)` would mask the status, which is the whole mechanism. The magic string is gone, and with it the collision question.

### Bot MEDIUM — the detector had no in-file negative control — **FIXED**

Every other guard in this file has a companion control; the died-before-exec detector didn't (my exit-7 evidence was all out-of-band mutation). A literal `exit 7` hook now pins it: breaking the status channel (`|| true` on the `cat`) reds it.

### Also folded
`env -u SKIP_PREFLIGHT` (it said "cleared" but set-empty; now literally true, and robust to a future `${VAR+x}` test) · the check message names *reaching exec*, not just leaking · the fixture-init comment's stated consequence had inverted (stale three lines above the diff).

### Verified
Lens 1 proved the two original halves **mutually necessary** (sentinel alone false-reds every dev who exports `SKIP_PREFLIGHT`; env-clear alone leaves V-1 open) and ran this revision against an 11-mutant matrix: dead control reds · detector-removed reds · pristine + ambient skip stays green (no false red). Lens 1 also probed **19 ambient variables** (`GIT_CEILING_DIRECTORIES`, `BASH_ENV`, `CDPATH`, `IFS`, …) — none can divert a hook under test — and accounted for every early exit in `pre-push`.

### SURFACED (pre-existing, one line, owner's call)
The pre-commit exemption rationale names "cargo fmt"; the greppable authority is the recipe `just fmt-check`.

**Gates:** selftest 11 checks · `shellcheck -x` 0 · `just shfmt-check` 0 · preflight on push 3174 passed. A human merges.

PR #962, opened 2026-08-31
fix(tests): an indented wrapped `Pinned by` claim is no longer skipped silently

IvanWng97:
Closes the one machine-checkable gap `pinned_by_claims.rs`'s own module doc listed: the comment-leader collapse anchored at column 0, so an **indented** wrapped `Pinned by` claim survived the collapse and was skipped silently. The tree already had one live instance riding the gap — `install/mod.rs:144`, whose named function happens to exist, but nothing was checking.

## The fix

Indentation is stripped per line before the collapse (`flatten_wrapped_comments`, extracted so the negative control can exercise it directly).

## TDD trail

1. Negative control gained the indented-wrap case → **red** (the exact live shape from `install/mod.rs`).
2. Per-line `trim_start` before the collapse → green.
3. Mutation — re-anchor the collapse at column 0 → the control **reds again**.
4. Real-tree sweep stays green: the newly covered claims all name real functions.

Module doc updated in the same commit: three gaps → two (the `.rs`-only walk and the prose-declaration scan remain, both still without a live orphan).

`cargo test -p pixtuoid-core --test pinned_by_claims` 0 · fmt 0 · clippy 0 · preflight on push 3174 passed.

### DIFF
--- a/crates/pixtuoid-core/tests/pinned_by_claims.rs
+++ b/crates/pixtuoid-core/tests/pinned_by_claims.rs
@@ -4,10 +4,10 @@
 //! existed.
 //!
 //! Two gaps, neither currently harbouring an orphan — widen before trusting it
-//! further: the walk is `.rs` under `crates/`, so the claims in `desk.sprite`
-//! and `add-source.prompt.md` go unchecked; and `declared` is a raw text scan,
-//! so `fn foo` written in prose counts as a declaration. It checks the name
-//! EXISTS, never that the test pins the claim.
+//! further: the walk is `.rs` under `crates/`, so a claim anywhere else — a
+//! sprite header, a prompt, `docs/` — goes unchecked; and `declared` is a raw
+//! text scan, so `fn foo` written in prose counts as a declaration. It checks
+//! the name EXISTS, never that the test pins the claim.
 //!
 //! Reads sibling crates at runtime, so it is workspace-only and sits in
 //! `Cargo.toml`'s `exclude`.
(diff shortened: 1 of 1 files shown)


IvanWng97:
## Two-lens review — round 1 disposition (head `36cb814a`)

Two differentiated lenses (correctness · design) + the online bot; both lenses and the bot converged on the HIGH from three angles.

### HIGH — the insert stole `claims_in`'s doc — **FIXED**

Inserting `flatten_wrapped_comments` anchored on `fn claims_in(...)` landed between that fn and its doc comment: the new fn carried a 12-line two-paragraph doc whose first half described `claims_in`'s return type and asserted — on the function that fixes it — that the collapse anchors at column 0; `claims_in` went undocumented, losing the only explanation of its non-obvious `trim_start_matches` design. In the file whose purpose is stopping unbacked prose. No gate can catch it: rustdoc and `missing_docs` skip test binaries. (The string-anchor-insert trap's sixth recorded occurrence in this repo's history.)

Docs split back; the now-false sentence and its dangling pointer struck.

### MEDIUM — the fixture that hid the bug survived — **FIXED**

`wrapped_across_lines` fed `claims_in` a **hand-flattened** string, so the flatten was never under test — which is exactly how the column-0 gap stayed invisible. It now routes through `flatten_wrapped_comments` with a real wrap; tooth verified (deleting the `\n///` replace reds it).

### MEDIUM — the gap sentence's file list had already drifted — **FIXED**

Census: **three** matchable unchecked claims in the tree; the list named two (`docs/mutants-exclusions.md:85` was the miss — not an orphan, its target exists). Category shapes replace the list: the walk root one clause earlier is the authority; a file list is the restated value, and it rotted once already.

### Nits — **FIXED**
Assert message cited `install/mod.rs` by name — a reference with no failure mode; cites the shape now. Measured coverage delta is **two** newly-checked claims (`install/mod.rs:144`, `state/mod.rs:719`), not the one the original commit body named; both name real functions, zero orphans at base and head (25 → 27 claims).

### REFUTED / no-action (lens 2's design sweep)
Text-flattening stays the right altitude (`syn` discards non-doc `//` trivia — the exact shape under test) · no sibling flattener in the tree (`drift_surface.rs`'s `PINNED_ELSEWHERE` is same-concept/different-seam) · block comments covered incidentally; `#[doc =]` absent from the tree.

**Gates:** `cargo test --test pinned_by_claims` 0 (2 passed) · fmt 0 · clippy 0 · preflight on push 3174 passed. A human merges.

PR #964, opened 2026-08-31
feat(omp): the extension bridge — approval waits, instant presence, empty-session shutdown

### LINKED ISSUE #951: Use omp's extension API to complete lifecycle, approval, and source-health support
## What's the problem or use case?

The recent omp work has brought its transcript integration a long way. It now covers tool activity, nested subagents, clean exits, liveness, focus jump, model and reasoning effort, token usage, session titles, and decode drift. The remaining gap is no longer in the transcript decoder. It is that omp is still registered as a transcript-only, flag-only source even though omp now exposes a documented extension API for the events that transcripts cannot carry.

## Where parity still breaks

| Area | Current support | Remaining gap | Omp extension signal |
|---|---|---|---|
| Tool approval | The transcript exposes `ask`, tool calls, and tool results. | A normal tool approval can leave the sprite looking active while omp waits for the user. | `tool_approval_requested` and `tool_approval_resolved` |
| Immediate session presence | The JSONL watcher registers a session once durable transcript content exists. | Omp persists sessions lazily, so a new session may not appear before its first assistant response. | `session_start` |
| Empty-session shutdown | A persisted `session_exit` removes ordinary completed sessions promptly. | Omp omits that entry when no assistant message was produced, leaving the stale sweep as the fallback. | `session_shutdown` |
| Sources-panel health | Omp can be connected as a transcript-only, flag-only source. | It is omitted from first-run target detection, an absent CLI is indistinguishable from an available targetless source, and `doctor` has no bridge artifact to verify. | An installed extension target |
| Concurrent profiles | `omp_sessions_dir` mirrors Omp's profile, XDG, override, and environment rules for one resolved root. | Two omp processes using different profiles cannot both be observed by an ordinary pixtuoid process. | Separate multi-root watcher work |

The clearest missing user state is tool approval. It is exactly the kind of blocked session the office is intended to make visible, and the transcript cannot recover it after the fact.

## Proposed solution

Promote omp to a hybrid extension and transcript source. Keep the JSONL watcher as the durable authority and fallback, then install a small auto-discovered extension at the active omp agent directory's `extensions/pixtuoid.ts` path.

The installer can follow the existing opencode plugin-as-config pattern. Pixtuoid would own the TypeScript file, bake in the absolute `pixtuoid-hook` path, verify a sentinel and the baked path, replace the file with a valid inert module on disconnect, and detect omp independently of the plugin file. Connecting should explain that already-running omp sessions must restart before the extension is loaded.

The extension should forward only an explicit allowlist of local identity and state fields: event type, session file, session id, cwd, tool name, tool call id, approval state or reason, and `process.pid`. It must not forward prompts, messages, tool arguments, tool results, or model output.

The extension must never block omp. Omp extensions run in-process, and `tool_call` handlers fail closed when they throw. The bridge should therefore copy the OpenClaw plugin's safety posture: spawn `pixtuoid-hook` detached, await nothing, catch every exception, ignore child-process errors, and never return a blocking or mutating result.

Hook identity must be derived from `sessionManager.getSessionFile()` through the existing `omp_id_from_path`. Using omp's bare header UUID would split the hook and transcript sides because pixtuoid currently keys roots by the timestamp and UUID stem, then keys children by the nested stem chain. A UUID fallback is still useful for `--no-session`, where there is no transcript file to coalesce with.

The first event set should cover:

- `session_start`, `session_switch`, `session_branch`, and `session_shutdown` for immediate and empty-session lifecycle
- `tool_approval_requested` and `tool_approval_resolved` for accurate `Waiting` state
- `process.pid` as an authoritative plugin stamp, while retaining the append-file-descriptor probe as the transcript-only fallback

The existing transcript path should continue to own model, effort, usage, title, subagent structure, tool detail, and durable exit. It must remain fully functional under `omp --no-extensions` or after a broken or removed plugin.

The approval transition needs one explicit reducer decision. Resuming the same tool-call id after approval should return the slot to `Active` without incrementing `tool_call_count` a second time. That avoids adding another known duplicate-count path while preserving hook-wins dedup for the ordinary tool stream.

Profile support should be a second slice of the same effort, or a follow-up issue if it makes the first patch too broad. One omp row should mean all locally discoverable profiles. The watcher can cover the default root and each existing named-profile root, then rescan for new profile directories. It should use explicit session roots and the existing path filter rather than recursively accepting arbitrary JSONL under the whole omp config tree. Arbitrary per-process `PI_CODING_AGENT_DIR` roots will still require explicit configuration or an installed bridge.

## Acceptance criteria

- A newly started omp session appears before its first assistant message is persisted.
- A normal omp tool approval displays `Waiting`, then returns to `Active` or ends cleanly when resolved.
- A session that shuts down before producing an assistant message is removed promptly.
- Extension and transcript events coalesce into one agent id for root sessions and nested subagents.
- One logical tool call increments the tool count once when both transports are enabled.
- Missing pixtuoid, a missing socket, a stale shim path, and extension handler errors never block an omp tool call or shutdown.
- `omp --no-extensions` retains the current transcript-only behavior.
- The Sources panel can detect omp, install or remove the bridge, and report a stale or broken bridge through the existing health path.
- The upstream drift check covers every omp extension event name and field that the bridge reads.
- Captured fixtures cover early start, approval and resolution, empty-session shutdown, resume or switch, and a nested task agent.

## Alternatives considered

Adding more transcript decoding cannot recover approval events because omp does not persist them. Decoding the existing `tool_execution_start` custom entry would improve execution timing, but it would duplicate the assistant `toolCall` already decoded and still would not expose the approval wait.

Replacing transcripts with the extension would make the integration less reliable. Extensions can be disabled, and transcripts carry the durable model, usage, title, subagent, and exit information that the extension should not duplicate.

Adding only a presence probe would improve onboarding and the Sources panel, but it would leave the lifecycle and approval gaps untouched. The install target is more useful when it installs the bridge that closes those gaps.

## Area

New agent-CLI source

## Contribution

I would be willing to open a PR for this.

IvanWng97:
Closes #951 (slice 1 — the extension bridge; multi-profile watcher roots and plugin-stamp pid arming remain follow-ups, as the issue allows).

omp becomes a hybrid extension+transcript source. The JSONL watcher stays the durable authority; a pixtuoid-owned TS extension auto-discovered from `omp_extensions_dir()` forwards the allowlisted lifecycle + approval events through `pixtuoid-hook`. What transcripts can never carry now reaches the office: pre-persist presence, the tool-approval Waiting, empty-session shutdown.

**Live dogfood** (worktree binary, omp 18.0.11, real approval round through the real shim + socket):

```
hook SessionStart (onDisk:false — before any transcript)  → slot appears
agents=[…, om·Run shell command echo hi@2:waiting(bash)]  → gate visible
agents=[…, om·Run shell command echo hi@2:active(bash)]   → approval resume
agents=[om·proj@0:idle, om·proj@1:idle]                   → shutdown walkout
```

Both transports mint ONE AgentId (the hook decoder keys the `sessionFile` through the watcher's `normalize_path_key` fold + `omp_id_from_path`; upstream's `getSessionFile()` returns the allocated path pre-persist). Hermetically re-proven per recorded round by `each_bridge_round_coalesces_both_transports_onto_one_id`, including the one-count claim through a real `Reducer` in the recorded human ordering.

## Review record (two-lens gate, run pre-PR)

Hand-written churn (~2.1k added) exceeds the ~1500 budget. Rather than splitting the interleaved stack into two never-tested-together trees post-review, it was REVIEWED as two lens-scoped halves — **PR-A** (the reducer mechanism, `3da6f25d..08cf4055`) and **PR-B** (the bridge, `08cf4055..HEAD~1`) — under eight differentiated lenses: correctness + design + reducer-escalation on A; correctness + design + install-axes + new-source-live-replay + comments on B. All upstream claims were re-fetched in-session by the lenses independently. **If you prefer the physical stacked split, say so and I'll cut it mechanically.** Round 1's findings are folded in one commit (`fix(review): round-1 fold…`, with two `plan-miss:` lines); round 2 verified the fold's tests against the mechanism (all new fixes ship with a test that fails without them).

### Dispositions (every finding, one terminal state)

FIXED — in the fold commit:
- Resolve wipes the whole gate set (3 lenses, HIGH-adjacent): member-scoped retirement; None-id hook end keeps the turn-end wipe; pinned by `a_denial_resolving_one_gate_keeps_the_queued_siblings`.
- The ungated-Start whole-wipe strips pending/queued gates (straggler + auto-approved-parallel scenarios): spared while Waiting or for an already-counted call; two pinning tests.
- Leak census enumerated 7 of 8 correlation maps (3 lenses, conf 95-98): `counted_calls` joins all three arrays.
- semver: `Waiting.tool_use_id` is `enum_struct_variant_field_added` (exit 100 observed) → workspace 0.18.0, changelog arm with real bullets, dep pins bumped; `cargo semver-checks` now exits 0.
- Windows fold class (HIGH, conf 88): hook-keyed expectations were raw literals red only in `windows-test`; every expectation now derives through `normalize_path_key`, the three hybrid rounds moved to the `sources/omp` module suite (their platform-dependent snapshots deleted, replaced by the hermetic coalescing pin), consequence documented on the decoder.
- The extensions-discovery seam had no watch and fails GREEN (2 lenses, HIGH-class): core owns `omp_extensions_dir()` + `EXTENSIONS_SUBDIR`, shipped into the drift fragment, anchored on `builtin.ts`'s join AND `getAgentDir()`'s un-redirected return, census-carried; plus the local `default_path…` pin.
- `approved` required→optional would silently read every approval as a denial (conf 90): a targeted non-optionality arm + a bespoke selftest.
- A relative `PI_CODING_AGENT_DIR` became a WRITE path: refused with an error naming the fix (upstream resolves against omp's cwd, which pixtuoid can't know).
- Recorder identity in three provenance `command` fields via the dash-encoded form the PII gate can't see (conf 82): the pty driver is now committed (`scripts/lib/capture-omp-tui.py`) and referenced descriptively; the identity config gains the dash-encoded rule with both selftest halves.
- `session_switch` had zero recorded bytes: a real `/new` round recorded through the committed driver (`switch-recorded`) + a decode pin.
- Comment truth (comment lens, 10 items + 3 echoes from other lenses): `_pid` is stamped-but-discarded (3 sites now say so), `verify_schema` enumerates its actual four checks, the `.env` cwd file is FIRST-and-strongest not fourth, empty sessions persist a HEADER but no `session_exit` (module doc + provenance), authorities named instead of restated (shim write-timeout, `verified_version`), rationale hoists onto declarations, cross-crate paths qualified.
- The bridge-only focus boundary + the copilot-divergence wire fact (transcript writes the call 4 ms before the approval request; 7.5 s human latency) documented on the declarations they constrain.

REFUTED — with the mechanism:
- "Move empty-session into conformance" — the same Windows fold divergence would red `windows-test`; the module coalescing test carries the coalescing evidence instead (its doc corrections were taken).
- Gated-Jsonl detail loss — deliberate label-only cost, pinned by `the_approval_resume_labels_from_the_wire_tool_name`; CC's instance is harmless (its hook Start already set the rich detail).
- "The TOOL_ASK sibling comment is stale under retain semantics" — checked, still accurate (the ask path is inference-gated; the whole-set clear it references still fires) — flagged-then-dropped by the lens itself after the sharp-edge check.

DROPPED (taste, per the blocking bar):
- Nested-map reshape of `counted_calls` / gate+count merge into one registry (design commentary, marked non-blocking by its lens; the flat composite retain is O(small)).
- A TTL for gate members (bounded by the resolve/turn-end/new-tool clears).

SURFACED (pre-existing; owner's call, one line each):
- `install/target.rs` opencode row's comment claims sentinel-based presence detection; `opencode::detect_installed` probes opencode's own dirs.
- Deleting an installed plugin/extension FILE silently greens doctor for opencode and omp alike (`has_hooks` short-circuits `verify_target`).
- `PI_CODING_AGENT_DIR` divergences from upstream in the shared resolver: no `path.resolve` absolutization; `isDefault` hardcoded false for overrides (sessions-XDG only); the profile-derived-override drop compares `PathBuf`s where upstream string-compares (trailing-slash flip; note the per-CLI mirror rule cuts against the house structural-compare rule here).
- omp discovers extensions through a gitignore-filtered glob — a dotfiles-repo ignore rule can suppress loading while everything reads green (accepted residual documented on `default_config_path`; a `git check-ignore` doctor note is the candidate fix).
- snapbox goldens ESCAPE a `.claude/worktrees/*` checkout (`cargo_rustc_current_dir!`'s outermost-Cargo.toml heuristic): verify compares the MAIN clone's snapshot and `SNAPSHOTS=overwrite` writes into it (observed; main restored; local workaround `CARGO_RUSTC_CURRENT_DIR=<worktree>`).
- The live drift run reports upstream codex `RolloutItem` grew `realtime_item` (advisory, untouched here).
- `tests/CLAUDE.md:4` says "9 test binaries"; the tree holds 10 (predates this change).

## Verification

`just preflight` green at HEAD (fmt·lint·clippy·hack·test·deny·shellcheck·fixpii·e2escrub·ci-obs). `cargo semver-checks` exit 0 at 0.18.0. Drift selftest + LIVE upstream run green (three new anchored rows). `just gen-drift-surface`, `just api-surface`, `just gen-wasm` regenerated and committed; `sources --json` golden is snapbox's own overwrite output verified under a pinned snapshot root; site row + manifest test green; README unchanged by `gen-readme`. CI-only gates this diff can red were each addressed in-diff (semver/api-surface/windows-test/insta/gen-check); `npm-check` owns the Raycast half (schemas byte-identical, so no regen owed).

## Follow-ups (slice 2, per the issue)
- Multi-profile watcher roots (one omp row = all locally discoverable profiles).
- Plugin-stamp pid arming (the TS already stamps `_pid`; `TranscriptProbe` discards it today — documented on the row — so the wire shape is ready).


findyourexit:
You legend, thank you 🙏

IvanWng97:
@findyourexit tested it locally with my own OMP config. feel free to raise any issues~

PR #966, opened 2026-09-01
chore(ci): tier A of the CI audit — composites, caches, and one-authority pins

IvanWng97:
Tier A of the CI audit (items 1-7): the mechanical, low-risk consolidations. Net −50 lines with two new composites; items 8-11 (cache tuning, non-Rust-PR skips, site/pages composite) follow in a second PR.

## What

- **`setup-alsa` composite** — 11 bare `apt-get install libasound2-dev` copies (4 comment variants, none retried) become one composite carrying the #633 WHY and the smoke job's mirror-retry envelope; the flake class costs a full 47-min re-run per occurrence.
- **`require-jobs` composite** — the three byte-identical jq aggregators share one shape (with a guard: a caller omitting `results:` fails loud instead of gating nothing — `required:` inputs are advisory per GitHub's own docs). ci.yml's gate deliberately keeps its hand-written env reads: the policy pins them (`must read needs.*.result`).
- **go-tools cache** — the hygiene job stops compiling five Go tools from source every run (148s of a 189s job); key hashes ci-lint.yml itself so a version bump can never serve a stale binary.
- **Three one-authority fixes** — the Linux `--no-default-features` flag derives inside `build-target` from the target; `just workspace-version` owns the version parse for the tag check and `just bump`; `API_PUBLIC_API` + a runtime assert in `_api-toolchain` pin the cargo-public-api pair so divergence fails loud instead of churning goldens.
- **Policy** — the dead `supplemental` exemption (job deleted in #956) goes, with its clause, fixture row, and test; the behavior test's seven publisher blocks share `run_publisher`.
- **Trivia** — codspeed gains the standard concurrency block (keeping the read floor); the Chromium-before-check comment moves onto the step it describes.

## Review (two-lens gate)

3 lenses (correctness / design / whole-file comment audit): APPROVE-WITH-NITS ×3 effectively; fold `d32afcb5` carries the dispositions — highlights: the vacuous-pass guard (probe-verified: actionlint accepts an omitted `results:`, jq exits 0 on empty), the "non-advisory" deny-message rot, the stolen `build` comment block, codspeed's lost permissions floor. All local gates green: preflight (3218 tests), ci-obs (117 rego + 113 conftest), actionlint, actionlint-composites, zizmor 1.30, shellcheck/shfmt, standalone behavior test.

**Tag-time-only surfaces** (release.yml never runs on a PR): `build-target`'s derivation and `just workspace-version` — exercised locally instead: native `just build-target aarch64-apple-darwin false` exit 0; the derivation replayed over all 7 matrix targets matches the deleted expression; `just workspace-version` → 0.18.0. Recommend one by-hand musl `build-target` before the next tag.

## Surfaced to owner (pre-existing / trades, one line each)

- `packaging-build/action.yml:41` keeps an independent, differently-implemented version parse (deliberately just-free; `cargo metadata --no-deps` could replace it if wanted).
- `release.yml`'s musl-tools install is still a bare un-retried apt call on the tag-only path (a generalized apt composite would collect it and the ffmpeg install).
- The three aggregators gained a checkout — a transient checkout failure now reds the merge-gate job (fail-closed, but a new network dependency).
- The go-tools cache populates only on a fully green hygiene job (actions/cache semantics); fork PRs read-only, degrade gracefully.

### DIFF
--- a/.github/actions/packaging-build/action.yml
+++ b/.github/actions/packaging-build/action.yml
@@ -10,10 +10,7 @@ runs:
         prefix-key: v1-rust
         # rust-cache isolates by OS, so one shared-key is safe across platforms.
         shared-key: packaging-build
-    - name: Install ALSA headers (Linux only)
-      if: runner.os == 'Linux'
-      shell: bash
-      run: sudo apt-get update && sudo apt-get install -y libasound2-dev
+    - uses: ./.github/actions/setup-alsa
     - name: cargo install --locked (both binaries, default features)
       shell: bash
       run: |
--- a/.github/actions/require-jobs/action.yml
+++ b/.github/actions/require-jobs/action.yml
@@ -1,0 +1,24 @@
+name: 'Require Jobs'
+description: 'Fails unless every needed job succeeded — skipped and cancelled red too'
+
+inputs:
+  results:
+    description: 'toJSON(needs) of the calling aggregator job'
+    required: true
+  label:
+    description: 'Failure-message prefix naming the group'
+    required: true
+
+runs:
+  using: 'composite'
+  steps:
+    - shell: bash
+      env:
+        RESULTS: ${{ inputs.results }}
+        LABEL: ${{ inputs.label }}
+      run: |
+        failures="$(jq -r 'to_entries[] | select(.value.result != "success") | "\(.key)=\(.value.result)"' <<< "$RESULTS")"
+        if [[ -n "$failures" ]]; then
+          echo "::error::${LABEL}: ${failures//$'\n'/, }"
+          exit 1
+        fi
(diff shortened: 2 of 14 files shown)


IvanWng97:
Disposition of the auto-review finding at `0232c053` (1 MEDIUM, ci.yml:41):

**REFUTED — the fourth copy is policy-mandated, not a DRY gap.** `main.rego` requires the gate job to literally read each group result (`must read needs.<name>.result`, the deny at the `ci_gate_membership` block; regression-tested by `test_ci_gate_need_on_the_advisory_group_is_denied`'s surviving membership helper). Routing ci.yml's gate through `require-jobs` erases those env reads and reds `ci-observability`. The correctness lens verified the pin holds against the real `ci.yml` before the composite conversion deliberately skipped this site; the commit message and PR body both record it. A future `require-jobs` semantic change therefore *shouldn't* propagate here — the gate's explicit per-group reads are the policy's chosen shape for the one branch-protection context.

Issue #157, opened 2026-06-07
Source death is invisible in default TUI mode (no tracing subscriber)


IvanWng97:
Surfaced by the PR #155 silent-failure review (Windows arc, #128), but **pre-existing and platform-independent**.

## Problem

When a source's `run()` returns `Err` mid-session (e.g. the new Windows pipe listener's recreate-bail, or any future fatal source error), the chain works as designed — `claude_code.rs`'s `select!` aborts the sibling transport, `SourceManager::spawn` logs `tracing::error!("source died: {e}")` — but in the product's **default TUI mode no tracing subscriber is installed** (`main.rs`: file logging only under `PIXTUOID_LOG` or debug/trace). The error goes nowhere.

Net effect: the office silently stops updating — existing sprites stale-sweep out over 10–60 min, new sessions never appear, zero feedback anywhere. The worst class of partial failure.

## Why deferred

Out of scope for PR #155 (compile-on-Windows): the real fix is a product surface, not a logging tweak.

## Fix sketch

A source-health indicator in the TUI: `SourceManager` publishes died-source info into `SceneState` (or a side channel), and the status footer / wall display shows a one-shot "⚠ claude-code source died — restart pixtuoid" message. Alternatively (cheaper): always install a minimal file-logging subscriber so `~/.cache/pixtuoid/log` captures `error!`-level events even in default mode.

PR #969, opened 2026-09-01
feat(omp): every profile gets a desk — one watcher per profile sessions root

IvanWng97:
#951 slice 2, the roots half: every omp profile gets a desk. Pid-arming stays #951's last open piece, so this PR does not close the issue.

## What

omp keeps a sessions tree per profile; the watcher covered only the env-selected one, so a session under any other profile never became a sprite.

- **Enumeration** — `omp_profile_sessions_dirs` lists `profiles/*` under the base config root and derives each profile's sessions dir with the selected-profile precedence the resolver already encodes (own `agent` dir; the XDG arm flattens `agent/` away where its candidate exists), excluding the env-selected primary. Names go through `normalize_profile_name`, so a stray `default` dir or an invalid name can never be materialized as a root by the watcher's `create_dir_all`.
- **One watcher per root** — `OmpSource::run` drives a `JoinSet`, rescanning on the watcher poll authority (`spawn_blocking`, the liveness-probe precedent) so a profile created mid-run gains a watcher without a restart. One root failing logs and leaves the rest alive, but the failure is REMEMBERED: when every watcher is gone the error propagates to the deaths surface (#157) — pinned by a native-backend test on an unwatchable root.
- **First-party probes** — roots from `default_paths`' enumeration get the fd-liveness probe by construction (the layout gate exists to exclude user-supplied replay roots; a resolver-derived root never is one), so XDG- and `PI_CONFIG_DIR`-relocated profiles keep real liveness. `single_root` (the test/replay shape) stays gated.
- **Focus** — `FocusPaths` carries every omp root; clicking a profile sprite resolves its pid instead of silently no-oping under doctor's probe ✓.

## Review (two-lens gate)

3 lenses (correctness / design / whole-file comment audit): REQUEST-CHANGES ×2 on one HIGH — the first cut's log-and-continue loop swallowed total watch failure, regressing the #157 death-signal contract while multiplying its likeliest trigger (inotify watches × N roots). Fold `23bb5240` carries all dispositions: 7 FIXED (death signal, probe trust, focus re-scope, junk-name filter, blocking rescan, named path-segment consts + doc(hidden) seams, comment findings), 1 REFUTED (exact-start-count dedup teeth — one watcher emits SessionStart twice by design, seed + decoded header), 2 SURFACED below. Correctness lens ran eight gates itself, all exit 0 (including semver — the new pub fields are 0.x-major-class, absorbed by the unreleased 0.18.0 — and api-surface, doc-check, windows-structural).

## Surfaced to owner

- `JsonlWatcher::run` does `create_dir_all` on its root, so pixtuoid now materializes `profiles/<p>/agent/sessions` for profiles that never ran omp — pre-existing behavior for the primary root, multiplied by profile count.
- Only the selected profile's `.env` overlays enumeration (upstream freezes its resolver per process the same way); a profile relocating its own dirs via its own `.env` is unseen — documented on the fn.
- doctor/corpus census still reports the primary root only (the root row now says PRIMARY).

### DIFF
--- a/api/pixtuoid-core.txt
+++ b/api/pixtuoid-core.txt
@@ -524,9 +524,13 @@ impl !core::panic::unwind_safe::RefUnwindSafe for pixtuoid_core::source::manager
 impl !core::panic::unwind_safe::UnwindSafe for pixtuoid_core::source::manager::SourceManager
 pub mod pixtuoid_core::source::omp
 pub struct pixtuoid_core::source::omp::OmpSource
+pub pixtuoid_core::source::omp::OmpSource::profile_sessions_roots: alloc::vec::Vec<std::path::PathBuf>
+pub pixtuoid_core::source::omp::OmpSource::rescan: pixtuoid_core::source::omp::OmpProfileEnumerate
+pub pixtuoid_core::source::omp::OmpSource::rescan_interval: core::time::Duration
 pub pixtuoid_core::source::omp::OmpSource::sessions_root: std::path::PathBuf
 impl pixtuoid_core::source::omp::OmpSource
 pub fn pixtuoid_core::source::omp::OmpSource::default_paths() -> Self
+pub fn pixtuoid_core::source::omp::OmpSource::single_root(std::path::PathBuf) -> Self
 impl pixtuoid_core::Source for pixtuoid_core::source::omp::OmpSource
 pub fn pixtuoid_core::source::omp::OmpSource::name(&self) -> &str
 pub async fn pixtuoid_core::source::omp::OmpSource::run(alloc::boxed::Box<Self>, pixtuoid_core::TaggedSender) -> anyhow::Result<()>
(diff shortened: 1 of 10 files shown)


IvanWng97:
Live verification (this session, local omp 18.0.11): `OMP_PROFILE=pixtuoid-live-probe` + empty-session `/exit` created exactly `~/.omp/profiles/pixtuoid-live-probe/agent/sessions/<encoded-cwd>/`, and production `omp_profile_sessions_dirs()` run against the real env enumerated exactly that root. (Empty-session `/exit` persisted dirs but no header this time — the slice-1 "header always persists" note over-generalized; watcher-on-root remains covered by the integration tests.) The probe was also the first thing to ever execute `capture-omp-tui.py`, which turned out unparseable on main (`rr"""`, shipped in #964) — fixed in `a6cbc9cc`.

PR #980, opened 2026-09-03
fix(deps): remove ttf-parser from the graph, retiring the last deny ignore

### LINKED ISSUE #440: Drop the deny.toml ttf-parser advisory ignore when ab_glyph moves off it
**Tracker for the ONE remaining `[advisories] ignore` entry in `deny.toml`.**

Opened for the three entries added in the 2026-07 cleanup (they revived the daily `security audit` cron, silently red since 2026-06-30). The two quick-xml rows are **done**: wayland-rs shipped against a fixed quick-xml, and RUSTSEC-2026-0194 / RUSTSEC-2026-0195 have been removed from `deny.toml`. Only the ttf-parser row is left.

| Advisory | Path | Unblock condition |
|---|---|---|
| RUSTSEC-2026-0192 (ttf-parser unmaintained) | `ttf-parser <- owned_ttf_parser <- ab_glyph`, and `ab_glyph` enters the graph by **two** routes: our OWN direct dep (`crates/pixtuoid/Cargo.toml`, rasterizing in `src/aa_text.rs`) on every platform, plus `sctk-adwaita <- winit` on Linux (Wayland CSD titlebar) | `ab_glyph` migrates off ttf-parser (the advisory points at `skrifa`), or we drop `ab_glyph` |

## Corrections to the original filing

Verified 2026-07-27 against `cargo tree -i ttf-parser` and the RustSec entry itself.

- **Not Linux-only.** `cargo tree -i ttf-parser --target x86_64-apple-darwin` resolves `ttf-parser <- owned_ttf_parser <- ab_glyph <- pixtuoid` — the direct dep puts it in the macOS and Windows graphs too. The original table recorded only the `sctk-adwaita <- winit` route.
- **Consequently, winit 0.31 / sctk-adwaita moving off ab_glyph would NOT clear this row** — the stated unblock condition was wrong. #441 has been corrected accordingly.
- **There is no upstream release to wait for.** RUSTSEC-2026-0192 is `informational = "unmaintained"` with `patched = []`; the crate will not be fixed. The original title ("when upstream releases land") described a condition that cannot occur.

## Exposure while ignored: nil

`aa_text` parses exactly one face, the `include_bytes!`-compiled Monaspace Neon. The Linux sctk-adwaita route parses whatever local system font `fc-match` resolves (or its embedded Cantarell fallback), on a titlebar our floating window disables anyway (`with_decorations(false)`). Neither route sees attacker-supplied input.

## Notes

The same rationale is duplicated at the ignore's use site in `deny.toml` — keep the two in sync. A stale ignore (advisory no longer matching any crate) now fails the daily `security audit` cron via `[advisories] unused-ignored-advisory = "deny"`, so this row cannot rot silently once the unblock lands.

Deferred from the 2026-07 whole-codebase cleanup (deps-rust audit).

IvanWng97:
## Summary

Cuts both routes that put the unmaintained `ttf-parser` (RUSTSEC-2026-0192) in the dependency graph, so `deny.toml`'s last `[advisories] ignore` entry can go: `aa_text` parses with `skrifa` instead of `ab_glyph`, and `winit` takes `wayland-csd-adwaita-notitle` instead of the `ab_glyph`-backed CSD titlebar it never draws.

## Related issue

Closes #440

## Type of change

- [x] Refactor / chore

## How I tested it

**Dependency graph** — `ttf-parser` is absent on `x86_64-unknown-linux-gnu`, `aarch64-apple-darwin` and `x86_64-pc-windows-msvc` under `--all-features` (the graph `deny.toml` configures). Net lockfile churn is three out (`ttf-parser`, `owned_ttf_parser`, `ab_glyph`) and three in (`skrifa`, `read-fonts`, `font-types`); the new crates are `MIT OR Apache-2.0` and `Apache-2.0`, both already in the `[licenses] allow` list.

**Pixel identity** — the reason `ab_glyph_rasterizer` is kept rather than moving the whole stack. Two layers of evidence:

1. *Unit* — a differential oracle dumped every coverage sample the public API emits, run against both implementations. **At 8-bit output the streams are identical (max blend delta `0`)**, which is what every consumer sees.

   An earlier revision of this section claimed **max f32 coverage delta `0.000e+00`**. That was measured on too coarse a grid (4 sizes × 6 origins, none past ~1024) and is **false as written** — review widened it and found a real divergence: this port had collapsed `ab_glyph`'s trunc/fract pen split into `pen + offset`, equal in exact arithmetic but not in f32. Wherever `pen + coord·k` lands within one f32 ulp of an integer the offset's low bits vanish and the box origin moves a whole device pixel — on **both** axes, so it reaches down to `pen_x` 0 through the baseline term, and fires in ~0.0156% of (glyph, size, pen, baseline) combinations. Concretely, at `pen_x` 2047.6804 with a scaled offset of 6.3196507 the exact sum 2054.0000705719 rounds to `2054.0`, so `ceil` gives 2054 where the split gives 2055. Measured before the fix: max coverage delta `1.77e-5`, 250,034 pixels lit only by the old path, 15,541 of 99,553,428 swept origins disagreeing — all of it inert at 8 bits (`8.23e-5 × 255 = 0.021` against a 0.5 flip threshold), which is why the stills never moved. **`ab_glyph`'s split is now restored**, so the streams agree in f32 too rather than merely below the rounding threshold.
2. *End-to-end* — every pixel-gated committed artifact reproducible without the media venv re-renders byte-identical:

   | artifact | `aa_text` path |
   |---|---|
   | `docs/images/reference-screenshot.png` | TUI cell text (`encode.rs`, 14.7px) |
   | `docs/images/reference-night.png` | same |
   | `site/public/demos/proof-poster.png` | `--proof` panel (`proof.rs`) |
   | `site/public/demos/proof-tall-poster.png` | same |

   Those cover both call paths, so `gen-check`'s zero-threshold diff needs no reshoot and no media is committed here.

**Tests / lint** — `cargo test -p pixtuoid --lib`: 1005 passed. The one failure, `onboarding_overlay_renders_roster_and_hint`, is a pre-existing `--no-default-features` artifact (it asserts the audio offer; this environment cannot install ALSA headers) and fails identically on the base commit. `cargo clippy -p pixtuoid --all-targets`: clean.

**Since verified locally** (the gates the authoring environment could not install): `cargo deny check advisories` and `bans licenses sources` on cargo-deny **0.20.2**, the exact CI pin — both exit 0 with `ignore = []` under `unused-ignored-advisory = "deny"`. `cargo machete`: 0. `TZ=UTC scripts/gen-media.py --check`: 0, every committed artifact in sync at threshold 0, re-run after the pen-split fix. Full `just preflight` (lint → clippy → hack → test) passes via pre-push: 3,252 tests, 0 failures.

## Notes on the approach

`ab_glyph` is two crates: a parsing half (`owned_ttf_parser` → `ttf-parser`) and a rasterizing half (`ab_glyph_rasterizer`) that depends on nothing but an optional `libm`. Only the parser is the advisory's subject, so only the parser is replaced — keeping the rasterizer is what holds the AA curve, and therefore the committed stills, unchanged. `swash` or `skrifa` + `zeno` would have moved every anti-aliased edge for no gain.

Two constants of the old API had to be preserved deliberately:

- **Size convention.** `ab_glyph` sized text so `px` spans ascent−descent; skrifa's `Size` is ppem, and for this face those differ (1000 units against 1155). `px_per_unit` converts, rather than re-tuning ~40 call sites.
- **Bounds.** Control-point bounds are looser than the outline box `ab_glyph` sized its rasterizer with. That costs a few all-zero coverage rows and nothing else — the absolute pixel a sample lands on is `origin + rasterizer coord`, and a larger box shifts both terms by the same integer.

Performance, measured on an 11-char label: `draw_text_at` 20.4µs → 30.5µs (skrifa validates CFF charstrings harder, matching the numbers on the upstream ab-glyph skrifa port), against a 33.3ms frame the whole text pass now takes 1.24% of, up from 0.82%. `text_width` goes the other way, 0.30µs → 0.17µs, because `Charmap` is now held instead of rebuilt per character — `MetadataProvider::charmap()` re-runs cmap subtable selection on every call.

`winit` dropping `default-features` means its other four defaults are listed verbatim; a missing one is a Linux runtime failure (no backend), not a build error.

## Visual verification

- [x] Not a visual change — it is a rendering-stack change whose verification is the pixel identity above: the four committed stills that exercise `aa_text` re-render byte-for-byte, so there is nothing to look at.

## Checklist

- [x] I read the relevant `CLAUDE.md` (root + `crates/pixtuoid/`).
- [x] New behavior has a test — no new behavior; the existing `aa_text` tests are the regression net and the byte-exactness check above is the evidence for the swap.
- [x] Docs updated in the same commit — the rationale lives on `aa_text`'s module doc and the two `Cargo.toml` declarations; no module structure or public API changed.
- [ ] `just preflight` passes locally — **could not run**: `cargo-deny`, `cargo-hack`, `cargo-machete` are unavailable in this environment (see above). Everything reachable here (`cargo test`, `cargo clippy --all-targets`) is green.

## AI assistance

https://claude.ai/code/session_01AEqEB4zXiPhLSbwGLqMJM9

---
_Generated by [Claude Code](https://claude.ai/code/session_01AEqEB4zXiPhLSbwGLqMJM9)_

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -26,6 +26,12 @@ updates:
       - dependency-name: "x11rb"
         update-types:
           ["version-update:semver-minor", "version-update:semver-major"]
+      # winit's default set is copied verbatim into crates/pixtuoid/Cargo.toml and
+      # cargo cannot subtract one default, so a bump ADDING a default silently
+      # drops a backend — a Linux runtime failure no gate catches. Bump by hand.
+      - dependency-name: "winit"
+        update-types:
+          ["version-update:semver-minor", "version-update:semver-major"]
     cooldown:
       default-days: 7
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1201,6 +1201,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "77ce24cb58228fbb8aa041425bb1050850ac19177686ea6e0f41a70416f56fdb"
 
 [[package]]
+name = "font-types"
+version = "0.12.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e64eb721ca85a34323425f4041adc5d82704d3782d5f8f03793bc012419dce23"
+dependencies = [
+ "bytemuck",
+]
+
+[[package]]
 name = "foreign-types"
 version = "0.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1738,14 +1747,14 @@ checksum = "b6d2cec3eae94f9f509c767b45932f1ada8350c4bdb85af2fcab4a3c14807981"
 
 [[package]]
 name = "libredox"
-version = "0.1.19"
+version = "0.1.23"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2026a5056764a10b2bf5d56488cba40da507f5493a6a429340e2004d9ed085fa"
+checksum = "8d8f1ea3f21fd3405dcaf6c9b5c1630af9afc422d9073ea39c5f6d6c772e08ed"
 dependencies = [
  "bitflags 2.13.1",
  "libc",
  "plain",
- "redox_syscall 0.9.1",
+ "redox_syscall 0.9.3",
 ]
 
 [[package]]
@@ -2676,7 +2685,7 @@ checksum = "a89322df9ebe1c1578d689c92318e070967d1042b512afbe49518723f4e6d5cd"
 name = "pixtuoid"
 version = "0.18.0"
 dependencies = [
- "ab_glyph",
+ "ab_glyph_rasterizer",
  "anyhow",
  "chrono",
  "clap",
@@ -2697,6 +2706,7 @@ dependencies = [
  "schemars",
  "serde",
  "serde_json",
+ "skrifa",
  "snapbox",
  "softbuffer",
  "tempfile",
@@ -3161,6 +3171,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "read-fonts"
+version = "0.43.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "005c8acf251756c478b0bf402885bfd88a1476020c4c7e6060edc1aa68da38ea"
+dependencies = [
+ "bytemuck",
+ "font-types",
+ "once_cell",
+]
+
+[[package]]
 name = "redox_syscall"
 version = "0.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3180,9 +3201,9 @@ dependencies = [
 
 [[package]]
 name = "redox_syscall"
-version = "0.9.1"
+version = "0.9.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "07507be7b4a5f9f26eeb41eeaebb1f5a7ff29dfb29739facc21d35bf8b11c21e"
+checksum = "d678d17679829e73d371e96880897e98fee2ded7acc0a50bdf8af2affa4b2fe5"
 dependencies = [
  "bitflags 2.13.1",
 ]
@@ -3580,6 +3601,16 @@ name = "siphasher"
 version = "1.0.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8ee5873ec9cce0195efcb7a4e9507a04cd49aec9c83d0389df45b1ef7ba2e649"
+
+[[package]]
+name = "skrifa"
+version = "0.46.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d4febebda507fb889c545a37a135517769d1391941013561292897961d1887d6"
+dependencies = [
+ "bytemuck",
+ "read-fonts",
+]
 
 [[package]]
 name = "slab"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,16 +3,6 @@
 version = 4
 
 [[package]]
-name = "ab_glyph"
-version = "0.2.32"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "01c0457472c38ea5bd1c3b5ada5e368271cb550be7a4ca4a0b4634e9913f6cc2"
-dependencies = [
- "ab_glyph_rasterizer",
- "owned_ttf_parser",
-]
-
-[[package]]
 name = "ab_glyph_rasterizer"
 version = "0.1.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2490,15 +2480,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1a80800c0488c3a21695ea981a54918fbb37abf04f4d0720c453632255e2ff0e"
 
 [[package]]
-name = "owned_ttf_parser"
-version = "0.25.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "36820e9051aca1014ddc75770aab4d68bc1e9e632f0f5627c4086bc216fb583b"
-dependencies = [
- "ttf-parser",
-]
-
-[[package]]
 name = "palette"
 version = "0.7.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3419,9 +3400,7 @@ version = "0.10.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b6277f0217056f77f1d8f49f2950ac6c278c0d607c45f5ee99328d792ede24ec"
 dependencies = [
- "ab_glyph",
  "log",
- "memmap2",
  "smithay-client-toolkit",
  "tiny-skia",
 ]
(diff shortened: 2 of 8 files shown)


IvanWng97:
## two-lens-review — diff scope, round 1

Five differentiated lenses over `fc480bc3..HEAD` in an isolated worktree: correctness, design/blast-radius, whole-file comment audit (the always-fires row), behavior-preserving-refactor (adversarial toward revert), and visual. Churn 270 added — inside the budget. **Zero open confirmed HIGH.**

### The one real mover

The adversarial lens found what the other four missed: `ab_glyph` deliberately splits the pen position before rounding —

```rust
// ab_glyph-0.2.32/src/outlined.rs, px_bounds()
// "Use subpixel fraction in floor/ceil rounding to eliminate rounding error"
let (x_trunc, x_fract) = (position.x.trunc(), position.x.fract());
(min.x * scale_factor.horizontal + x_fract).floor() + x_trunc
```

— and this port had collapsed it to `pen + offset`. Equal in exact arithmetic, not in f32: wherever `pen + coord·k` lands within one ulp of an integer the offset's low bits vanish, on **both** axes (so it reaches `pen_x` 0 via the baseline term), in ~0.0156% of combinations. Concretely at `pen_x` 2047.6804 with a scaled offset of 6.3196507 the exact sum 2054.0000705719 rounds to `2054.0`, so `ceil` yields 2054 where the split yields 2055. Measured divergence before the fix — max coverage delta `1.77e-5`, 250,034 pixels lit only by the old path, 15,541 of 99,553,428 swept origins disagreeing — all provably inert at 8 bits (`8.23e-5 × 255 = 0.021` vs a 0.5 flip threshold). `gen-check` compares rendered u8 bytes, so its exit 0 is not evidence the mover was absent — it is confirmation the mover was invisible to it. **Restored**, so the identity is exact rather than merely bounded. `gen-check` re-run after the change: exit 0.

The PR body's "max f32 coverage delta `0.000e+00`" was measured on 4 sizes × 6 origins, none past ~1024, and is false as written. Body corrected rather than deleted.

### Dispositions

| Finding | State |
|---|---|
| Pen split dropped (behavior mover) | **FIXED** — upstream arithmetic restored, gen-check exit 0 |
| `OutlineCollector` doc: "control bounds are looser than the true outline box" | **FIXED** — false for this face. It is OTTO/CFF, and ttf-parser's CFF builder extends its bbox by both control points *and* the endpoint (`tables/cff/mod.rs`), so ab_glyph measured the *same* control-point box |
| `px_per_unit` doc restated "1000 units against 1155" | **FIXED** — the face is `unitsPerEm` 2000, typo ascent 1890 / descent −400, so ascent−descent is 2290. Both values were half the truth. Deleted, not corrected: the code reads `FACE.v_metrics` at runtime and never the literals |
| `OutlineCollector::close`'s closing edge had no test | **FIXED** — mutation-verified. Removing the `push` floods coverage to 15.301 and fails the new test, while all five pre-existing `aa_text` tests still pass. `a` and `M` (the glyphs the old tests draw) have already-closed contours, and `encode.rs`'s `★` test clamps before asserting |
| `static FONT` redundant | **FIXED** — skrifa's provider methods bind to the font *data* lifetime, not `&self`, so a local `FontRef` yields the same `'static` fields. One `LazyLock` instead of two |
| ttf-parser/#440 story told in four places; dead `[OutlineGlyph::draw]` link; dev-narration; restated counts | **FIXED** — −27 lines; story reduced to one home |
| Four unresolved private-item intra-doc links | **FIXED** — `--document-private-items` now exits 0 (was 4 errors). `NullSink` is `cfg(test)`-only so it can never be a link target; `new`/`render` needed `Self::` |
| winit default-set drift had no mechanism | **FIXED** — dependabot now holds winit at minor/major, the same discipline `objc2-app-kit` and `x11rb` already carry for their winit coupling |
| `aa_text` module doc named 3 of its 4 consumers | **FIXED** — `examples/cutaway_snapshot.rs` added |
| `px_per_unit` needs a metric test tooth | **REFUTED** — a mechanism already exists: `[[example]] snapshot` has `test = true`, so `cell_font_px_fits_the_cell`'s `assert_eq!(line_height(CELL_FONT_PX), CELL_H)` is an exact equality running in the normal suite, not gen-check-only |
| "no glyph cache" / allocation churn | **REFUTED** — ab_glyph allocated identically (one curve `Vec` + one rasterizer buffer per glyph). Worst case measured at 2.8% of the frame budget (580 glyphs/frame) |
| `ab_glyph_rasterizer` trades one unmaintained crate for another | **REFUTED** — 0.1.10, repo commits through 2026-08-30, exactly one optional dependency (`libm`), no RUSTSEC row |

### Evidence

- **Independent reproduction of the identity claim**: 404,354 non-zero pixels compared across both implementations pinned to the real face — 0 value mismatches at 8-bit; `has_glyph` 0 disagreements over U+0000–U+1FFFF; `text_width`/`line_height` 0 disagreements.
- **Renders**: 9 head-vs-base pairs byte-identical, including `floating_snapshot` — the `aa_text` consumer `gen-check` does not cover at all. Binaries confirmed genuinely different first (head: 21 `skrifa` strings / 0 `ttf-parser`; base: 0 / 5). Symbol vocabulary rasterized and read, not merely `has_glyph`-asserted; every glyph a distinct real shape, `U+00A0` correctly yielding an empty box.
- **Gates** run locally, exit codes observed unpiped: `cargo deny check advisories` and `bans licenses sources` on cargo-deny **0.20.2** (the CI pin) → 0 with `ignore = []`; `cargo machete` → 0; `TZ=UTC gen-media.py --check` → 0; `cargo doc --document-private-items -D warnings` → 0; full `just preflight` via pre-push → 3,252 tests, 0 failures.

### Surfaced — owner's call, not grown into this PR

- `doc-check` runs `--workspace` without `--document-private-items`, so rustdoc never processes a private item's docs and no broken link on one can red the gate. That class is now empty, so adding the flag would be a clean follow-up rather than a red one.
- 4 open high dependabot alerts on the default branch, all npm `fast-uri` in `site/package-lock.json` — unrelated to this PR's Rust graph.

https://claude.ai/code/session_01H6jVkwU39et5UTDdFfKJYU

PR #987, opened 2026-09-10
fix(drift): repin the dsh session anchor, ledger codex retained_context and grok hook_run_started

### LINKED ISSUE #981: New upstream events to review
# New upstream events to review

## 🔎 New upstream events to review
- upstream `RolloutItem` has `retained_context`, a sibling of the `turn_context` we already decode, and source/codex.rs decodes it to nothing. Decode it, or add it to CODEX_KNOWN_OMITTED with the reason.

## 🩺 Probe could NOT verify — this is not evidence of upstream change

A lookup missed, so all the watcher knows is that **its own probe failed**. A stale pin, a moved declaration, or a refactor that left a re-export facade at the old path all land here, and every check riding the affected surface was SKIPPED rather than reported as drift. **Verify upstream by hand and repin — do NOT change a decoder on these lines alone** (#793 did, against three working env vars).

- could not verify dsh session event types: the document no longer contains `assistant/chunk` at https://raw.githubusercontent.com/deepseek-ai/deepseek-harness/master/packages/core/session/src/types.ts — Upstream may have moved or reshaped it, or our pin/parser may be stale; this is NOT evidence that upstream changed. It still fetches, so the probe landed on the wrong content — most likely a stale pin (a moved declaration, or a re-export facade left at the old path) rather than an upstream rename. Every presence check riding this document was SKIPPED, NOT reported as drift. Re-verify upstream by hand, then repin — do NOT change a decoder on this alone.

IvanWng97:
## Summary

Both items of #981, plus one that landed after it was filed. No decoder changes — every fix is in the drift watcher's own ledger/anchors, and `check_upstream_drift.py` now exits 0 against live upstream.

## Related issue

Closes #981

## Type of change

- [x] Refactor / chore

## What each item turned out to be

**dsh `assistant/chunk` — a stale watcher anchor, not a broken plugin.** This was the trap the issue warned about, and it was a real upstream reshape rather than a moved declaration: dsh's session format v2 dropped `assistant/chunk` as a top-level event (`.agents/notes/implemented/architecture/2026-09-01-v2-embedded-assistant-streams.md`: *"Session format v2 has no top-level `assistant/chunk` event"*), and `packages/core/session/src/types.ts:88` now declares `SESSION_FORMAT_VERSION = 3`. The bundled `dsh_plugin.mjs` was never affected — it listens for `assistant/message`, the v2 replacement — and all nine of its events still resolve across the four pinned docs (checked by hand). The watcher's anchor aborted that check before it could run. Repinned to the declaration that owns the keys, `export interface SessionEventMap {` (`types.ts:269`), matching the owner-grade form the Codex rows use; an event-key anchor is a sibling of the names it guards and can vanish independently, which is exactly what happened. The v3 canonical-envelope note adds `surfaceOp` to `assistant/message` / `tool/result` — additive; `dsh.rs` has a `_ =>` arm and reads none of it.

**Codex `retained_context` — ledgered, not decoded.** `RolloutItem` gained `RetainedContext(RetainedContextEvent)` (`history/src/lib.rs:134`). Upstream's own doc reads *"Sparse, model-invisible updates. Only the host may produce these records."* (`history/src/retained_context.rs:124`) — the same reason `realtime_item` is already in `CODEX_KNOWN_OMITTED`. Sole variant `verified_answer` is authorization evidence: no lifecycle, activity, model, or token signal. The `_context` kinship with `turn_context` is lexical.

**grok `hook_run_started` — ledgered (new since the issue).** `SessionUpdate` gained `HookRunStarted` (`xai-grok-shell/src/extensions/notification.rs`): grok's own pager spinner phase — *"the turn is blocked on an awaited hook batch that just started"*. The reason it is ledgered rather than decoded, fetched from the whole upstream tree: upstream emits it through `send_xai_notification_transient` (`session/acp_session_impl/hook_dispatch.rs`), whose body only `forward_fire_and_forget()`s to the live client and never appends to `updates.jsonl` — unlike `send_xai_notification`, which appends with `AppendDurability::Buffered`. This decoder reads `updates.jsonl`, so the tag cannot appear in it. (The review fold replaced an earlier reason — that decoding the start would strand the session Active — which the repo's own `turn_completed → ActivityEnd` test refutes.)

Owner rule applied throughout: only the latest version of each CLI is supported, so there is no dual anchor for the v1 dsh name and nothing to fall back to.

## How I tested it

- `just drift-selftest`: 0 (baseline on main was also 0).
- `just drift-selftest` and live `python3 scripts/check_upstream_drift.py` re-run after the review fold (`a143060e`): 0 and **0** — `No drift. Every name we depend on is present upstream.` (was 1 with the three items).
- Every upstream `path:line` above was fetched this session, not recalled; the dsh search was over the whole upstream tree via code search, not one plausible file.

## Surfaced — owner's call, not grown into this PR

- `DSH_RUNTIME_TYPES_URL` (`'agent/pre-step'`) and `DSH_SESSION_INDEX_URL` (`'session/created'`) are the same fragile event-key anchor class that just bit here. Both still resolve today. Converting them is **not** a small follow-up: their upstream owners are ambient `declare module '@deepseek-ai/cordis' { interface Events {` augmentations with no unique exportable name, so an owner-grade anchor needs a multi-line pattern — a gate-shape change, its own PR. Related pre-existing nit: those two rows plus `ApprovalRequestId` sit under the `# owner-grade:` label at `check_upstream_drift.py:299` but are identity-grade by the file's own definition; moving them under the `# identity-grade:` label is a pure reorder.
- A `*_KNOWN_OMITTED` key that upstream later removes is never reported (the sweeps compute `upstream - mine - ledger`), so a dead ledger row is inert forever. A stale-ledger guard is its own small PR.
- Candidate one-line bullet for `CLAUDE.md` Conventions: *Only the latest released version of each agent CLI is supported — no fallback arms, no dual-listening for an old event name beside its replacement.*
- grok's `verified_version` lagged the installed CLI (0.2.102 vs 1.0.25); re-recorded in #988.

## Visual verification

- [x] Not a visual change.

## Checklist

- [x] New behavior has a test — the selftest fixtures embed the new anchor; the ledger entries are exercised by the existing sibling-sweep tests.
- [x] Docs updated in the same commit — the ledger values are their own WHY.
- [x] Checked against the recurring pitfalls (#793: do not change a decoder on a probe failure alone — none changed).

## AI assistance

### DIFF
--- a/scripts/check_upstream_drift.py
+++ b/scripts/check_upstream_drift.py
@@ -569,6 +569,11 @@ GROK_XAI_KNOWN_OMITTED: dict[str, str] = {
     "subagent_progress": "cumulative per-child totals (turns, tool calls); "
     "summing a running total into the delta-accumulating reducer double-counts "
     "(codex's `token_count_emits_fresh_usage_from_last_reading` is the precedent)",
+    "hook_run_started": "grok's pager spinner phase — 'the turn is blocked on "
+    "an awaited hook batch that just started' (notification.rs:571-582); its "
+    "END is `hook_execution`, which decodes to nothing except session_end, so an "
+    "ActivityStart here would strand the session Active (codex's "
+    "`local_shell_call` precedent)",
 }
 
 
--- a/scripts/check_upstream_drift.py
+++ b/scripts/check_upstream_drift.py
@@ -555,7 +555,7 @@ CODEX_KNOWN_OMITTED: dict[str, str] = {
     "Started/Closed bracket the voice layer, speech segments are content no "
     "decoder reads, and `BemItemPromoted` re-presents items already decoded "
     "from `response_item` (upstream: 'sparse, model-invisible facts')",
-    "retained_context": "host-only, model-invisible guardian records "
+    "retained_context": "host-only, model-invisible records "
     "(history/src/retained_context.rs: 'Sparse, model-invisible updates. Only "
     "the host may produce these records.'); the sole variant `verified_answer` "
     "is authorization evidence — no lifecycle, activity, model or token signal. "
@@ -569,11 +569,12 @@ GROK_XAI_KNOWN_OMITTED: dict[str, str] = {
     "subagent_progress": "cumulative per-child totals (turns, tool calls); "
     "summing a running total into the delta-accumulating reducer double-counts "
     "(codex's `token_count_emits_fresh_usage_from_last_reading` is the precedent)",
-    "hook_run_started": "grok's pager spinner phase — 'the turn is blocked on "
-    "an awaited hook batch that just started' (notification.rs:571-582); its "
-    "END is `hook_execution`, which decodes to nothing except session_end, so an "
-    "ActivityStart here would strand the session Active (codex's "
-    "`local_shell_call` precedent)",
+    "hook_run_started": "never in the transcript: upstream emits it through "
+    "`send_xai_notification_transient` (session/acp_session_impl/hook_dispatch.rs), "
+    "which only forwards to the live client and never appends to updates.jsonl, "
+    "the file this decoder reads. A pager spinner phase — 'the turn is blocked on "
+    "an awaited hook batch that just started' (extensions/notification.rs) — whose "
+    "end `hook_execution` decodes to nothing but session_end",
 }
 
 
--- a/scripts/check_upstream_drift.py
+++ b/scripts/check_upstream_drift.py
@@ -302,7 +302,7 @@ ANCHORS: dict[str, Anchor] = {
     CODEX_PROTOCOL_URL: Anchor(r"pub enum HookEventName\b", "`HookEventName`"),
     CODEX_ROLLOUT_ITEM_URL: Anchor(r"pub enum RolloutItem\b", "`RolloutItem`"),
     DSH_RUNTIME_TYPES_URL: Anchor(r"'agent/pre-step'", "`agent/pre-step`"),
-    DSH_SESSION_TYPES_URL: Anchor(r"'assistant/chunk'", "`assistant/chunk`"),
+    DSH_SESSION_TYPES_URL: Anchor(r"export interface SessionEventMap\b", "`SessionEventMap`"),
     DSH_APPROVAL_TYPES_URL: Anchor(r"ApprovalRequestId", "`ApprovalRequestId`"),
     DSH_SESSION_INDEX_URL: Anchor(r"'session/created'", "`session/created`"),
     GROK_HOOK_URL: Anchor(r"pub enum HookEventName\b", "`HookEventName`"),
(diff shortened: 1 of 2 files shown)


IvanWng97:
## two-lens-review — diff scope, round 1

Four differentiated lenses over `07911cf9..7de4f9bb` in an isolated worktree (correctness ×2, design + whole-file comment audit ×2 — the second pair re-run after the first stalled; all four eventually reported). Churn +13/−3. All four verdicts **APPROVE-WITH-NITS**; **zero HIGH**. Every upstream claim was fetched this session by each lens independently and confirmed: dsh `SessionEventMap` (`types.ts:269`, unique, owns all nine checked keys; `'assistant/chunk'` gone), codex `RetainedContextEvent` (`retained_context.rs:124`, sole variant `verified_answer`), grok `HookRunStarted`/`HookExecution` (`notification.rs:573/:584`). Gates: `just drift-selftest` 0, live `check_upstream_drift.py` 0, on every lens's run and again after the fold.

### Fold — `a143060e`

| Finding | Lens | State |
|---|---|---|
| `hook_run_started` ledger claimed decoding the start "would strand the session Active" | design ×2, correctness | **FIXED** — false for grok: `turn_completed` decodes to `ActivityEnd` (`grok.rs:452`, pinned by `turn_completed_settles_the_turn_to_idle`), and upstream says the phase ends on "HookExecution *or later turn output*". Replaced with the reason that holds, verified from the whole upstream tree: upstream emits `HookRunStarted` via `send_xai_notification_transient` (`session/acp_session_impl/hook_dispatch.rs`), which only `forward_fire_and_forget()`s and never appends to `updates.jsonl`; the decoder reads `updates.jsonl`, so the tag cannot appear |
| Same entry cited `notification.rs:571-582` | correctness ×2, design | **FIXED** — a raw line range into upstream `main` is a restated value that rots; sibling entries cite path only |
| Same entry's `local_shell_call` parenthetical | design ×2 | **FIXED** — an echo of the ledger's block comment, importing a mechanism that does not hold here |
| `retained_context` says "guardian" | design | **FIXED** — not upstream's word; the variant is authorization evidence |
| Anchor regex robustness / circularity | design ×2 | **REFUTED** — `SessionEventMap` is not a checked name (`script:1572-1573`); `\b` survives generics/`extends`; the `declare interface` reshape would file as *blind*, the designed #793 mode — mechanism: `test_every_swept_url_declares_an_anchor` + `ANCHOR_SAMPLES` |
| Selftest fixtures should keep `'assistant/chunk'` as a negative control | design | **REFUTED** — the dead key is inert to both tests; the real negative control (no anchor → blind) is untouched, and composing a legacy name in is exactly what the owner's latest-only rule bans |
| Sibling event-key anchors left in place — plan-miss? | correctness ×2, design ×2 | **SURFACED** — correct scoping. Not a small follow-up: their upstream owners are ambient `declare module '@deepseek-ai/cordis' { interface Events {` augmentations with no unique exportable name (runtime-types.ts:245, index.ts:33); an owner-grade anchor needs a multi-line pattern — a gate-shape change, its own PR. Both resolve today (runtime-types.ts:330, index.ts:50) |
| Three identity-grade dsh rows sit under the `# owner-grade:` label (`:299`) | design ×2 | **SURFACED** — pre-existing since #971; pure reorder |
| A `*_KNOWN_OMITTED` key upstream later removes is never reported (`upstream - mine - ledger`) | correctness ×2, design | **SURFACED** — pre-existing; a stale-ledger guard is its own small PR |
| `Anchor` docstring's "never a checked name" has no selftest mechanism | correctness | **SURFACED** — pre-existing |
| Latest-CLI-only rule is prose-only | design ×2 | **SURFACED** — candidate `CLAUDE.md` Conventions bullet in the PR body; not added |

Test-teeth (correctness): reverting the anchor to `'assistant/chunk'` fails the selftest and re-blinds the live run; both ledgers have fires-on-gap / silent-when-ledgered selftests.

PR #989, opened 2026-09-10
test(fixtures): re-record twelve headless scenarios at the installed CLIs

IvanWng97:
## Summary

Re-records twelve headless fixture scenarios off the wire at the CLI versions actually installed, and bumps each source's `verified_version` to match. Owner rule applied: only the latest version of each agent CLI is supported. **No decoder changed** — every decoder decodes its new bytes as-is, which is the finding.

## Related issue

Follow-up to #987 (the drift watcher is clean at the name level; this re-verifies the recorded bytes) and #988 (grok, same treatment).

## Type of change

- [x] Refactor / chore

## What was re-recorded

| source | verified | scenario(s) |
|---|---|---|
| claude-code | 2.1.251 → **2.1.261** | tool-run-recorded |
| codex | 0.147.0 → **0.153.4** | tool-run-recorded (the one pinned numeric drift `fixture-age` reported) |
| cursor | 2026.08.11 → **2026.09.02** | tool-run, tool-failure — needed `--model auto` (see below) |
| hermes | 0.20.1 → **0.21.0** | tool-run |
| kimi | 0.36.0 → **0.40.1** | tool-run |
| opencode | 1.18.15 → **1.18.20** | tool-run |
| reasonix | 1.25.2 → **1.37.0** | tool-run |
| omp | 18.0.11 → **18.1.10** | tool-run-recorded, bridge-run-recorded, task-recorded, resume-recorded |

Each is `just capture-fixture <source> <scenario> <cmd>` with the command its old provenance recorded (omp `task` with its subagent prompt, omp `resume` with `-c` after another omp run). One billed model turn each.

**cursor needed `--model auto`.** The first two `cursor-agent -p --trust` runs died before any tool call — `ActionRequiredError: Named models unavailable. Free plans can only use Auto.` — and would have replaced real tool-run / tool-failure evidence with a degenerate run, so they were discarded and re-recorded with `--model auto`; the command is recorded in provenance. The retries carry the real `preToolUse`/`postToolUse` and `postToolUseFailure` payloads.

## Golden review

Nine conformance goldens moved. Each was diffed at the event-header level; **no lifecycle event is lost anywhere**:

- claude-code, codex: one fewer assistant / token-count turn in this run (run variance).
- cursor: finished tool-run in four calls instead of six and tool-failure in one instead of three; the per-call shape and the failure path are intact.
- hermes, kimi: one adjacent payload pair arrives in swapped order.
[...]

PR #988, opened 2026-09-10
test(grok): re-record tool-run at 1.0.25, bump verified_version

IvanWng97:
## Summary

Re-verifies the grok decoder against grok **1.0.25** (the installed CLI; the fixture and `registry.rs` were pinned at 0.2.102) by re-recording the headless `tool-run-recorded` scenario off the wire and bumping the verified version. No decoder code changes: the 1.0.25 bytes decode with the decoder as-is, which is the point of the exercise.

## Related issue

Follow-up surfaced in #987 (grok `verified_version` lagged the installed CLI).

## Type of change

- [x] Refactor / chore

## What was recorded

`just capture-fixture grok tool-run-recorded grok -p '{prompt}'` — one billed model turn against the real CLI, default prompt. The recorder halted at its PII gate (as designed: it embeds the recorder's own home path), so the bytes were redacted `/Users/navepnow → /Users/dev` per `fixtures/README.md` and the provenance promoted from `provenance.json.new` with the standard redaction note. The 0.2.102 session dir is removed; one recorded scenario per version.

Two things the new capture carries that the 0.2.102 one did not:
- `hook-payloads.jsonl` — the hook-side record (11 payloads) beside the transcript, so the fixture now covers both seams the recorder captures.
- Every key in both spellings — `hookEventName`+`hook_event_name`, `toolUseId`+`tool_use_id`, `transcriptPath`+`transcript_path` — on all 11 payloads. The pre-1.0 interactive captures carry camelCase only, so this is new at 1.0.x; the decoder reads the camelCase envelope and never read the snake twins, but the `tool_use_id` twin is what the tool-id census pins. (An earlier draft of this bullet called the session path "absent at 0.2.102" — that was recalled, not fetched: 0.2.102 had no hook capture, and the pre-1.0 captures do carry `transcriptPath`.)

Event census of the new hook record: SessionStart, UserPromptSubmit, PreToolUse ×3, PostToolUse ×3, Stop ×2, SessionEnd — the same lifecycle shape as before.

## Bumped in step

- `crates/pixtuoid-core/src/source/registry.rs` — `verified_version: "1.0.25"`.
- `crates/pixtuoid-core/tests/sources/captures.rs` and `scripts/fixture-age.py` — the banner-parse cases now use the 1.0.25 banner (`grok 1.0.25 (f7e67d6988e2) [stable]`), same shape, new values.

## Not done here (needs a human at the keyboard)

[...]

PR #995, opened 2026-09-12
test(fixtures): a capture that LISTS an inventory fails, because no keyword could

IvanWng97:
## Summary

The operator's installed plugin roster reached committed bytes **three times** (#988, #989, #993) and nothing saw it any of those times. This is the gate that would have.

## Related issue

Surfaced in #992's and #993's review threads, with the measurement.

## Type of change

- [x] New test / CI gate

## Why no keyword could do it

#992 tried exactly that and had to delete all three markers it added, because each fired on the CLIs' **own stock bytes**:

| marker | where it fired with no operator present |
|---|---|
| `AGENTS.md` | codex's system prompt — "applicable `AGENTS.md` instructions" |
| `userId` | copilot's built-in `edit` tool docs — `let userId = guid();` |
| `CLAUDE.md` | a captured `ls -la` of the workspace |

Five committed fixtures went from refusing nothing to refusing unconditionally, and no redaction could clear them. A gate that cannot be satisfied trains you to ignore it.

## What separates a roster from prose is cardinality

A stock prompt mentions `SKILL.md` once; a roster lists eighty. Measured over the nine files #993 fixed, before and after:

| carrier | before | after |
|---|---|---|
| copilot `<skill>` elements | 34 | 2 |
| `SKILL.md` roster rows | 85 | 8 |
| `name: description` rows | 112 | 10 |

`MAX_SIBLING_ENTRIES = 16` is bounded from both sides: the redacted tree's worst line carries **10** (codex's five bundled skills plus placeholders), and the smallest of the three leaks carried **34**.

## I tried gitleaks first, and it cannot express this

Real entries run **248–290 characters at the median** (max 1610), and Go's RE2 refuses `{16}` over a window that size. Seven repetition/window combinations — `{16}`×60 through `{4}`×500 — all failed to compile, and **gitleaks panics on a rule it cannot compile rather than reporting it**, so a bad rule there is worse than none. A short toy sample made it look possible; that was my own test fixture, not the data.

So the check lives where JSON is legible, in `captures.rs`, which already walks every capture's wire files.

## How I tested it

- Applied the test's own counters to the **pre-redaction bytes read from git** (`git show 209554cc:<path>`): **8 of the 9 files fire**. The ninth is `decode/mod.rs`'s hand-written test literal, not a roster.
- The redacted tree passes: `cargo test -p pixtuoid-core --test sources` 98 → 99, `just test` 3260 passed.
- `just lint` green, `just clippy` green (the carrier array needed a `type Carrier` alias for `type_complexity`).
- The `name: description` counter was deliberately narrowed after a first version flagged 90 prose bullets in claude-code's own system prompt: a roster row carries an *identifier* left of the colon and a capitalised description right of it; a prose bullet does not.

## Surfaced — owner's call

- `just fixture-age` reports reasonix already at 1.37.0 → 1.38.7 and omp 18.1.10 → 18.1.18, two days after #989 re-recorded them. Chasing versions by hand is Sisyphean; the real gap is that nothing *tells* you a re-record is due — `fixture-age` is advisory and local-only.

## Visual verification

- [x] Not a visual change.

## Checklist

- [x] I read the relevant `CLAUDE.md` (root + core + `tests/CLAUDE.md`).
- [x] New behavior has a test — the test IS the behavior, and it is verified against the real bytes it would have caught.
- [x] No `unwrap()` in non-test code. No new `println!`/`eprintln!`.
- [x] Docs updated in the same commit — the test's own doc comment carries the why.

## AI assistance


IvanWng97:
Closing unmerged, as intended: this PR was kept open as the evidence that the blocklist path was the wrong one. Its five cardinality rules counted roster entries in the bytes; #997 (capture-time allowlist derived by probing the decoders) and #998 (the committed corpus re-stripped) remove those fields structurally, so the rules have no bytes left to count. The one capture exempt from the strip (`cursor/fixtures`, pinned in `captures.rs`) is still scanned by the identity rules, which stay. The dead placeholder sentinels this branch also carried are deleted in the follow-up PR.

PR #996, opened 2026-09-12
chore(publish): keep the recorded fixture corpus out of the crates.io tarball

IvanWng97:
## Summary

The recorded fixture corpus — 176 files of captured CLI wire bytes — shipped inside the published crates.io tarball. This removes it with one `exclude` entry.

## Why it matters

Those fixtures carry an operator's fingerprint: installed plugin/skill/agent rosters, global-instruction attachments, home paths, git identity. Three re-records (#988/#989/#993) leaked one of these classes each, caught and redacted after the fact. **A crates.io tarball is immutable** — the repo copy can be redacted in a later commit, the published `.crate` cannot. Keeping the corpus out of the tarball removes the one irreversible distribution channel.

## What changed

One line: `"tests/sources"` added to `pixtuoid-core`'s `exclude`. It removes the fixture **data** and the grouped `sources` test binary (`tests/sources/main.rs`) that reads it — cargo autodiscovers that binary, and both shipped on base.

## How I verified it

- `cargo package --list`: `tests/sources/` entries in the tarball go **176 → 0**. Nothing else drops: `src/`, benches, examples, and the other packaged test binaries (`reducer`, `render`, `transport`, `watcher`, `e2e.rs`) all still ship.
- `cargo package` with the verify build: **exit 0** — the extracted crate compiles. This is the check the `exclude` block's own comment exists for.
- No `include_str!`/`include_bytes!` points under `tests/sources/`, so no compile-time break.
- `just lint` green, `just test` 3259 passed; every workspace/CI test reads the git checkout, which is untouched. Homebrew's formula asserts the binary's CLI output; the site build reads none of this.

**One packaged reader survives**, surfaced by review and now recorded on the exclude: `tests/watcher/sources.rs:729` reads the codex `permission-flow` fixture, so `cargo test` on the *extracted* crate fails there. Latent — no gate runs the extracted crate's tests, and that suite is already red from `tests/render/format.rs`'s reads outside the crate. Excluding a real test binary to chase it would be the wrong fix; left as an owner call.

## Scope

This is the **irreversibility** half of the de-identification work. The **in-repo** exposure (the corpus is still readable in the public repo) is handled separately by deriving a capture-time allowlist from the decoders — a follow-up PR, deliberately not bundled here because it touches the recorder, not the publish config.

## Type of change

- [x] Chore (publish configuration)

## Checklist

- [x] `cargo package` verify-build passes.
- [x] No behavior change to the library or tests; `just test` 3259 pass.

## AI assistance

### DIFF
--- a/crates/pixtuoid-core/Cargo.toml
+++ b/crates/pixtuoid-core/Cargo.toml
@@ -16,11 +16,12 @@ categories.workspace   = true
 # but not `cargo test` on an extracted crate. They stay FLAT so they can be
 # excluded at all: a grouped binary's submodule cannot be excluded on its own.
 exclude                = [
-    # The recorded fixture corpus is a test-only input to the grouped `sources`
-    # binary (`tests/sources/main.rs`), which its `/`-in-path already keeps out of
-    # the tarball; the DATA it reads was still shipping. It carries captured wire
-    # bytes, so an immutable crates.io copy is the wrong home for it — the corpus
-    # lives in the public repo, not the published crate.
+    # The recorded fixture corpus is captured wire bytes — an operator's plugin
+    # roster, instructions, home paths — and a crates.io tarball is immutable
+    # where the repo copy can still be redacted. The grouped `sources` binary
+    # that reads it goes with it. `tests/watcher/sources.rs` still ships and
+    # reads one scenario, so the extracted crate's `cargo test` fails there —
+    # as it already does at `tests/render/format.rs`'s outside-crate reads.
     "tests/sources",
     "tests/pinned_by_claims.rs",
     "tests/socket_path_parity.rs",
--- a/crates/pixtuoid-core/Cargo.toml
+++ b/crates/pixtuoid-core/Cargo.toml
@@ -16,6 +16,12 @@ categories.workspace   = true
 # but not `cargo test` on an extracted crate. They stay FLAT so they can be
 # excluded at all: a grouped binary's submodule cannot be excluded on its own.
 exclude                = [
+    # The recorded fixture corpus is a test-only input to the grouped `sources`
+    # binary (`tests/sources/main.rs`), which its `/`-in-path already keeps out of
+    # the tarball; the DATA it reads was still shipping. It carries captured wire
+    # bytes, so an immutable crates.io copy is the wrong home for it — the corpus
+    # lives in the public repo, not the published crate.
+    "tests/sources",
     "tests/pinned_by_claims.rs",
     "tests/socket_path_parity.rs",
     "tests/supported_sources_manifest.rs",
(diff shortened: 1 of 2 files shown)


IvanWng97:
**Review disposition** (one lens, re-derived the tarball contents from `cargo package` rather than trusting the body):

| Finding | State |
|---|---|
| MEDIUM — the exclude comment (and the first commit message) claimed the grouped `sources` binary was already out of the tarball via its `/`-in-path. Cargo autodiscovers `tests/sources/main.rs`; it shipped on base, and this exclude is what removes it. | **FIXED** in 1d3f9c34 — the comment now states the real mechanism; the PR body no longer makes the claim. The squash message should be taken from the body, not the first commit. |
| LOW — `tests/watcher/sources.rs:729` ships and reads the now-excluded codex `permission-flow` fixture, so `cargo test` on the *extracted* crate panics there. | **SURFACED** — recorded on the exclude entry itself so nobody "fixes" it by excluding a real test binary. Latent: no gate runs the extracted crate's tests, and that suite is already red from `tests/render/format.rs`'s outside-crate reads. Owner call whether extracted-crate `cargo test` is a goal at all. |

Verified myself before editing: base `Cargo.toml` has no `tests/sources` entry (so the binary shipped), and the packaged test set is `reducer`/`render`/`transport`/`watcher`/`e2e.rs`. Functional change unchanged: 176 → 0 fixture files in the tarball, verify build exit 0.

IvanWng97:
**Review round 2 — second lens (design / blast-radius) disposition**, fold in the commit after 1d3f9c34:

| Finding | State |
|---|---|
| The comment's "home paths" overstates the WHY (the recorder already redacts them to `/Users/dev`); the defensible part is crates.io immutability. Framing read as chasing a green extracted-crate `cargo test`, which no pipeline runs. | **FIXED** — comment rewritten: a tarball-contents choice, not a green-test one. |
| `tests/CLAUDE.md`'s "FLAT + publish-excluded" line is now non-exhaustive. | **FIXED** — names `sources/` as excluded whole. |
| `exclude` is a fail-open denylist with no tarball-contents gate; an `include` allowlist would withhold a future `tests/<grouped>/` by default. | **SURFACED** — a separate PR; the shippable set is small and stable (`src/**`, the declared example and bench). Not this PR's scope. |
| The pre-existing block header (`Cargo.toml:13-17`) frames "cargo test fails on the .crate" as the goal; nothing holds that goal and the extracted suite is already red at `tests/render/format.rs`. | **SURFACED** — pre-existing prose; owner call whether to reframe. |
| docs.rs and homebrew-core are unaffected (docs.rs builds the lib only; the formula builds from the GitHub tag tarball, which `exclude` does not touch). | Verified, no action. |

Lens 1's two findings stand as dispositioned above. Zero open HIGH.

PR #997, opened 2026-09-12
feat(capture): blank every subtree no decoder reads — allowlist derived by probe

IvanWng97:
## Summary

The recorder now blanks every subtree of a capture that no decoder reads, before the bytes land. The allowlist is **derived by probing the decoders**, never written down.

## Why

Three re-records in a row (#988, #989, #993) leaked a different class of the operator's fingerprint — the installed plugin/skill/agent roster, a global-instruction attachment, a home path — each caught after the fact and redacted by a new blocklist entry (`PII_MARKERS`, five gitleaks rules, the #995 cardinality gate). Every one of those chases an **open set**: the carrier is whatever the CLI vendor adds next week. #995 is left open as the evidence.

Flip the question. Not "is this field sensitive?" (vendor decides, open) but "does a decoder read it?" (our `source/*.rs` decides, closed). The roster leaves because no decoder reads `payload.content[].text` — not because anyone recognised it as a roster. That covers the carriers nobody has named yet, which is the property no blocklist has.

This is the record-time filter VCR uses (`before_record_response`), with the allowlist-over-blocklist stance the PII literature recommends.

## How

`strip_unread` in `capture_fixture.rs`, after every file is on disk and before `refuse_on_pii`:

- For each recorded file, decode the whole file through its registry drive (`Drive::transcript` / `Drive::hooks`) → the baseline event stream.
- Walk each record **top-down**: replace the subtree with a type-neutral value (`""`/`0`/`false`/`[]`/`{}`); if the whole file still decodes to the same events (and the same decode-error/panic counts), keep it blanked and stop; otherwise restore and descend into the children.
- Top-down matters: an unread container collapses to `[]`, so **cardinality goes too** — 67 blanked entries would still say "67", which is exactly the fingerprint #995 was built to catch.
- Whole-file, not per-line: a cross-line key (codex hook `session_id`, CC hook `session_id`) shows up in every event's agent id, so blanking it changes the stream and it stays. (CC's `transcript_path` is a fallback key only, so on the happy path the probe blanks it — review caught the earlier claim that it survives.)
- **The signature is everything the repo asserts on the bytes, not just `AgentEvent`s.** A daemon's envelopes decode to zero events by design — presence rides the registry's `presence_decoder` — so presence is part of the signature, or an openclaw capture would strip to `{}` lines. Decode errors and panics are compared as values, not counts.
- **A hook envelope's top-level `_` keys are kept by namespace.** They are the shim's own stamps (`_pixtuoid_source`, `_shim_ts_ms`, `_pid`); no decoder reads them, but the provenance-falsifiability test dates a capture by `_shim_ts_ms`, and the probe cannot see that test. No list to drift: the namespace is the rule.
- A hand-kept list would drift from the decoder in the one direction nothing catches (a field the decoder stopped reading stays in the bytes). A probe can't drift: it asks the decoder.
- **A line nothing was blanked in is written back byte-for-byte**, and a rewritten line keeps the wire's key order: the `harness` feature now enables `serde_json/preserve_order`, so the recorder's own `-p pixtuoid-core` build and a workspace build agree (the bot caught the recorder alphabetizing every line — untouched ones included — in the fold head). Write-back is atomic (temp file + rename), so a failed strip leaves the recorded bytes intact.

`provenance.json` gains `"deidentified": {"method": "decoder-allowlist", "blanked": N}`. The schema has no `additionalProperties` restriction, so committed provenance stays valid.

## Evidence

- `stripping_blanks_what_no_decoder_reads_and_changes_no_event` — on the committed codex rollout + hooks (temp copy): decode identical before/after; the placeholder roster is gone; the model name (read) and the hook `session_id` (coalescing key) survive.
- `every_committed_scenario_decodes_the_same_once_stripped` — walks **both** fixture roots (conformance + module-owned), every registered source, comparing the full signature (events, failures, presence): 33+ scenarios strip identically. Whole corpus in 3.4 s; the slowest single scenario (cursor/tool-run, 528 subtrees) takes 725 ms.
- `a_daemon_capture_keeps_what_its_presence_decoder_reads` — openclaw's `gateway_start`/port survive; the review lens reproduced the pre-fix strip reducing that line to `{"_pixtuoid_source":"openclaw", …everything else neutral}`.
- `the_shim_stamps_survive_the_strip` — every `_shim_ts_ms` on the CC hook file is byte-identical after the strip; pre-fix they went to `0` and the provenance date gate would have read `1970-01-01`.
- Byte fidelity: openclaw's `gateway_start` line (every field read or a shim stamp) comes back byte-identical; the first rewritten codex line still starts with `timestamp`, where an alphabetized rewrite starts with `ordinal`.
- On the codex rollout: 216 subtrees blanked; the decoder's own read set is 9 fields (effort/model/name/type/token counts), all structural.

What a stripped file looks like: unread record types become `{}`; a read record keeps its read fields and its siblings are neutralised in place. Sparse by design — conformance, not the fixture, documents the wire.

## What the strip does NOT do

It keeps every field a decoder reads, and some of those carry identity: `cwd` (a real path), a tool's command string, a `session_id` that happens to be a path. `refuse_on_pii` and the gitleaks identity rules stay as the live backstop for exactly that residue; the README now says so in one sentence.

A field blanked today stays blank in that fixture. A decoder later extended to read it meets the neutral value and its golden encodes "absent" — re-record the scenario when a decoder grows a read (README).

## Scope

- Only NEW captures are stripped. The committed corpus is untouched here; re-stripping it offline (no billed turn — the strip is local) is a follow-up, as is retiring #995's gate once the corpus has been through this.
- `refuse_on_pii` stays as the key-level backstop.
- The rule relaxation is the owner's call, made in-thread: "fixtures are recorded, never composed" holds — nothing is invented — but a recorded byte the decoders never read no longer ships. README and CLAUDE.md say so in the same commit.

## Comments

18 added comment lines across the two commits, each delete-tested: the derived-not-listed constraint and the top-down/cardinality reason (`strip_unread`); why presence is in the signature (`events_of`); why the `_` namespace is kept (`probe`); the not-JSON fallback (`parsed_lines`); hand-vs-machine edits on the edited-capture test. Two pre-existing docs were corrected because the strip made them false (`WORKSPACE`, `write_payloads`), and one that my insertion had stolen from `refuse_on_pii` was put back.

## Type of change

- [x] New feature (recorder)
- [x] Docs (README, CLAUDE.md bullet)

## Checklist

- [x] TDD: red commit then green.
- [x] `just lint`, `just test` green locally.
- [ ] `just preflight` runs on push.

## AI assistance

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -59,7 +59,7 @@ cargo run --release --example snapshot -- /tmp/snap.png   # render TUI to PNG
 - Renamed a decoded/registered wire name → `just gen-drift-surface`, commit both `crates/*/drift-surface.json` — the crate's own test fails on a stale fragment; regenerate, don't hand-edit.
 - Look-changing PR → `just gen`, commit `docs/images/` + `site/public/demos/`; a scene/web change ALSO needs `just gen-wasm` + commit `site/public/wasm/` (`gen` deliberately excludes it, and nothing catches a skip).
 - Real wire bytes ride ONE pipeline: `pixtuoid_core::harness::Drive` (dev-only `harness` feature). A driver keyed off anything but the source's registry row registers NOTHING.
-- Fixtures are RECORDED, never composed (`just capture-fixture` — BILLED); every scenario declares `provenance.json`. Rules: [`fixtures/README.md`](crates/pixtuoid-core/tests/sources/fixtures/README.md). `just corpus-all` censuses local corpora; `just fixture-age` is advisory/local.
+- Fixtures are RECORDED, never composed (`just capture-fixture` — BILLED), and the recorder blanks every subtree no decoder reads (derived by probe, never listed); every scenario declares `provenance.json`. Rules: [`fixtures/README.md`](crates/pixtuoid-core/tests/sources/fixtures/README.md). `just corpus-all` censuses local corpora; `just fixture-age` is advisory/local.
 - Visual verification for sprite work: snapshot example → `scripts/crop-snapshot.py` → READ the PNG; loop in `.claude/skills/beautify-decoration/SKILL.md`.
 - CI-only gates preflight can't see (a green preflight is NOT a green PR): [`CONTRIBUTING.md#ci-gates`](docs/CONTRIBUTING.md#ci-gates) lists them and what each catches.
 - On-demand advisory (never gates): `just mutants`, `just coverage`, `just bench`, CodSpeed.
--- a/crates/pixtuoid-core/examples/capture_fixture.rs
+++ b/crates/pixtuoid-core/examples/capture_fixture.rs
@@ -641,7 +641,7 @@ mod recorder {
     }
 
     /// A hook envelope's top-level `_` keys are the shim's stamps, kept by namespace:
-    /// no decoder reads them, but `captures.rs` dates a capture by `_shim_ts_ms`.
+    /// `_shim_ts_ms` is read by no decoder, only by `captures.rs`, which dates a capture by it.
     fn probe(
         drive: &pixtuoid_core::harness::Drive,
         lines: &mut [serde_json::Value],
--- a/crates/pixtuoid-core/examples/capture_fixture.rs
+++ b/crates/pixtuoid-core/examples/capture_fixture.rs
@@ -267,9 +267,8 @@ mod recorder {
     fn prepare_workspace() -> std::io::Result<PathBuf> {
         let ws = PathBuf::from(WORKSPACE);
         let base = ws.parent().expect("WORKSPACE has a parent");
-        // A fixed name in shared temp is pre-plantable, so a foreign owner is
-        // REFUSED rather than removed: under /tmp's sticky bit the remove would fail
-        // and the capture would land inside their directory.
+        // A fixed name in shared temp is pre-plantable; a foreign owner is refused, not
+        // removed — under the sticky bit the remove fails and the capture lands in theirs.
         if base.exists() && !owned_by_us(base) {
             eprintln!(
                 "{} exists and is not yours — remove it or run as its owner",
@@ -341,9 +340,8 @@ mod recorder {
         c.args(&cmd[1..])
             .current_dir(ws)
             .env("PIXTUOID_SOCKET", sock);
-        // The driver and the per-CLI cycle scripts write their transcripts beside the
-        // socket, in this run's private sandbox — a fixed shared-temp name is
-        // symlink-followable and two concurrent captures would interleave into it.
+        // Transcripts go beside the socket in this run's private sandbox: a fixed
+        // shared-temp name is symlink-followable and two concurrent captures interleave.
         if let Some(dir) = sock.parent() {
             c.env("TUIDRIVE_LOG", dir.join("tuidrive.log"));
         }
@@ -392,9 +390,8 @@ mod recorder {
             "command": raw.join(" "),
             "deidentified": { "method": "decoder-allowlist", "blanked": blanked },
         });
-        // `command` is the UN-expanded argv, so a scenario driven by an override
-        // records a `{prompt}` placeholder and nothing else says what ran. Only
-        // written when set, so the already-committed records stay schema-clean.
+        // `command` is the UN-expanded argv, so an override-driven scenario records a
+        // `{prompt}` placeholder; written only when set, so committed records stay schema-clean.
         if let Some(map) = prov.as_object_mut() {
             for (key, value) in overrides {
                 if let Some(v) = value {
@@ -404,10 +401,8 @@ mod recorder {
         }
         let out = no_clobber(dest.join("provenance.json"));
         std::fs::write(&out, format!("{}\n", serde_json::to_string_pretty(&prov)?))?;
-        // Keyed on the PROBE's outcome, not on the name looking like a script: a
-        // pty driver is `python3`, passes the name test, and its "unknown" then
-        // disarms `a_recorded_capture_anchors_its_sources_verified_version` for
-        // that whole source. Loud, because the record it just wrote is unusable.
+        // Keyed on the probe's outcome, not the name: a pty driver is `python3` and
+        // passes a name test. Loud, because the record just written is unusable.
         if version.trim() == "unknown" {
             eprintln!(
                 "WARNING: {} records version \"unknown\" — `{cli} --version` did not \
@@ -464,19 +459,14 @@ mod recorder {
         (y % 4 == 0 && y % 100 != 0) || y % 400 == 0
     }
 
-    /// PII is not always a key you can drop — kimi's arrived as the owner column
-    /// inside a captured `ls -la`, and a CLI's ACCOUNT identity is a different
-    /// namespace from the host's (`user_email` slipped past a `$HOME|$USER` grep).
-    /// Identity/inventory keys whose VALUES are the capturer's, not the wire's.
-    /// `user_email` alone was the first cut and it saw none of the MCP-server and
-    /// skill roster that shipped in nine fixtures — a different namespace, same
-    /// class. `just fixture-pii` re-scans the committed tree for the same class
-    /// but matches VALUE shapes, not this list — as gitleaks rules these KEYS fire
-    /// on the tree's own `dev@example.com` redactions. A key with no value shape
-    /// (`obsidian`, `account_id`) is refused only at capture time.
-    /// Both cases of each spelling, because a CLI picks one and its next version
-    /// may pick the other: `userEmail` arrived at claude-code 2.1.261 beside the
-    /// `user_email` already here, and slipped past.
+    /// Identity and inventory keys whose VALUES are the capturer's, not the wire's,
+    /// refused on the stripped bytes — so they guard the fields a decoder reads.
+    /// Not a `$HOME|$USER` grep: a CLI's ACCOUNT identity is a different namespace
+    /// from the host's. `just fixture-pii` matches VALUE shapes instead (as gitleaks
+    /// rules these keys would fire on the tree's own `dev@example.com`), so a key
+    /// with no value shape (`obsidian`, `account_id`) is refused only here. Both
+    /// cases of each spelling: a CLI's next version may pick the other, as
+    /// `userEmail` did beside `user_email`.
     const PII_MARKERS: &[&str] = &[
         "user_email",
         "userEmail",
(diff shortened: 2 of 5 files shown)


IvanWng97:
**Review round 1 — disposition** (three lenses: correctness, design, whole-file comment audit; fold in 5744e80e)

| # | Finding | State |
|---|---|---|
| 1 | **HIGH** (correctness) — daemon envelopes decode to zero `AgentEvent`s, so the probe's baseline was empty and an openclaw capture stripped to `{"_pixtuoid_source":…}` lines; the corpus test compared `[] == []`. | **FIXED** — presence (`registry::presence_decoder_for`, routed per line by `_pixtuoid_source` like the dispatcher) is part of the signature; `a_daemon_capture_keeps_what_its_presence_decoder_reads` + the corpus test now compare the full signature. Lens reproduced the pre-fix strip. |
| 2 | **HIGH** (correctness) — `_shim_ts_ms` is decoder-unread, so it went to `0`; `a_recorded_captures_claims_are_falsified_by_its_own_bytes` dates a CC capture by it and would read `1970-01-01`. | **FIXED** — a hook envelope's top-level `_` keys (the shim's stamps) are kept by namespace, a rule rather than a list; `the_shim_stamps_survive_the_strip`. |
| 3 | **MEDIUM** (all three lenses) — my `strip_unread` insertion stole `refuse_on_pii`'s doc comment; `strip_unread` opened with "A REFUSAL, not a warning" and the refusal gate had no doc. | **FIXED** — restored onto `refuse_on_pii`. (7th instance of the anchor-insert trap in this repo's history; it is in my memory file, and I still did it.) |
| 4 | **MEDIUM** (comment audit) — `WORKSPACE` doc "lets the bytes ship unedited" now false. | **FIXED** — the WHY is that `cwd` is read and kept, so the workspace must already be generic. |
| 5 | **MEDIUM** (comment audit) — README "Only ONE edit… redact PII" / "a line the decoder ignores is the cheap case — drop it" describe by hand what the strip does. | **FIXED** — rule restated on top of the strip; names the read fields (`cwd`, command strings) the strip cannot touch. |
| 6 | **MEDIUM** (comment audit) — bare "67" in the `strip_unread` doc. | **FIXED** — "taking its element count with it". |
| 7 | **MEDIUM** (design Q2b) — a field blanked today meets a decoder extended tomorrow as "absent"; nothing said so. | **FIXED** — one README sentence: re-record when a decoder grows a read. |
| 8 | **MEDIUM** (design Q3) — `a_recorded_capture_that_was_edited_says_so` doc's "edited ⟺ note" no longer tells the whole story; `provenance.schema.json` silent on `deidentified` while the README documents it. | **FIXED** — test doc says hand edits only; schema lists `optional: [prompt, seed, deidentified]`; README row aligned to the schema's `means`. |
| 9 | (design, for correctness) — decode errors/panics compared by count; an error moving between lines cancels out. | **FIXED** — compared as values. |
| 10 | **LOW** (correctness) — in-place `fs::write` could leave a truncated line on an I/O error. | **FIXED** — temp file + rename. |
| 11 | **LOW** (correctness) — PR body claimed CC hook `transcript_path` survives; it is a fallback key and the probe blanks it. | **FIXED** in the body; `plan-miss:` recorded in the fold's message. |
| 12 | **LOW** (correctness) — every line is re-serialized with sorted keys, even when nothing was blanked. | **REFUTED-with-trace** — the workspace `serde_json` `Map` IS a `BTreeMap` (no `preserve_order`), so every decoder already sees sorted keys at runtime; on-disk order carries nothing the decode path can observe. The lens verified no consumer reads order. |
| 13 | (design Q4) — identity in decoder-READ fields (`cwd`, command strings, path-shaped `session_id`) is not de-identified by the strip. | **REFUTED-as-designed, with the mechanism named** — `refuse_on_pii` and `.gitleaks-identity.toml` remain the live backstop for exactly that residue and run on the final bytes; the body's new "What the strip does NOT do" section says so. Not redundant, not contradictory. |
| 14 | (design Q3) — ~23 committed provenance notes say "bytes are otherwise verbatim"; true today, false the moment a scenario is re-recorded, and nothing guards the transition. | **SURFACED** — belongs to the corpus re-strip follow-up: a `recorded` provenance with `deidentified.blanked > 0` must not also claim "verbatim" in its `note`. |
| 15 | (design Q5) — probe could live in `pixtuoid_core::harness` once the offline re-strip tool needs it. | **SURFACED** — YAGNI until that tool exists. |
| 16 | (comment audit) — six pre-existing fn-body comments over the ≤2-line cap (`:271, :345, :396, :408, :1022, :1038` at the reviewed head). | **SURFACED** — pre-existing; each explains a specific block, so moving them loses locality; owner call. |

Gates at the fold: `just clippy` 0, `just test` 3263 passed, example 20 passed. CI-only gates named by the lens (semver, gen-check, wasm-check, windows-test, insta orphans, fixture-pii, doc-check): none reachable by this diff — the recorder is `#[cfg(unix)]` and the committed corpus is untouched.

Round 2 is a delta verify of the fold only.

IvanWng97:
**Review round 2 — delta verify of the fold (5744e80e): CONVERGED, no HIGH.** Gates the verifier observed: `just clippy` 0, example 20 passed, `--test sources` 98 passed. Rows 1–11 CONFIRMED-FIXED; the two HIGH tests were checked for teeth (openclaw's presence decoder reads `type`/`gatewayPort`, so the old signature let them blank; `_shim_ts_ms` is decoder-unread, so the old probe zeroed it).

Delta findings:

| Finding | State |
|---|---|
| **Row 12 wording** — I wrote "the workspace `serde_json` `Map` is a `BTreeMap`"; `preserve_order` IS enabled by the `pixtuoid` binary crate and flows into a workspace-wide build. The recorder's own `-p pixtuoid-core` graph has none, and the operative claim — no decoder reads key order — holds either way. | **Corrected here.** The refutation stands on the verified consumer fact, not on the map type. |
| **MEDIUM** — `probe`'s doc says the kept top-level `_` keys are read by "no decoder"; openclaw's presence decoder reads `_pid` (`openclaw.rs:101-110`). Code correct (the namespace is a safe superset); sentence inaccurate. | **FIXED** in 1519b11f — the sentence names `_shim_ts_ms` as the decoder-unread one. |
| **LOW** — `provenance.schema.json`'s new `optional` list has no test consumer. | **SURFACED** — accurate and documentary; giving it teeth is a new gate, which the convergence contract keeps out of a fix round. |
| **LOW** — `.map_err(\|e\| e.error)` on `persist` "strands a temp file". | **REFUTED-with-trace** — `tempfile::PersistError` carries the `NamedTempFile` in `e.file`; the closure drops it, and `NamedTempFile`'s `Drop` unlinks the file. Nothing is stranded. |
| Row 13 order (`strip_unread` at `main:122` before `refuse_on_pii` at `:137`) | CONFIRMED by the verifier. |

Also landed since the fold, at the owner's request and outside the contract's behavior scope: a whole-file comment pass on the recorder (22c1a83f, 133 → 115 comment lines; no fn-body comment over two lines remains).

IvanWng97:
**Bot review at 5744e80e — disposition**

| Finding | State |
|---|---|
| **HIGH** — `strip_unread` re-serialized every line through a `BTreeMap`-backed `Value`, alphabetizing keys on untouched lines too; the recorder's `-p pixtuoid-core` build has no `preserve_order` (only the `pixtuoid` binary enables it), and no test pinned byte structure. | **FIXED** — a line nothing was blanked in is written back from its original text, and the `harness` feature enables `serde_json/preserve_order` so a rewritten line keeps the wire's key order in every build. Tests shipped red first (e8ee560a), green in c4095908: openclaw's fully-read `gateway_start` line must come back byte-identical; the first rewritten codex line must still start with `timestamp`, where the alphabetized rewrite started with `ordinal`. Lens 1 had rated the same behavior LOW and I refuted it on "no decoder reads order" — the bot's framing (an edit past the sanctioned one, on lines the strip never touched) is the right one, and the refutation is withdrawn. |

Second and last fix round under the convergence contract: both changes are test-shaped (a test failing without them). Delta verify to follow.

IvanWng97:
**Delta verify of fix round 2 (1519b11f → c4095908): CONVERGED, no new defect.** Both red assertions were checked for teeth (openclaw's line is fully read-or-stamped, so the fixed code emits its raw text; the codex line the test pins is one that IS re-serialized, and `preserve_order` is what keeps `timestamp` first). Gates the verifier observed: example 20 passed, `--test sources` 98 passed, clippy 0, and `cargo check -p pixtuoid-core --no-default-features --features harness` 0. `harness` never reaches the wasm build (`pixtuoid-web` depends on core with `default-features = false`, and every `harness` enable is a dev-dependency).

**CodSpeed (advisory, never a gate — CLAUDE.md:65): `jsonl_200_turns` −16.9 %, `hook_200_turns` −15.9 %.** Disposition: **REFUTED-with-trace, and SURFACED as a baseline step.** `cargo codspeed build -p pixtuoid-core` builds the bench through core's self-dev-dependency, i.e. with `harness` on. Before this PR that graph had no `preserve_order`, so the bench measured `Value` on a `BTreeMap`; the shipped `pixtuoid` binary enables `preserve_order` itself (`crates/pixtuoid/Cargo.toml:44-48`, for config re-emission), so production decoders have always run on the `IndexMap`. The bench now measures the map type production uses. Nothing shipped got slower; the benchmark's baseline moved once, by the cost of that map on 200-turn decode. CodSpeed also flagged the comparison as cross-environment.

Two fix rounds used; no further behavior commits under the contract. Awaiting the bot's verdict at c4095908.

IvanWng97:
**Bot review at c4095908 — disposition**

| Finding | State |
|---|---|
| **HIGH** (as rated by the bot; a test-coverage gap, no defect shown) — `every_committed_scenario_decodes_the_same_once_stripped` walked two of the three capture shapes `captures.rs` knows, so a module's flat `fixtures/` (cursor, codex, codewhale, claude, grok) and `delegation/fixtures/<source>` were never stripped by the test, and every sentinel lived in a covered shape. | **FIXED** in e9e2d511 — the walk mirrors `capture_dirs`/`source_of` (conformance scenario, module scenario subtree, flat module fixtures); a file the walk cannot route is now a failure rather than a skip; sentinels pin `cursor/fixtures` and `delegation/fixtures/copilot`. Every newly covered capture strips invariantly, so this is test-only: no behavior commit beyond the two fix rounds already used. 20 passed, clippy 0. |

Premise verified before editing: `captures.rs:117-137` (`source_of`) and `:144-177` (`capture_dirs`) are the three-shape authority the bot cited; the five flat dirs each hold `hook-payloads.jsonl` + `provenance.json` at the root.

PR #998, opened 2026-09-12
chore(fixtures): strip the committed corpus offline — just restrip-fixtures, with exemptions and a verbatim gate

IvanWng97:
## Summary

The committed corpus goes through the same strip the recorder now applies at capture time (#997), offline and unbilled: `just restrip-fixtures`. Every `recorded` capture carries `deidentified: {method, blanked}`, the invented `example-skill-N` placeholders from #993 are gone (the fields they stood in collapse), and a gate keeps a stripped record from ever calling its bytes verbatim again.

## What running the strip over the corpus found

Stripping 47 recorded captures and running the whole suite surfaced four things the probe's signature did not cover, each fixed test-first:

| Found by | Gap | Fix |
|---|---|---|
| `connected_codex_rollout_becomes_a_sprite` (headless replay) | the watcher registers a transcript through the registry's `cwd_extractor`, a seam the harness never calls (it seeds a fixed cwd); codex's `session_meta.cwd` blanked and the replay rendered the unknown-cwd ghost | the extracted cwd joins the signature |
| grok `gate-both-ways` golden | grok reads `cwd` with a `workspaceRoot` fallback that differs only by a trailing slash; `PathBuf` equality drops it, the golden keeps it, so the probe was blind | the signature is compared as rendered, the way a golden is |
| `build_script_pins_the_fixture_beats` (README demo) | the composed `proof-session` is hand-built, and the demo reads its timestamps and text past the decoders | the corpus strip touches `recorded` captures only |
| `cursor::every_task_dispatch_id_is_unpaired_on_the_wire` | the fixture pins two wire premises the decoder drops on purpose (`tool_use_id`; `tool_name: "Task"` is unread because `subagent_type` is the signal); a test reading raw bytes is invisible to any probe | the provenance declares `deidentified: {method: none, why}`; both strip paths honor it; the gate requires the `why` |

Also: the walk missed copilot's and grok's nested `<session>/…jsonl` transcripts (a fourth shape), which is why their placeholders were the last standing — it recurses now, and a provenance with no bytes under it fails the corpus test instead of being skipped.

## Corpus

- 47 recorded captures stripped; composed and `unknown` ones untouched; cursor's flat capture exempt with its reason.
- 0 `example-` placeholders remain anywhere in the corpus.
- 22 notes that said "the bytes are otherwise verbatim" now don't; the three #993 notes describing hand-placed placeholders say the strip took those fields instead. Hand redactions of READ fields (home path in `cwd`, an `ls -la` owner column inside a kept result) still stand and are still noted.
- Every conformance golden is byte-identical; `wire_to_pixels` and the README demo unchanged.

## Evidence

- `every_committed_scenario_decodes_the_same_once_stripped` walks all four shapes and compares the full signature (events, failures, presence, cwds) as rendered.
- `strip_corpus_records_its_count_in_the_provenance_and_is_idempotent`, `strip_corpus_leaves_a_composed_capture_alone`, `strip_corpus_honors_an_exemption_the_provenance_declares`, `the_first_sight_cwd_survives_the_strip`, `a_cwd_whose_fallback_differs_by_a_trailing_slash_survives_the_strip` — each shipped red first.
- `a_stripped_record_does_not_call_its_bytes_verbatim_and_an_exemption_says_why` went red on the stripped corpus before the note rewrites, green after.

## Scope

Retiring #995's cardinality rules and the `captures.rs` placeholder sentinels is the next PR; they are dead on this corpus but the PR that deletes them should be the one that says so.

## Type of change

- [x] Feature (offline corpus strip + exemption)
- [x] Fixture corpus rewrite (mechanical, decode-invariant)
- [x] Docs

## AI assistance


IvanWng97:
**Review round 1 — disposition** (three lenses + the bot at 83f1449b; fold in f8a98789 · b475e7f5 · 1656dfaf)

| # | Finding | State |
|---|---|---|
| 1 | **HIGH** (bot) / **MEDIUM** (design F1) — the strip's own walk resolved the `claude` module's flat capture to the unregistered id `claude`; `captures.rs` maps it to `claude-code`. Harmless today by two coincidences, and a byte-mutating walk must not rest on them. | **FIXED** — one enumeration: `pixtuoid_core::harness::captures` (module-to-source map and the three shapes live there); `captures.rs`, `--strip-corpus`, and the corpus test all call it. Red first: the corpus test asserts every walked source is registered (f8a98789 fails on `claude`). |
| 2 | **MEDIUM** (design F2) — the strip's walk had only a sentinel pin; the canonical walk has the on-disk provenance-count pin. Two copies of one population. | **FIXED** by the same move — there is one walk, and `the_walk_sees_every_provenance_on_disk` now pins what the strip iterates. |
| 3 | **MEDIUM** (design F3) — the schema's `means` for `recorded` asserted an unconditional strip; cursor's exempt capture contradicts it, and the exemption's shape lived only in the README. | **FIXED** — `means` (schema and the README row it is pinned to, verbatim) names `deidentified.method: none` as the reviewed exemption with its reason. |
| 4 | **LOW** (comment lens, 19 notes; correctness lens ×6) — notes still narrated hand redactions of fields the strip has since blanked. | **FIXED** — every claim checked sentinel-by-sentinel against the stripped bytes; the stale sentences go; a note left with nothing true says the strip took the fields. Zero stale claims remain. |
| 5 | (correctness) the corpus rewrite is reproducible: base + this PR's tool → byte-identical `.jsonl` and identical counts; the only divergence is the deliberate cursor exemption. | Verified, no action. |
| 6 | (correctness 3a) every content-reading seam on the registry is in the signature: line decoder, hook decoder, `cwd_extractor`, `presence_decoder`; the rest are path- or argv-only. | Verified, no action. |
| 7 | (comment lens) the `example-skill`/`example-agent`/… sentinels in `a_recorded_capture_that_was_edited_says_so` now match zero committed bytes. | **SURFACED** — dead on this corpus; the #995-retirement PR is the one that deletes them and says so. |

Gates at the fold head: example 25 passed, `--test sources` 99 passed, clippy 0. Round 2 is a delta verify of the fold only.

IvanWng97:
**Round 1, design-lens tail — disposition** (fold in 2ce4c208)

| Finding | State |
|---|---|
| **MEDIUM** (Q1) — an exemption re-opens the one path inventory data the gitleaks backstop cannot see; nothing caps the set. | **FIXED** — `EXEMPT_FROM_STRIP` in `captures.rs` pins the exempt set exactly (one entry); the gate asserts equality, so adding an exemption is a reviewed diff there, not a key in a record. An inventory-shape scan over exempt captures is the #995-retirement PR's question: those captures are the only remaining target such a rule has. |
| **MEDIUM** (Q2) — a registry-derived freeze test for the probe's content-reading seams, so a fifth seam cannot pass vacuously. | **SURFACED** — wanted, but a Rust-level derivation of "every `fn(&Value)` seam on the descriptor" is a design change to `registry.rs` (a regex over `.rs` files is out by the repo's own rule); its own small PR through the design gate. |
| **LOW** (Q3) — `blanked` could be misread as "differs from a fresh capture today". | **FIXED** — one line on `strip_corpus`: it counts the original recording. |
| **NIT** (Q5) — the README paragraph carries three ideas. | Dropped (taste): each sentence is load-bearing and they share one subject, the record's `deidentified`. |
| (Q6) precise #995 scope: only the `example-*`/`mcp__example` sentinels and the `mcp__example*` allowlist branch are dead; the bare `mcp__` rule and the five identity rules stay (reachable via kept `sessionFile`/`cwd` and the exempt capture). | Recorded for the retirement PR. |
| (Q4) cursor is the only raw-byte test reading a dropped field, and it fails loud, not vacuously. | Verified, no action. |

IvanWng97:
**Review round 2 — delta verify of the fold (83f1449b → 2ce4c208): CONVERGED, no HIGH.** Verified by the lens with gates it ran: clippy 0, example 25 passed, `--test sources` 99 passed, `cargo doc --features harness` 0. Rows 1–5 all CONFIRMED-FIXED: one walk (`harness::captures`), zero leftover references to the deleted ones, the red commit's teeth (`claude` unregistered → the new assertion fails), the on-disk completeness pin still against an independent oracle, docs on every new pub item, schema/README `means` pinned verbatim, every remaining redaction sentence backed by a sentinel still in the bytes, the exempt-set pin failing in both directions and platform-safe. Bot at 2ce4c208: `Findings: 0`.

| Finding | State |
|---|---|
| **LOW** — the completeness pin's comment still named `capture_dirs`, deleted in the fold. | **FIXED** in 9cc6fbc6 (doc-only). |
| **LOW**, pre-existing — codex `permission-recorded` note missing a sentence boundary. | **FIXED** in 9cc6fbc6 alongside, since this PR rewrote that note. |

9cc6fbc6 changes one comment and one note; no behavior. Merge-ready pending CI at that head.

PR #999, opened 2026-09-13
chore(capture): drop the capture-time PII key chase for a corpus invariant

IvanWng97:
## Summary

#995 is closed unmerged, as intended: it was the evidence that the blocklist path was wrong. This PR deletes what that path left behind. **Net +101/−56**, and the shape is not a pure deletion: the recorder's capture-time PII key list goes entirely, and one invariant replaces it.

## Why the key list goes

`pii_hits` matched key names as substrings. A list of key names is the open-set chase #997 escaped — the vendor picks the keys. After #997 (capture-time strip) and #998 (corpus re-stripped) a field no decoder reads does not exist in the bytes, so the only keys that can carry a value are ones a decoder reads, and those are covered twice without a list:

- **`identity_needles`** refuses the operator's OWN identity at capture time (`$HOME`, `$USER`, `$LOGNAME`, the git name) — the one set gitleaks cannot know, because it is this machine's.
- **`.gitleaks-identity.toml`** re-scans the committed tree by VALUE shape (email, home path in both separators and omp's dash-encoded form, `mcp__`, `Bearer`, the rendered `Git user:` label) — the one set a key name cannot reach.

So `PII_KEYS`, `PII_MARKERS`, `filled_pii_keys` and the two tests that pinned them are gone; `pii_hits` is the needle comparison and nothing else.

An intermediate commit taught the gate to walk JSON so the list could survive a blanked value (`"user_email":""` would otherwise refuse a re-record). That was the wrong direction — the list itself was the problem — and the deletion supersedes it. The red test that exposed the false refusal is kept in history, not in the tree.

## Census (deletion-shaped PR)

Criterion: **no committed byte matches, and no capture the recorder can now produce would.**

| Artifact | Introduced | State |
|---|---|---|
| `PII_KEYS` (8 keys) + `PII_MARKERS` + `filled_pii_keys` + 2 tests | #929, extended since | **deleted** — see above |
| `captures.rs` — `example-skill`/`example-agent`/`example-plugin`/`mcp__example` sentinels | #993 | **kept** — 0 corpus bytes, but a hand edit can still write one, which makes them dormant like `/home/dev` and ` dev  wheel` rather than dead |
| `.gitleaks-identity.toml` — the `mcp__example*` allowlist on the MCP-name rule | **#929** (3d570e6d) | **deleted** — nothing matches it; the bare `mcp__` rule is strengthened, with no false-positive risk (0 `mcp__` bytes in the corpus) |
| `justfile` `fixture-pii-selftest` — the `mcp__exampleThing` quiet probe | **#929** | **kept**, with its allowlist — the allowlist waives the SENTINEL LITERAL in `captures.rs`, which this scan reads, not placeholder bytes in fixtures; the two are one mechanism. (Measured by review: gitleaks 8.30.1 waives any match containing `example` on its own, so no probe can tell our allowlist from that built-in. It stays because it is the guarantee we own.) |
| `obsidian` marker | #929 | **deleted** — one machine's roster item name |

Kept, with the reason each is still reachable — every one guards a decoder-READ field the strip cannot touch, verified present in the corpus: the five remaining gitleaks identity rules (`cwd`/`transcript_path`, omp's `sessionFile`, tool names, the `Git user:` label), `identity_needles`, and the hand-redaction sentinels (`/Users/dev` in 15 scenarios, `dev@` in the exempt capture — `/home/dev`, ` dev  wheel`, ` dev  staff`, `[redacted` have no bytes today but specify what a hand redaction of a read field looks like, which is still the documented convention).

## What replaces the key list

`no_identity_key_holds_a_value_outside_a_pinned_exemption` (`captures.rs`) walks every committed capture and fails if an identity key holds anything. It fires on all three ways one could return: an exempt capture nobody reviewed, a hand edit, and a decoder that starts reading such a field so the strip keeps it. A line it cannot parse is a panic, not a skip — that is the input the strip declines a whole file for.

Why this is not the chase it replaces: at capture time the list had to guess every key a vendor might invent. As a corpus invariant it asserts what the strip already provides, and a key nobody has named is removed by the strip regardless — so the list never needs to grow to stay sound. Teeth: emptying `EXEMPT_FROM_STRIP` reds it on the exempt capture's `user_email`, the one filled identity key in the corpus.

## Evidence

- `just fixture-pii-selftest` OK (each rule still reds on its own probe, quiet on legitimate paths); `just fixture-pii` no leaks.
- `just test` 3268 passed; example 24 passed; `--test sources` 99 passed; `just clippy` 0.

## Note for the reviewer

The security-review job at 59169084 read red for a workflow reason, not a finding: the review text quoted the `"Bearer "` literal, the runner masked it, and GitHub drops a job output containing a masked value, so the publisher got an empty `REVIEW_JSON` and exited 4. Surfaced in-thread; its one MEDIUM (dropping `obsidian`) is dispositioned there.

## Type of change

- [x] Chore (deletion)

## AI assistance


IvanWng97:
**Security review at 59169084 — read from the analysis job's log**, because the `Publish review` job received an empty `REVIEW_JSON` and exited 4. Cause, from the run: the review text quotes `PII_MARKERS`' `"Bearer "` literal, the runner masks it (13 `***` in the analysis log), and a job output that may contain a masked value is dropped by GitHub, so nothing crossed to the publisher. **SURFACED** — pre-existing workflow fragility (`claude-security-review.yml`): any review that quotes a masked string never publishes and the job reads as red. Not this PR's scope; owner call whether the publisher should read the review from an artifact instead of a job output.

| Finding | State |
|---|---|
| **MEDIUM** — `obsidian` was dropped from the marker list and carried nowhere; a roster field naming it directly would no longer trip `pii_hits`. | **REFUTED-with-trace** — `obsidian` was one machine's roster item (#929), matched as a bare substring. Rosters ride fields no decoder reads; since #997 the strip removes those before `refuse_on_pii` runs (`main`: strip, then refuse), and #998 left zero `obsidian` bytes in the corpus (`git grep` finds none outside this thread). A list of roster item NAMES is the open-set blocklist #997 retired; the residue in decoder-READ fields is covered by the identity needles (`$HOME`/`$USER`/git name), `PII_KEYS` counted only with a value, and the gitleaks value rules — the design lens enumerated every class with its catcher (its table above). The mechanism is the strip's ordering plus `every_committed_scenario_decodes_the_same_once_stripped`. |

IvanWng97:
**Bot review at 8ceaf4a9 — disposition** (fold in 28748e3b)

Both HIGHs hold up once narrowed, and they are the same gap: the strip covers what it touches, but an **exempt** capture skips it, and there the key check was the only catcher. Deleting an open-set chase was right; leaving its reachable population uncovered was not.

| # | Finding | State |
|---|---|---|
| 1 | **HIGH** — `pii_hits` lost the key check; a real `account_id`/`user_id` value is now refused by neither the recorder nor the tree scan, and the test that pinned it went too. | **FIXED**, as an invariant rather than a capture-time guess: `no_identity_key_holds_a_value_outside_a_pinned_exemption` walks every committed capture and fails if an identity key holds anything. Teeth: emptying `EXEMPT_FROM_STRIP` reds it on cursor's `user_email`; the walker proves itself on a filled key at depth beside a blanked one. Verified premise: no decoder reads any of these keys, so outside an exemption the strip blanks them — the corpus holds exactly one filled identity key, the exempt capture's `dev@example.com`. The list no longer needs to grow to stay sound: a key nobody has named is removed by the strip regardless. |
| 2 | **HIGH** — dropping the `example-*` sentinels reopens the gap where a hand-edited fixture passed the edited-capture check silently. | **FIXED** — restored. Their carrier is gone from the bytes, but a hand edit can still write one, which makes them dormant like `/home/dev` and ` dev  wheel`; keeping those and deleting these was inconsistent. |
| 3 | (found by the fix) the `mcp__example*` gitleaks allowlist and the sentinel are ONE mechanism — the allowlist waives the sentinel LITERAL in `captures.rs`, which this scan reads, not placeholder bytes in fixtures. | **FIXED** — allowlist and its quiet probe restored with that reason on them. `just fixture-pii` went red the moment only one half came back, which is how this surfaced; the earlier "nothing matches it" reading was true only while the sentinel was also gone. |

Net now **+101/−56**: what leaves is the capture-time key-name chase (`PII_KEYS`, `PII_MARKERS`, `filled_pii_keys` and their tests) plus `obsidian`; what stays is the operator's-own-identity needles at capture time, the gitleaks value shapes over the tree, and one new invariant.

Gates: `just fixture-pii` no leaks, `fixture-pii-selftest` OK, `just test` 3269 passed, `--test sources` 100 passed, clippy 0.

PR #1000, opened 2026-09-13
build(deps): refresh every in-range npm dependency in site and raycast

IvanWng97:
## Summary

Every in-range npm dependency in `site` and `integrations/raycast`, refreshed in one pass rather than waiting out dependabot's 7-day `cooldown` per release. Lock-only except two lines of `site/package.json`.

## What moved

**site** — the top-level `@astrojs/compiler-rs` 0.3.2 → 0.4.0 is a DEDUPE, not a build-path change: astro 7.3.2 already resolved 0.4.0 from its own nested copy, and the 0.3.2 belonged to `astro-eslint-parser`, which widens to `^0.4.0` at 3.1.0 and collapses the two. shiki 4.4.1 → 4.4.3, mermaid 11.16.1 → 11.17.2, playwright + `@playwright/test` 1.62.1 → 1.63.0, `@astrojs/sitemap`, knip, globals, eslint, typescript-eslint, `@types/node`.

`web-vitals` 6.1.0 → 6.2.1, **keeping its exact pin**: `tests/e2e/font-layout.spec.ts` injects the library's own source into the page to measure CLS against a fixed budget, so the version is part of the measurement, not just a dependency.

**raycast** — `@raycast/api` 2.2.0 → 2.3.1, `@types/node`, `@types/react`, eslint. (`@raycast/utils` was already 2.3.1 and did not move; an earlier revision of this body listed it.)

## Three majors tried and left out

| | Why |
|---|---|
| `mermaid` 12.0.0 | Pulls chevrotain → `lodash-es` carrying five HIGH advisories (GHSA-r5fr-rjxr-66jc code injection via `_.template`; GHSA-f23m-r3pf-42rh prototype pollution). `npm audit` reds on it. |
| `prettier-plugin-astro` 1.0.0 | Reformats 18 of the 23 `.astro` files, where 0.14.1 flags none. A formatter major is its own change, not a rider on a dependency refresh. |
| `typescript` 7 | Already ignored in `.github/dependabot.yml`, both ecosystems: peer-rejected by `@astrojs/check` and by `@raycast/eslint-config`'s typescript-eslint, so `npm ci` ERESOLVEs. |

## Evidence

- `just site-check` **0** — format, lint, astro check, knip, unit, build, doc-render, audit; `npm audit` 0 vulnerabilities.
- raycast: `npx tsc --noEmit` 0, `npx eslint .` 0, `npm run audit` 0 vulnerabilities, and `npm run gen:contract` leaves the committed contract types unchanged.

[...]

PR #1004, opened 2026-09-19
build(deps): dependabot leaves mermaid's major to mermaid-isomorphic

IvanWng97:
## Summary

Adds a dependabot `ignore` row for mermaid's major in `/site`, with the reason on the row — the same discipline as the objc2-app-kit / x11rb rows in the cargo section.

mermaid 12.0.0 (released 2026-09-10) clears the 7-day cooldown this week and the site's direct `mermaid` dep is not ignored. That dep exists only to pin what rehype-mermaid renders with, and mermaid-isomorphic 3.1.0 (the newest) still declares `mermaid: ^11.0.0`, so a direct bump to 12 resolves a SECOND mermaid tree beside the one the rehype plugin actually loads: the lockfile would say 12 and the site would render with 11.

## Related issue

None; weekly dependency triage.

## Type of change

- [x] CI / build

## Verified in-session

- `npm view mermaid-isomorphic@latest dependencies` → `mermaid: '^11.0.0'`
- `npm view rehype-mermaid@latest dependencies` → `mermaid-isomorphic: '^3.0.0'`
- `npm view mermaid time` → `12.0.0: 2026-09-10`
- typescript 7 stays ignored for the reason already on its rows: typescript-eslint and @raycast/eslint-config peer `typescript >=4.8.4 <6.1.0`, @astrojs/check `^5.0.0 || ^6.0.0` (all checked today). prettier-plugin-astro 1.0.1 (09-17) and @raycast/api 2.4.1 (09-14) are not ignored and clear cooldown on their own.

Two lenses — correctness (every registry fact re-fetched in-session, `just ci-observability` 0, zizmor 0) and design — both APPROVE-WITH-NITS. Dispositions:

**FIXED (in the fold):** the row's lift condition had no observer and the row makes a new state reachable (mermaid-isomorphic moves to `^12` → the group PR bumps rehype-mermaid alone → nested 12 renders while the root pin says 11, every gate green; the same path opens through a Dependabot *security* PR, which `update-types` ignores never gate — dependabot-core `ignore_condition.rb:42`) → `site/config/npm-policy.test.mjs` asserts one mermaid in the lock and equal majors with mermaid-isomorphic's declared range, failure message carries the lift instruction; mutation-checked both ways · the row's comment restated the chain `site/knip.jsonc` owns → names that home and the test · `knip.jsonc` gains the back-reference every other held row has.

**REFUTED-with-trace:** "delete the direct dep instead" — the direct pin is load-bearing per `site/knip.jsonc:2-9` (the SVG feeds the visual gates) and is what keeps mermaid on dependabot's roster for in-range bumps (options reference: version updates track manifest-declared deps only).

**SURFACED (pre-existing, owner's call):**
- `.github/dependabot.yml` objc2-app-kit row: a second objc2 core stack (0.6.4) is already in-tree via cpal/softbuffer, so "builds a SECOND objc2 stack" overstates; the operative rule (no second objc2-app-kit, activate API) still holds.
- `.github/dependabot.yml` winit row: "copied verbatim" — `crates/pixtuoid/Cargo.toml` substitutes `wayland-csd-adwaita-notitle`, as its own comment says.
- The four pre-existing ignore rows share the no-observer gap; cargo's get `multiple-versions = "warn"`, the npm ones nothing.
- The main checkout's `site/node_modules` is stale since #1000 (11.16.1 installed vs `^11.17.2`); local hygiene.
- Context: today a mermaid 12 bump would also red `npm run audit` (12 pulls lodash-es with five HIGH advisories, #1000's body) — transient, the ignore is still right.

## How I tested it

- `just ci-observability` exit 0 (106 policy tests, 2 files linted, no violations), re-run after the fold.
- `node --test config/npm-policy.test.mjs`: 3 pass; the new test mutation-checked on an in-memory lock (two copies → fails; moved range with pin held → fails).
- prettier and eslint (the site's own config) clean on the two site files.

## Visual verification

- [x] Not a visual change.

## Checklist

- [x] Docs updated in the same commit — the reason lives on the row.


IvanWng97:
Superseded: the site moves off rehype-mermaid/mermaid-isomorphic to beautiful-mermaid (no browser), which removes the direct mermaid pin this row guarded. Follow-up PR incoming.

PR #1006, opened 2026-09-20
feat(tui): the version popup links to the GitHub release instead of shipping notes

IvanWng97:
## Summary

The "What's new" popup shipped a curated bullet list per version inside the binary (`release_notes()`, 17 match arms), drafted by `just bump` at two awk markers, hand-trimmed on every release branch, and gated by a `notes-curated` CI job plus three framing tests that pinned the prose to 80x24. Every caller passed `CARGO_PKG_VERSION`, so 16 of the 17 arms were data nothing could read.

The popup is now a title, the release address, and the link it already had:

```
Updated to v0.19.0 — Enter to close

  github.com/IvanWng97/pixtuoid/releases/tag/v0.19.0

  ↗ Release notes
```

The changelog's only home becomes the GitHub release, which `release.yml` already publishes with git-cliff from the commit log. The body row is the address itself so the popup's one payload survives a terminal with no mouse reporting (tmux's default, most SSH), where the link cannot be clicked — the notes used to be readable there. It is truncated, not wrapped: a `Paragraph` clips a too-wide line silently, while `truncate` marks the cut. `VERSION_POPUP_W` rises 52 → 56 so the URL lands whole at the popup's own measure.

Found on the way: the `[bump-version-list-here]` awk anchor also matched the marker quoted inside an assert message, so `just bump` injected the `SHIPPED_VERSIONS` row **twice**. The parked local `release/v0.19.0` branch carries that and does not compile. Both markers are gone with the notes.

## Related issue

n/a — owner call, 2026-09-20.

## Type of change

- [x] Refactor / chore (user-visible: the popup's content)

## Deletion census

| Deleted | Why it can go |
|---|---|
| `release_notes()` + 18 arms, `SHIPPED_VERSIONS` (17 rows — it never listed 0.4.0) | only the current version was ever read |
| `[bump-inject-here]`, `[bump-version-list-here]` | nothing injects any more |
| `UNCURATED_MARKER`, `release_notes_are_uncurated` | no draft window exists |
| `just notes-curated` + its ci-lint job (and the `required` manifest row) | nothing to curate |
| `just bump`'s drafting step + its two marker guards | same |
| `release_notes_*`, `current_version_*`, `the_shipped_release_notes_*`, `the_notes_marker_*`, the windowing tests | their subject is gone |
| `a_full_height_modal_never_covers_the_footer_row` | its premise was an overflowing notes band; the footer reservation stays pinned at the geometry level by `a_too_tall_panel_stops_one_row_short_of_the_footer` |

`boot_decision` also loses its `release_notes(...).is_some()` conjunct: the popup fires on any newer version now, which is what it always meant to say.

Correctness and design lenses, both APPROVE-WITH-NITS. Folded: the wrap budget did not subtract the paint indent (three copies of that indent became one `INDENT` const), the mouse-less gap above, two test doc comments that made claims the code did not hold, and two stale references. **SURFACED:** `link_click_rect_is_the_painted_link_cell` derives its expectation from the same constant it tests, so a changed link row stays green — pre-existing, same shape as before.

## How I tested it

- `cargo test -p pixtuoid --lib` — 1001 passed.
- `just lint` — green (the deleted CI job is pinned by the OPA `required`-manifest rule, so a stale manifest would red `ci-observability`).
- Rendered the composed frame, not just the strings: `cargo run --release --example snapshot -- /tmp/snap.png --popup` and read the PNG.

## Visual verification

- [x] Screenshot below.

## Checklist

- [x] Read the root + `crates/pixtuoid/src/tui/` CLAUDE.md.
- [x] No `unwrap()` outside tests; no `println!` on a production path.
- [x] Docs updated in the same commit (`docs/CONTRIBUTING.md`).

## AI assistance

---


IvanWng97:
## Review dispositions (two-lens gate, round 1)

Correctness and design lenses, both APPROVE-WITH-NITS. Both bots: `Findings: 0`.

**FIXED**
- The wrap budget ignored the paint indent: the body was wrapped to the full inner width, then painted with a two-space indent, and a `Paragraph` truncates rather than wraps. Three copies of that indent became one `INDENT` const used by the body, the link span and the click column.
- With the notes gone, a terminal without mouse reporting (tmux's default, most SSH) had no way to reach the release at all — the notes used to be readable there. The body row is now the address itself, truncated with `…` so a cut row says so, and `VERSION_POPUP_W` rises 52 → 56 so it lands whole at the popup's own measure.
- Two test doc comments claimed things the code did not hold (a "wrap path never exercised" that runs on every entrance frame; a "tightest terminal the office lays out on" inherited from a deleted test and below the actual floor).
- The census in the commit message said 17 arms; the base had 18 (`SHIPPED_VERSIONS` listed 17 because it never carried 0.4.0). It now also names the helpers and the renames, and cites **both** tests that pin the footer reservation.
- Two stale references: the justfile's release-group header still advertised a `notes` gate, and `CONFIGURATION.md` still called it the "what's new" popup.

**SURFACED (pre-existing, not grown into this PR)**
- `link_click_rect_is_the_painted_link_cell` derives its expectation from the same constant it tests, so a changed link row stays green. Same shape before this change.

**Cleanup verified, not assumed:** `rg --hidden` across the tree for every notes-era identifier returns nothing; the only surviving "Release notes" strings are the link's own label. No shared helper was orphaned (`panel_inner_width` 4 uses, `window_range` 7, `overflow_cue` 2, `truncate` 12, `borderless_panel` 12, `cell_rect` 10). `just doc-check` exits 0, so nothing links to a deleted item.

PR #1007, opened 2026-09-20
ci(release): release-plz derives the bump and tags a merged release PR

IvanWng97:
## Summary

The only judgement left in the release flow was the bump level: a human picked it, typed `just bump X.Y.Z`, and "features also bump minor" was upheld in review. [release-plz](https://release-plz.dev) derives it from the conventional-commit log plus cargo-semver-checks and opens the release PR, so the human steps become **dispatch → `just gen` on the PR → merge**.

release-plz owns versions and the tag **only**. `release.yml` is untouched and still owns every irreversible publish (crates.io + npm over OIDC, the GitHub release, provenance). `just bump` is deleted.

> **Needs a secret BEFORE this merges** (not before the first dispatch): `RELEASE_PLZ_TOKEN`, a fine-grained PAT scoped to this repository with **Contents** and **Pull requests** read/write. The `tag` job runs on every push to main and the action refuses an empty token, so without it main reds on every push.

## Related issue

n/a — owner call, 2026-09-20. Stacked on #1006 (the version popup no longer ships notes, so the release PR has nothing to hand-curate). **Merge that one first.**

## Type of change

- [x] Refactor / chore (CI / release automation)

## Config, and why each key

`release-plz.toml` carries the WHY on every key; the load-bearing ones:

| Key | Why |
|---|---|
| `publish = false` | crates.io/npm/provenance stay in `release.yml`'s OIDC jobs — no standing token |
| `git_tag_name = "v{{ version }}"` on `pixtuoid` alone | release-plz's multi-package default is `{{ package }}-v{{ version }}`, which matches neither `release.yml`'s `v*` trigger nor homebrew-core's autobump |
| `changelog_update = false` | no CHANGELOG.md here; git-cliff renders the release body |
| `release_always = false` | only a merged `release-plz-*` branch tags |
| `features_always_increment_minor = true` | CONTRIBUTING#versioning, pre-1.0 |
| `semver_check = false` | **the step holding the PAT must run no compiler** — cargo-semver-checks builds both baselines, so every dependency build script would run with a write credential in its environment |

No `version_group`: release-plz already takes the max next version for every `version.workspace = true` package before the group is consulted (`updater.rs`), verified by running it here with and without — identical output.

## Where "what bump?" is answered now

ci-builds' `semver` job, narrowed to release PRs and guarded per **step** (a job-level `if` would make it `skipped`, which `require-jobs` reds). Between releases main carries unreleased changes at the shipped version, where a baseline check reds by design — the question only exists on the release PR. A red there means raise the bump on the branch with `cargo set-version --workspace`.

## Two silent failures, two mechanisms

Both are classes that reach **green** while the release quietly does not happen:

1. **A tag created with `GITHUB_TOKEN` starts no workflow**, so `release.yml` would never fire. Policy pins both release-plz steps to a repository secret (never the automatic token), resolved through the effective step<job<workflow `env` the zizmor token rule already walks — generalised over the variable name rather than copied.
2. **release-plz declines to tag silently** when HEAD's PR did not come from a `release-plz-*` branch. The `tag` job asserts main's workspace version is tagged on origin: true at rest on every push, false exactly when a bump landed untagged.

Policy also pins what else is invisible: two live action steps carrying the exact `{release-pr, release}` command set — **the action runs BOTH commands when `command:` is absent**, so a step that loses the input opens a release PR on every push while the count still reads two — each `[workspace]` kill switch with the reason true for *it*, exactly one tagging package, and the two cross-file pairs a one-sided edit breaks (`git_tag_name` against `release.yml`'s tag glob, `pr_name` against `cliff.toml`'s skip parser). Seven `deny` rules with tests; `release-plz.toml` and `cliff.toml` join the conftest document set (yq fixes its input format per call, so the TOML pair rides a second call joined with jq).

## Why `release-pr` is dispatch-only

release-plz refreshes an open release PR on every push to main, and **closes and re-opens** one that carries non-bot commits (`release_pr.rs`) — which would orphan the human's regenerated stills on the next dependabot merge, and re-run the full CI matrix each time.

## How I tested it

- Ran release-plz 0.3.169 against a throwaway worktree of `main`: 0.18.0 → 0.19.0, "(✓ API compatible changes)", diff = exactly the files `cargo set-version` touched.
- `release --dry-run`: `pixtuoid: creation of tag 'v0.19.0'`; core/scene/hook "no release method enabled".
- `just lint` green (includes `ci-observability`: 133 Rego unit tests, 113 policy checks).

Design-gate lenses (best-practice / adversarial / deepening) ran on the spec before any code; two review lenses ran on the diff. Folded: the command-census hole above, a step-level `if:` that could empty the existence rule, one shared kill-switch message that was false for four of five keys, `pr_branch_prefix` declared rather than inherited, a `${{ }}` interpolated into a `run:` block (zizmor), the secret's real deadline, and five one-home-per-story trims.

**SURFACED to the owner, not fixed here:**
- The `tag` job is the shallow half: with publish/release/changelog off, `release-plz release` reduces to a tag-exists check plus a create-ref, and the assertion step then re-checks the same invariant. A `gh api` create-ref would delete the silent-decline class and the per-push install of three cargo tools — at the cost that a hand-bumped merge would publish with no second human click.
- Per-PR API-break *classification* is gone (goldens show the diff; a human classifies). A per-PR `cargo semver-checks --release-type <t>` keyed off the PR's conventional marker would restore it, as its own PR through the design gate.
- Every upstream claim cited above was fetched this session from release-plz's docs and source, not recalled.

## Checklist

- [x] Read the root CLAUDE.md.
- [x] New behavior has a test (six policy rules, one test each).
- [x] Docs updated in the same commit (`docs/CONTRIBUTING.md` #releasing/#versioning/#ci-gates, `CLAUDE.md`).

## AI assistance

---


IvanWng97:
## Review dispositions (two-lens gate, round 1)

Both lenses returned APPROVE-WITH-NITS. Design lens: 6 MEDIUM. Correctness lens: 1 MEDIUM, 1 LOW.

**FIXED**
- The action was pinned to the **annotated tag object's** sha, not its commit — Actions resolves `uses:` against commits, so the step could never have checked out. Caught only by zizmor's `ref-version-mismatch`, which needs the API; local zizmor is offline by default. Now `b8d6b54b…` (`52860a4f`).
- `conftest fmt` formats with the OPA it embeds, and a dev build disagrees with the pinned v0.68.2 — CI's `fmt --check` red, local green (`c4e43d61`).
- The existence rule counted action steps but never read `with.command`, and **the action runs BOTH commands when the input is absent** — a step that lost it would open a release PR on every push with the gate still green. Added a command-census rule.
- A step-level `if:` could empty the existence rule; it is now a softener like `continue-on-error`.
- One kill-switch message asserted of five keys what held for one. Each key now carries the reason true for it.
- `pr_branch_prefix` declared in `release-plz.toml`, and now **pair-pinned** to ci-builds' `semver` guard: either side edited alone silently leaves every release PR ungated. Verified by mutation.
- A `${{ }}` interpolated into a `run:` block (zizmor template-injection) → routed through `env`.
- The secret's deadline: **before this merges**, not before the first dispatch — the `tag` job runs on every push.
- Five one-home-per-story trims across the toml, the workflow, the justfile and CONTRIBUTING.

**REFUTED (with the mechanism)**
- *"Add a `concurrency` group to the `tag` job"* (LOW, a tag-race red). A bare group **cancels a pending run by default** ([workflow-syntax#concurrency](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#concurrency)), which would trade a self-healing red for a silently skipped tag. The lossless form (`queue: max`) is forbidden outside `release.yml` by this repo's own policy rule, deliberately. The race reds visibly and the message says to re-run.
- *"A job whose steps all skip might report `skipped` and red `require-jobs`"* — unverified by the lens, and it is the leg that would red every PR. Proven empirically on this PR: its head branch is not `release-plz-*`, all four steps skipped, `semver-checks` reports **SUCCESS**.

**SURFACED (owner's call, not done here)**
- The `tag` job is the shallow half: with publish/release/changelog off, `release-plz release` is a tag-exists check plus a create-ref, and the assertion re-checks the same invariant. A `gh api` create-ref would delete the silent-decline class and the per-push install of three cargo tools — at the cost that a hand-bumped merge would publish with no second human click.
- Per-PR API-break **classification** is gone (the goldens show a diff; a human classifies). A per-PR `cargo semver-checks --release-type <t>` keyed off the PR's conventional marker would restore it, as its own PR through the design gate.

IvanWng97:
### Addendum — the one leg the correctness lens left open, now doc-cited

The lens flagged that `ci-builds.yml`'s guard reads `github.head_ref` inside a **called** workflow (`ci.yml` → `ci-builds.yml` via `workflow_call`), and that no run in this repo distinguishes "head_ref is empty there" from "the branch simply doesn't match the prefix" — both skip. If head_ref were empty in a called workflow, `semver` would never run on any branch and the release PR would silently lose its only bump gate, always green. That is exactly the class this PR exists to remove, so it should not rest on reasoning.

GitHub's reusable-workflows reference settles it:

> When a reusable workflow is triggered by a caller workflow, the `github` context is always associated with the caller workflow.

The caller is `ci.yml`, triggered by `pull_request`, so `head_ref` inside `ci-builds.yml` is the caller's PR head branch. The two neighbouring legs are doc-cited too: the context-availability table lists `env` as available to `jobs.<job_id>.steps.if`, and `github` as available to `jobs.<job_id>.env`.

**Free confirmation on the first release PR:** `semver-checks` should show its four steps **running**, not skipped. If they skip on a `release-plz-*` head, the guard is inverted and the gate is vacuous — say so and it becomes a one-line fix.

IvanWng97:
## Reworked: three of the four toggles now let release-plz own the thing

Follow-up to the review above. The design gate and both lenses judged the *original* shape (everything release-plz does disabled, re-implemented in `release.yml`). Looking at what the ecosystem actually does — upstream's `releasing-binaries.md`, ZcashFoundation/zebra, grafana/augurs, orhun/flawz, cargo-generate — three of those `false`s were closing a feature this repo then rebuilt elsewhere.

**Now true**

- **Changelog** moves into `release-plz.toml`'s `[changelog]`. **`cliff.toml` is deleted** and one format renders both `CHANGELOG.md` and the release body; `release.yml`'s git-cliff step goes with it. `changelog_config` is deprecated upstream in favour of this section. *Verified against a real `release-plz update` run: four crates produce **one** changelog file, 40 lines, grouped by the same parsers.*
- **`cargo-semver-checks`** runs where the bump is chosen — release-plz writes the verdict into the release PR body. ci-builds' `semver` job, its per-step release-PR guard, `just semver`, `cargo-semver-checks` in `setup-tools`, and the `pr_branch_prefix` pair-pin that existed only to hold that guard's literal all retire with it.
- **The GitHub release is created as a DRAFT** carrying the changelog; `release.yml` attaches the binaries and publishes it. `softprops/action-gh-release` retains every release key it is not given and publishes a draft once assets land, so the job *loses* its `body:` and `prerelease:` inputs (the latter to `git_release_type = "auto"`). The version is never visible without its binaries — which is what the old ordering gate was really protecting.

**A bug the real run exposed:** `git_tag_name` was only on the package, so each changelog entry built its compare link as `pixtuoid-core-v0.18.0…` — a tag that does not exist, a 404 on every release. Moved to the workspace level (only `pixtuoid` still *creates* a tag); the link is now `compare/v0.18.0...v0.19.0`.

**`publish` stays false, for a different reason than before.** My original argument — don't publish until all six targets build — is weak: crates.io ships *source*, the matrix ships *binaries*, and a musl link failure does not make the published crate wrong. Only cargo-generate gates it; everyone else surveyed leaves it true. The real constraint is that **crates.io Trusted Publisher records are bound to the workflow presenting the OIDC claim**, and all four crates name `release.yml`. Flipping this needs a `release-plz.yml` publisher added to each crate on crates.io *first*, or the release fails after the tag already exists and with `release.yml`'s publish job deleted. That reason now lives in the config comment and in the policy's own failure message.

**Policy follows:** the kill-switch map drops the two keys that are now true, carries the crates.io reason on `publish`, and gains a rule pinning that exactly one package drafts the release — verified by mutation (removing `git_release_draft` reds `ci-observability`).

Gates: `just lint` green (134 Rego unit tests, 114 policy checks), actionlint, zizmor offline **and** online.

IvanWng97:
## Round 2 dispositions

Two lenses reviewed the delta `09d46dc0..HEAD` (the rework that handed four capabilities back to release-plz). Correctness: REQUEST-CHANGES, 1 HIGH + 2 MEDIUM + 2 LOW. Design: 1 HIGH + 4 MEDIUM + nits.

**FIXED**

- **The deleted pin took its own reasoning with it.** `semver_check = false` carried the WHY for a threat that flipping it *realizes*: cargo-semver-checks compiles inside the step holding the PAT, so every dependency build script runs with a repository-scoped write credential — and `publish = true` moved `cargo publish`'s build there too, where round 1 ran it in an OIDC-only job with no PAT. The decision stands; deleting its record did not. `release-plz.yml` now states the accepted risk where the credential is declared, and names the mechanism that bounds it (a GitHub App installation token, ~1 h).
- **The hand-tag recovery path published release candidates as stable.** This workflow's own error message offers `git tag && git push`; that path has no draft to inherit from, so the action creates the release with `prerelease` unset and an `-rc.N` tag lands as **Latest release**. The `prerelease:` input deleted with the draft handoff is back — on the normal path it restates what `git_release_type = "auto"` already chose. The message now says that path also loses the changelog body, and that re-running the job is the fix that keeps it.
- **`git_tag_name` was two copies of one literal and policy pinned the wrong one** — the workspace copy, which every changelog compare link reads, was unpinned and could drift green. One declaration now; the rule resolves it the way release-plz does (package override, else workspace). Verified by mutation.
- **The job that publishes four crates was called `tag`**, hiding the irreversible half in the Actions list → `release` / "publish and tag a merged release PR".
- **The Trusted Publisher filename coupling** was documented in `release-plz.toml` and CONTRIBUTING, neither of which is where someone renaming a workflow looks → moved to the file's own header.
- **`semver` was filed under #ci-gates**, a list headed "these run only in CI", when it is not a job → moved to #releasing. **`just semver` is restored**, marked local-only and not a gate: deleting it left no way for a human to reproduce the verdict before dispatching.
- Five stale references, the worst a `CONTRIBUTING.md` that contradicted itself about whether `semver-checks` is a job, plus two rego identifiers still named after the deleted `cliff.toml`.

**REFUTED, with the citation**

The design lens read release-plz's semver check as advisory — "a breaking change landing under a `fix` subject now gets a patch bump with no red anywhere". Upstream says the opposite: *"When release-plz detects API breaking changes, it updates the version of the package with a 'major semver Bump'"*, 0.4.21 → 0.5.0 ([semver-check.md](https://release-plz.dev/docs/semver-check)). It corrects the bump rather than reporting on it, which is stronger than the red job it replaced, not weaker.

**VERIFIED by the lenses against pinned sources, not READMEs**

- The draft handoff: `action-gh-release@v3` `github.ts:710` keeps `draft` on update, `:762` patches `draft: false` once assets land, `:694` retains the existing body when none is passed. The comment on that step is accurate.
- Publish ordering: release-plz iterates a vec its own source documents as release-ordered (`project.rs:109`) and `wait_until_published` (`release.rs:956`) gates each crate on the index before the next.

**OPEN — needs an owner decision, and it is the last one**

`publish = true` leaves the crates.io upload with **no CI gate**. The `release` job has no `needs`, and `ci.yml` fires on the same push *concurrently*, so release-plz can publish before `ci-gate` finishes. Round 1 gated it on `needs: [check, build, deb]`; npm, equally irreversible, still carries that gate — the two are now asymmetric. Add that branch protection here is not `strict`, so the tag can land on a tree no CI ran.

Three shapes, owner's call:
1. **Turn on *Require branches to be up to date* for `main`** — makes "the tree that was tested is the tree that merged" structural. Cost: every PR must update before merge. (Both lenses' recommendation, and mine.)
2. **Wait for `ci-gate` inside the `release` job** — closes staleness *and* concurrency, at the cost of new machinery and a slower release.
3. **`publish = false`** — back to round 1's gate, giving up this simplification.

`docs/CONTRIBUTING.md` currently asks the human to update the branch in prose, and CLAUDE.md is explicit that prose has no failure mode. Whichever shape wins, that sentence should become the mechanism.

IvanWng97:
**SURFACED, merging anyway** — the security reviewer's MEDIUM at `release-plz.yml:65` is real and this change introduces it: both release-plz jobs now compile the dependency tree inside the step holding `RELEASE_PLZ_TOKEN`, so a malicious `build.rs` could read a repo-write credential. Before this migration the release flow carried **no** long-lived credential at all — the human's own `git push` of the tag was the trigger.

Owner's call to accept it for now. The bound is real but narrow: the PAT is fine-grained, scoped to this repository, and carries only Contents + Pull requests. The named mechanism is a GitHub App installation token (~1 h, `actions/create-github-app-token`), which release-plz documents and its own repository uses; switching is a step-level change plus two secrets, fully decoupled from the release flow.

Ecosystem check for the record: of ~1968 workflow files invoking `release-plz/action`, 618 name a PAT secret and 255 use an App token, so the PAT shape is the majority. The ~1210 using `secrets.GITHUB_TOKEN` mostly ship no binaries and need no second workflow to fire.

PR #1009, opened 2026-09-21
feat(scene)!: the wall board drops its version, flaps its mood line, and glows like a neon sign

IvanWng97:
## What

The wall board stops carrying the version, and becomes worth looking at.

- **L1** is `pixtuoid` + `★ Star`. The brand line baked `CARGO_PKG_VERSION`, and the board is in every committed still, so each release re-rendered all of them. **A release PR no longer needs `just gen`** (runbook + CLAUDE.md updated; this PR carries the last version-driven regen).
- **L2** alternates the tally (`▲2 wait  ●4 work  ○6 idle`) with one plain-English line per office mood (`▲ 2 agents need you!`, `● heads down, shipping`, `○ coffee break?`), swapped by a **split-flap roll**: a column only rolls forward through the drum and stops on its letter, columns land left to right, a column that isn't changing stays put.
- **The panel is a lit sign**: near-white tube, a colored halo that lands on the wall and on what hangs under it, faintly tinted interior. Only "someone is waiting" changes the hue (new theme role `ui.neon_alert`); busy = brand hue, calm = the same hue lower, a truly empty office = a starved tube that throws no light and stutters. A mood change eases over 1.6 s; the halo halves by day.

The owner ratified the look on a running prototype before any of this was built.

## Reuse, not new wheels (owner directive)

| need | reused / consolidated |
|---|---|
| epoch ms | three copies (`creatures`, `chitchat`, `pixel_painter`) → one `anim::epoch_ms`, below both layers |
| distance-falloff light | `paint_radial_falloff` + `paint_warm_halo` + the new glow share one `blend_falloff` loop |
| mood fade | `anim::eased_progress` (what `FloorTransition` uses) — no new easing primitive, `LightingState` untouched |
| "really empty" debounce | `LightingState`'s own VERDICT (`dimmed()`), not its smoothed level — the sign starves only once the room is judged empty |
| tint → white / black | five local literals → `palette::WHITE` / `BLACK` |
| drum fallback | `col % N` — no hash |

**Byte-identity proof for the refactors** (gen-check can't see them — this PR regenerates every still): six fixed-clock stills (day, night, `--empty`, cyberpunk, 100×34, gateway) rendered on `main` and on this branch are identical in every pixel outside the sign's reach (panel grown by the halo radius).

## Design gate

Best-practice (fetched): ratatui 0.30.2 `Paragraph` keeps the old bg under a glyph and shows only the bottom half-block pixel → the interior is ONE flat color and the halo never enters it (pinned). scottbez1/splitflap firmware: forward-only, fixed drum, on-target modules stay → the roll model. EAW: every glyph in play is 1 col (pinned TUI-side).

Adversarial, dispositioned:
- **HIGH — floating paints an empty office at its ambient 1 fps** → a roll would freeze mid-scramble. FIXED: an empty office has no persona; L2 stays on `— office empty —`.
- **"stills render on the hour" was false** for the proof poster (+10 s) and the wasm stills (+100 s, one at :30). FIXED: both guards read the capture instants out of `scripts/media.json`.
- **Empty had no debounce** → FIXED via `LightingState` (above).
- **Site rows collapsed the baked spaces** and jumped as the segment count changed mid-roll. FIXED: `white-space: pre`, no flex gap; new Playwright test is red on the old CSS.
- **Semver**: `feat!`. API-golden delta is two lines (`OfficeColors::neon_frame_base` removed — the tube derives from `ui.neon_brand`; `UiColors::neon_alert` added). `SimFrame` also gains a `pub(crate)` field, so it is no longer literal-constructible — the golden can't show that, cargo-semver-checks will; `feat!` covers it. The sign's state types and `OfficeMood` are `pub(crate)`.

## impl-plan answers

1. **Data shapes** — `OfficeMood` (new identity; `StateCounts` is its input). `NeonState` per `FloorCtx`, like `LightingState`.
2. **Consumers** — `now` → `board_mood_at`; `OfficeMood` → tally's empty line, persona, `NeonState::tick`; `SimFrame.neon` → `paint_frame` (panel + late glow).
3. **Siblings** — three painters: TUI calls `build_board` directly, floating + web go through `FloorSession::board`; all three switched. Six themes lose the field together.
3b. **Reachability** — "version in committed media": only `board_brand()` (`git grep CARGO_PKG_VERSION`; no media job opens the version popup). "board text built": TUI painter, `FloorSession::board`, tests.
4. **Untrusted input** — none (counts + clock).
5. **Tests** — timeline both sides of every boundary (offsets derived from the consts); left-to-right settle; drum predecessor; blank-blank stays; persona fits or is withheld; singular/plural; no-persona offices never leave the tally; media.json-driven "no committed frame mid-roll / mid-flash"; `NeonState` snap, ease, reversal, backward clock, room-debounce, flash windows both edges; look hue/daylight/power; interior flat + halo radius both sides; TUI display-width == chars for every L2 face; site columns through a roll.
6. **Sharp edges read** — board.rs 1-col vocabulary; wall_board strict `<` fit assert; the old pulse's f32-cast freeze (breath reduces by integer modulo first); `render_floor` projected-scene eviction (state is per-`FloorCtx`, not per-agent); sim/paint split (tick in `sim_step`).
7. **Verification** — preflight; WATCHED: six themes day+night, empty, 100×34, floating snapshot, live hero in headless Chromium (234 samples across a roll: left edge fixed, width == chars × advance). CI-only gates this can red: api-surface (regenerated), gen-check (`just gen`), gen-wasm-check (`just gen-wasm`), site e2e.
8. **Layering** — no new seam. New `pub`: `BOARD_BRAND` (two out-of-crate tests read it), `UiColors::neon_alert`, `build_board`'s `now` param.

## Two-lens review (2 rounds, the cap)

Five lenses in round 1 (correctness · design · whole-file comment audit · adversarial-toward-revert on the refactors · generated art, frames READ), one delta verifier in round 2.

- **HIGH, found by two lenses independently and reproduced on the real crate — FIXED.** The sign starved on `room_level < 1.0`, but an f32 ease that once left 1.0 stalls a few ulps short of it forever: on any floor that had ever dimmed, a transcript gap or the last agent walking out dropped the sign in a fully lit room. `LightingState` now exposes its own debounce verdict; `snap_to_empty` back-dates the debounce so the `--empty` still keeps judging the floor empty. Pinned twice: on the real dim → repopulate → walkout sequence, and at the `sim_step` seam (mutation-checked red).
- **Wasm stills captured the mood-CHANGE frame**, lit for the old mood (they step twice, 100 s apart) — FIXED: a sign nobody has drawn for longer than a fade snaps. Same fix for a floor switched back to.
- **The stutter under floating's 1 fps empty-office cadence** held a flash for a full second (three lenses) — FIXED: a flash is only emitted when the frame gap is shorter than the shortest flash, so a still and a slow painter get the steady starved tube.
- **Alert hue did nothing in gruvbox** (`label_waiting` == `neon_brand`) and little in cyberpunk — FIXED with the `neon_alert` role + a per-theme distinctness test. **Owner: the gruvbox (bright red `#fb4934`) and cyberpunk (electric yellow) values are my picks — please eyeball.**
- **media.json guard panicked on a crates.io-packaged `cargo test`** — FIXED (skips like its siblings); the "an hour is an even count of halves" premise is a const assert.
- **A/B blind spot**: the west sun spot sits inside the region the A/B render masked — FIXED with a pixel-exact test of the shared falloff loop against the two loops it replaced.
- **Round 2**: doc-check was red (two published docs intra-linked the now-private `OfficeMood`) — reverted/deleted; the join test above added; a starved tube throws no halo (art lens; matches the prototype).
- REFUTED / dropped as taste: README GIF halo bands (tried 256 colors: no better, +70 % size — the banding is the GIF encoder's per-frame quantization; still reads as a warm glow at README scale); the `pets` clip's loop seam lands at the end of a roll (every clip seam is already a hard cut of the whole scene); state-driven text changes snap without a flap (declared non-goal).

## Declared deltas from the prototype the owner ran

Breath periods 2600/9000 → 3000/6000 and stutter cycle 5200 → 5000; an empty office keeps `— office empty —` (no persona lines); the sign holds its calm light through the room's empty-debounce before starving; gruvbox/cyberpunk get their own alert hue.

**Heads-up for the visual check:** the sample scene has two waiting agents, so every regenerated still shows the sign in its ALERT hue, not brand cyan.

## SURFACED (pre-existing, owner's call)

- `release-plz.toml:7,26` still say "`tag` job" (it is `release`).
- `paint_ceiling_halos` (ambient.rs) still open-codes a distance falloff the new shared loop could carry.
- A fourth epoch-ms copy lives in the binary's marquee (`tui/widgets/mod.rs`), out of reach of the scene-private helper.
- At buffer widths 37–47 a wall fixture overpaints part of the panel interior.
- Several pre-existing fn-body comments exceed two lines in functions this PR touched (`wall_board.rs:67-71`, `pixel_painter/mod.rs` ×4, `floor/mod.rs:646-648`).

- Floating reads the clock three times per redraw (`window.rs`), so the board's `now` isn't the render's `now` (ms skew, harmless).
- At small terminals wall decor covers part of the sign's bottom tube (z-order predates this PR; the glow makes it more noticeable).
- `release-pr` stays dispatch-only; with no `just gen` commit to orphan, the remaining reasons are a human-raised bump and CI cost per refresh.


IvanWng97:
## Disposition — the one red check (`site` › Lighthouse `interactive`)

**SURFACED to the owner, who decided: merge, then fix the budget's root cause in its own PR before the release.**

- Failure: home-page TTI median **9022 ms** vs `maxNumericValue` 9000 (runs 9039 / 8959 / 9022).
- Cause, from the uploaded `lighthouse-reports` artifacts: the CI preview serves the wasm **uncompressed** (transfer 1,064,386 B), and under simulated slow 4G the home page's TTI/LCP is essentially that download. The last six passing runs sit at medians **8953–8966 ms** — ~40 ms (0.45 %) of headroom. This PR grows the wasm by 10,393 B (+1 %) ≈ +55 ms.
- Measured split of the growth (throwaway builds): flap + persona ≈ 4.5 KB, neon sign + the rest ≈ 4.8 KB.
- Production (GitHub Pages, fetched today) serves the same file **gzip**-encoded at 414,696 B, so the budget is pinned to a transfer ~2.5× larger than any visitor sees; as written it means "the wasm may never grow".
- Follow-up PR (next): a size-oriented build profile for the wasm (the workspace `release` profile is speed-tuned: LTO + 1 CGU + `panic=abort`, but default `opt-level=3`), plus a typed persona pool that already measured −1 KB. Raising the budget is the fallback only.

Everything else at `775189c7`: 45 checks pass; `claude-review` and `claude-security-review` both report **Findings: 0** at this head; CodSpeed reports no change. Local lenses and their dispositions are in the PR body.

